PR #1, opened 2021-07-19
Initial spring, gradle, openapi, and testing.

wchamber:
Add some initial infrastructure for the tanagra service.

- Spring Boot application
- OpenApi with swagger and trivial status endpoint
- Gradle setup
- Gradle quality plugin setup, based on terra-common-libs
- Protobuf setup
- Initial junit test

(I'm going to be working backwards for a bit from my prototyping, so I may be including things that aren't used just yet.)


wchamber:
Asking for two reviews for this initial commit since it'd be nice to get any best practices in a good place from the start.

.gitignore
```diff
@@ -0,0 +1,53 @@
...
+### VS Code ###
+.vscode/
+
+### Emacs backup files
+*.*~
+
+### Mac directory metadata
+.DS_Store
+
+### OpenAPI CodeGen detritus
```
calbach, line 42: This is a bit suspect, in particular globally ignoring pom.xml. There's a `.openapi-generator-ignore` file for the purpose of omitting certain generated artifacts, feels like that could be used here instead - or if it's broken for some reason, maybe this could be more targeted to a specific directory

.gitignore
```diff
@@ -0,0 +1,53 @@
+HELP.md
```
calbach, line 1: I guess this might be a historical relic, but what's wrong with having a file called `HELP.md`?

build.gradle
```diff
@@ -0,0 +1,91 @@
...
+        exclude group: "org.junit.vintage", module: "junit-vintage-engine"
+    }
+}
+
+apply from: "$rootDir/gradle/dependency-locking.gradle"
+apply from: "$rootDir/gradle/javadoc.gradle"
+apply from: "$rootDir/gradle/open-api.gradle"
+apply from: "$rootDir/gradle/protobuf.gradle"
+apply from: "$rootDir/gradle/spotless.gradle"
+apply from: "$rootDir/gradle/test.gradle"
```
calbach, line 91: Can you explain what is happening here? Is this basically a Gradle file subinclude?

gradle/dependency-locking.gradle
```diff
@@ -0,0 +1,19 @@
+// see https://docs.gradle.org/current/userguide/dependency_locking.html
```
calbach, line 1: I have not seen a gradle lockfile before 🤦 . Very nice - we are stealing for Workbench..

gradle/config/pmd/pmd.xml
```diff
@@ -0,0 +1,151 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ruleset name="Base ruleset" xmlns="http://pmd.sourceforge.net/ruleset/2.0.0"
```
calbach, line 2: Nice! I tried and utterly failed to find a good PMD template for this config file in the past. May steal this.

gradle/config/spotbugs/html-report-style.xsl
```diff
@@ -0,0 +1,306 @@
...
+				</b>
+			</td>
+		</tr>
+	</table>
+	<p><i>(* Defects per Thousand lines of non-commenting source statements)</i></p>
+	<p><br/><br/></p>
+
+</xsl:template>
+
+</xsl:stylesheet>
```
calbach, line 306: This file doesn't look unique to the project AFAICT. Do we need this copy?

src/main/java/bio/terra/tanagra/app/controller/UnauthenticatedApiController.java
```diff
@@ -0,0 +1,30 @@
...
+
+/** Spring controller for the unauthenticated API methods. */
+@Controller
+public class UnauthenticatedApiController implements UnauthenticatedApi {
+  @Override
+  public ResponseEntity<Void> serviceStatus() {
+    return new ResponseEntity<>(HttpStatus.OK);
+  }
+
+  /** Required if using Swagger-CodeGen, but actually we don't need this. */
```
calbach, line 19: I'm not familiar with this `ObjectMapper` pattern - is this required? Or is this an exception to the normal controller pattern here?

src/main/resources/api/service_openapi.yaml
```diff
@@ -0,0 +1,36 @@
+openapi: 3.0.3
```
calbach, line 1: nit: Is the filename standard for Terra platform? might have just expected `tanagra.yaml`

.gitignore
```diff
@@ -0,0 +1,53 @@
...
+pom.xml
+
+# Package Files #
+*.war
+*.nar
+*.ear
+*.zip
+*.tar.gz
+*.rar
+
```
calbach, line 53: Please also add a LICENSE to the root directory. Probably BSD-3 clause (probably copy another, e.g. Workbench if you like.)

calbach (commented):
The gradle dir structure is really nice.

One comment on top-level directory structure: how do we want to lay things out for a monorepo?

Workbench has:
- /api
- /ui
- /e2e
- /...

Following this pattern, I would basically move all the gradle stuff into a subdir. Alternatively, gradle could be the "main" entrypoint, and the ui could be a subdir.

build.gradle
```diff
@@ -0,0 +1,91 @@
+buildscript {
+    repositories {
+        mavenCentral()
+    }
+    dependencies {
+        classpath "com.google.protobuf:protobuf-gradle-plugin:0.8.15"
```
yonghaoy, line 6: A general Tech stack question: there were some discussion whether to use protobuf, and seems the decision is not for now. 
Is Tanagra not part of McTerra, hence we can choose tech stack on our own? If yes, what is the criteria of choosing them?

src/main/java/bio/terra/tanagra/app/controller/UnauthenticatedApiController.java
yonghaoy: Swagger-codegen generated Class has that two methods

yonghaoy (commented):
Question: What is the reason having different `.gradle` files instead of having them in one `gradle` like the other Apps we are familiar with?

wchamber (commented):
I moved almost everything under an `api` directory to match AoU. Let me know if you have any other naming suggestions. Also considered `backend` or `service`.

I think there's a world where we eventually want to share some of the gradle config, or use gradle subprojects, but I'm inclined to keep it simple for now.

thanks!

wchamber:
> Question: What is the reason having different `.gradle` files instead of having them in one `gradle` like the other Apps we are familiar with?

Mostly organizational. It makes it clear which parts are related to which gradle functions, makin g it easier to find what's being done and understand/change it, instead of one giant build.gradle. I got the idea from terra-common-libs, and I liked it a lot.

.gitignore
calbach: That's what I would put. Maybe give a heads up to Shimon to double-check.

calbach (approved):
Nice - LGTM

build.gradle
yonghaoy: Nice reference! Thanks!

yonghaoy (approved):

.gitignore
wchamber: Thanks, verified with Shimon.

PR #2, opened 2021-07-20
Add initial Github action for testing

wchamber:
Adds an initial github action to test that the api builds and tests pass on pull requests and commits to main.
Based on the [terra buffer service test action](https://github.com/DataBiosphere/terra-resource-buffer/blob/master/.github/workflows/test.yml)


.github/workflows/api-test.yaml
```diff
@@ -0,0 +1,48 @@
...
+    branches: [ main ]
+  pull_request:
+    branches: [ '**' ]
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up AdoptOpenJDK 11
+        uses: joschi/setup-jdk@v2
```
calbach, line 17: What's the rationale for deferring installation rather than just running it directly on the openjdk11 image?

.github/workflows/api-test.yaml
```diff
@@ -0,0 +1,48 @@
...
+name: Run API Service Tests
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ '**' ]
+jobs:
+  test:
+    runs-on: ubuntu-latest
```
calbach, line 13: I searched a bit but couldn't find a definitive answer. Does GitHub actions have a caching policy on docker images?

.github/workflows/api-test.yaml
```diff
@@ -0,0 +1,48 @@
...
+          java-version: 11
+      - name: Cache Gradle packages
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle') }}
+          restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
+      - name: Grant execute permission for gradlew
```
calbach, line 28: Hmm.. I would expect it to be executable in source - is it not already?

.github/workflows/api-test.yaml
```diff
@@ -0,0 +1,48 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: Run API Service Tests
```
calbach, line 4: Would spotless / static analysis be a separate workflow?

wchamber: they're done right now as a part of `build`, so they are included. Actually, so is `test`, so I'll remove it for now with a comment.

.github/workflows/api-test.yaml
wchamber: Nothing deep, just what's suggested in the examples and the other services we've used. e.g. https://docs.github.com/en/actions/guides/building-and-testing-java-with-gradle

.github/workflows/api-test.yaml
wchamber: Really not an expert. I'm not sure if it's a big deal if we're not using something exotic. My latest run took 11s to setup the job, the actions, and the jdk, which seems fine. https://github.com/DataBiosphere/tanagra/runs/3116730025
Most of the caching literature seems to be about what's pulled into the environment, not so much the running docker image itself, https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows

.github/workflows/api-test.yaml
wchamber: You're right, cargo culting gone wrong.

wchamber (commented):
thanks!

.github/workflows/api-test.yaml
calbach: Aha, thanks, the comment is helpful.

calbach (approved):

PR #3, opened 2021-07-22
Add first search SQL interpretation for filters.

wchamber:
Starting in the middle, add the first data classes for entities and attributes and the first query classes for filters and expressions.
The query classes are meant to develop a query search grammar that can be walked for validation and sql construction.
Looking at `SqlVisitorTest` might be clearest place to start.

### DIFF
--- a/api/src/main/java/bio/terra/tanagra/service/search/Expression.java
+++ b/api/src/main/java/bio/terra/tanagra/service/search/Expression.java
@@ -1,0 +1,54 @@
+package bio.terra.tanagra.service.search;
+
+import com.google.auto.value.AutoValue;
+
+/** A construct in a query syntax tree that evaluates to a value. */
+public interface Expression {
+  /**
+   * A visitor pattern interface for an {@link Expression}.
+   *
+   * @param <R> the return value for the visitor.
+   */
+  interface Visitor<R> {
+    R visitLiteral(Literal literal);
+
+    R visitAttribute(AttributeExpression attributeExpression);
+  }
+
+  /** Accept the {@link Visitor} pattern. */
+  <R> R accept(Visitor<R> visitor);
+
+  /** An {@link Expression} that's a literal value. */
+  @AutoValue
+  abstract class Literal implements Expression {
+    public abstract DataType dataType();
+
+    public abstract String value();
+
+    public static Literal create(DataType dataType, String value) {
+      return new AutoValue_Expression_Literal(dataType, value);
+    }
+
+    @Override
+    public <R> R accept(Visitor<R> visitor) {
+      return visitor.visitLiteral(this);
+    }
+  }
+
+  /** An {@link Expression} that's an {@link bio.terra.tanagra.service.search.Attribute} */
+  @AutoValue
+  abstract class AttributeExpression implements Expression {
+    public abstract Attribute attribute();
+
+    public static AttributeExpression create(Attribute attribute) {
+      return new AutoValue_Expression_AttributeExpression(attribute);
+    }
+
+    @Override
+    public <R> R accept(Visitor<R> visitor) {
+      return visitor.visitAttribute(this);
+    }
+  }
+
+  // TODO implement math expressions as needed, e.g. minus, plus.
+}
--- a/api/src/main/java/bio/terra/tanagra/service/search/Filter.java
+++ b/api/src/main/java/bio/terra/tanagra/service/search/Filter.java
@@ -1,0 +1,78 @@
+package bio.terra.tanagra.service.search;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+
+/**
+ * A construct in a query that represents a filter on a {@link Entity}.
+ *
+ * <p>A filter is a function that evaluates to true or false on every instance of the associated
+ * {@link Entity}.
+ */
+public interface Filter {
+  /**
+   * A visitor pattern interface for a {@link Filter}.
+   *
+   * @param <R> the return value for the visitor.
+   */
+  interface Visitor<R> {
+    R visitArrayFunction(ArrayFunction arrayFunction);
+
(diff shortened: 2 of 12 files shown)


api/src/main/java/bio/terra/tanagra/service/search/SqlVisitor.java
```diff
@@ -0,0 +1,85 @@
...
+  /** A {@link Visitor} for creating SQL for filters. */
+  static class FilterVisitor implements Visitor<String> {
+    private final SearchContext searchContext;
+
+    FilterVisitor(SearchContext searchContext) {
+      this.searchContext = searchContext;
+    }
+
+    @Override
+    public String visitArrayFunction(ArrayFunction arrayFunction) {
```
calbach, line 21: I wish there were something better/safer available than raw SQL strings, but my searches have not been too successful. JPA Criteria API might be the closest thing (also portable across datasets), but I'm very skeptical it will be able to handle all the JOINs you want to construct. Data apps team may also have some experience with trying that and can better describe the limitations.

api/src/main/java/bio/terra/tanagra/service/search/SqlVisitor.java
```diff
@@ -0,0 +1,85 @@
...
+  static class ExpressionVisitor implements Expression.Visitor<String> {
+    private final SearchContext searchContext;
+
+    ExpressionVisitor(SearchContext searchContext) {
+      this.searchContext = searchContext;
+    }
+
+    @Override
+    public String visitLiteral(Expression.Literal literal) {
+      // TODO consider parametrizing output to avoid injection.
```
calbach, line 72: spelling nit: parameterizing

api/src/test/java/bio/terra/tanagra/service/search/SqlVisitorTest.java
```diff
@@ -0,0 +1,88 @@
...
+import com.google.common.collect.ImmutableList;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+@Tag("unit")
+public class SqlVisitorTest {
+  private static final SearchContext SIMPLE_CONTEXT =
+      SearchContext.builder().underlaySqlResolver(new SimpleUnderlaySqlResolver()).build();
+
+  private static final Entity PERSON = Entity.builder().name("person").underlay("foo").build();
```
calbach, line 16: What did `underlay("foo")` do here? Is this property necessary? Or is it just being ignored somehow due to the nature of the test?

api/src/main/java/bio/terra/tanagra/service/search/SqlVisitor.java
```diff
@@ -0,0 +1,85 @@
...
+      if (DataType.STRING.equals(literal.dataType())) {
+        return String.format("'%s'", literal.value());
+      }
+      return literal.value();
```
calbach, line 76: nit: Not sure this default will continue to work as you extend the enum - should this be another switch with an exception at the end?

api/src/main/java/bio/terra/tanagra/service/search/Filter.java
```diff
@@ -0,0 +1,78 @@
...
+      return visitor.visitArrayFunction(this);
+    }
+
+    public enum Operator {
+      AND, // All of the operands must be true.
+      OR // Any of the operands must be true.
+    }
+  }
+
+  // TODO implement a relationship filter tht allows chaining to another entity.
```
calbach, line 77: spelling nit: relationship

calbach (approved):
Some minor comments, but overall the interface looks well designed. Apologies for the slow review.

api/src/main/java/bio/terra/tanagra/service/search/Expression.java
```diff
@@ -0,0 +1,54 @@
...
+      return new AutoValue_Expression_AttributeExpression(attribute);
+    }
+
+    @Override
+    public <R> R accept(Visitor<R> visitor) {
+      return visitor.visitAttribute(this);
+    }
+  }
+
+  // TODO implement math expressions as needed, e.g. minus, plus.
```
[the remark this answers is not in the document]
wchamber: not afaik

api/src/main/java/bio/terra/tanagra/service/search/SqlVisitor.java
wchamber: One thing I'm eyeing is jOOQ DSL API, particularly as a "dynamic" schema. https://www.jooq.org/doc/3.16/manual/getting-started/use-cases/jooq-as-a-sql-builder-without-codegeneration/

I'd like to get some basics in place with raw strings, then do try implementing a parallel set of visitors for jOOQ and see how that goes.

api/src/main/java/bio/terra/tanagra/service/search/SqlVisitor.java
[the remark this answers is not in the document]
wchamber: Yes. I'll make sure to get to it before we're connecting to actual databases, but I'd like to finish the basics of query building first.

api/src/main/java/bio/terra/tanagra/service/search/SqlVisitor.java
wchamber: done.

api/src/test/java/bio/terra/tanagra/service/search/SqlVisitorTest.java
wchamber: Nothing in this test. It's roughly the name for the underlying dataset, which isn't used in these filters or expressions, but would be used in a table clause, e.g. [next PR](https://github.com/DataBiosphere/tanagra/pull/5/files#diff-3a2fe285e55d6e8eb1897fcb3c79cb33634c2762f60267437759e07a2b0cce75P). I'm imagining the underlay is the "name" of the "underlying dataset", where there may be one or more backing physical BQ datasets, e.g. indexes and initial dataset. All a little speculative at this point.
added a little more to comments on Entity.

PR #5, opened 2021-07-23
Add initial Query construct & SQL generation

wchamber:


api/src/main/java/bio/terra/tanagra/service/search/Query.java
```diff
@@ -0,0 +1,27 @@
...
+public abstract class Query {
+  /** What to select for the query. */
+  // TODO consider selections with a different entity than the primary entity.
+  public abstract ImmutableList<Selection> selections();
+
+  /** The primary entity being queried. */
+  public abstract Entity primaryEntity();
+
+  /** The filter to apply to the primary entity, if there is a filter. */
+  public abstract Optional<Filter> filter();
```
calbach, line 20: Looks like this is probably covered by the above TODO, but most of what the dataset builder (specifically via concept sets) is doing currently is applying filters to related entities. But first joined onto an initial cohort definition, i.e. set of person rows defined by a filter.

Not sure how this is modeled here.

api/src/main/java/bio/terra/tanagra/service/search/Selection.java
```diff
@@ -0,0 +1,67 @@
...
+
+import com.google.auto.value.AutoValue;
+import java.util.Optional;
+
+/**
+ * A construct in a query that represents a value to be selected.
+ *
+ * <p>Represents a single column in a SELECT SQL clause.
+ */
+public interface Selection {
```
calbach, line 11: Is there a clear path for "group by" in this model?

calbach (approved):
Looks good, just a few questions.

api/src/main/java/bio/terra/tanagra/service/search/Query.java
wchamber: Yep, very much TODO'd. right now I'm expecting it to be a subtype of filter.

api/src/main/java/bio/terra/tanagra/service/search/Selection.java
wchamber: Not yet. In many cases, I think we only care about any grouping by at the top most query level, but it's feasible to consider using group by's within a filter.

PR #6, opened 2021-07-28
Add variables for entities/attributes and a relationship filter for walking entity relationships

wchamber:
Introduces a variable concept for aliasing entities.
Introduces a relationship filter that works as a quantifier to bind a variable.

Naming/comment suggestions welcome.


api/src/main/java/bio/terra/tanagra/service/search/AttributeVariable.java
```diff
@@ -0,0 +1,25 @@
...
+
+import com.google.auto.value.AutoValue;
+
+/**
+ * An {@link Attribute} for a bound {@link Variable}.
+ *
+ * <p>e.g. "x.attribute" in "SELECT x.attribute FROM entity as x".
+ */
+@AutoValue
+public abstract class AttributeVariable {
```
calbach, line 11: Is "variable" used here as a Tanagra term, or a SQL term? I'd thought this maps more closely to something like a "select expression", though perhaps more narrowly constrained than that to support attributes.

api/src/main/java/bio/terra/tanagra/service/search/Variable.java
```diff
@@ -0,0 +1,13 @@
+package bio.terra.tanagra.service.search;
+
+import com.google.auto.value.AutoValue;
+
+/** A variable to alias an entity to allow referring to multiple aliases of the same entity. */
+@AutoValue
+public abstract class Variable {
```
calbach, line 7: Thoughts on Alias vs Variable for this class name? Or does this actually encompass both cases of normal references and aliases (if not, what is the value of `name` in the case where there's no alias, or does this model just not exist there)

api/src/main/java/bio/terra/tanagra/service/search/Filter.java
```diff
@@ -74,5 +76,66 @@ public static ArrayFunction create(List<Filter> operands, Operator operator) {
     }
   }
 
-  // TODO implement a relationship filter tht allows chaining to another entity.
+  /**
+   * A {@link Filter} that introduces a constraint that there must be a relationship to another
+   * entity.
+   *
+   * <p>This quantifiers introduces a new {@link Variable} and implicitly binds it to an entity in
```
calbach, line 83: grammar nit: "This quantifiers" I think is a typo, not sure what was intended though

api/src/main/java/bio/terra/tanagra/service/search/Filter.java
```diff
@@ -74,5 +76,66 @@ public static ArrayFunction create(List<Filter> operands, Operator operator) {
...
+   * {@link #boundAttribute()}.
+   *
+   * <p>This is equivalent to the pseudo-SQL ":outerAttribute IN (SELECT :boundAttribute FROM
+   * :boundAttribute.entity AS :boundAttribute.variable WHERE :filter".
```
calbach, line 87: nit: trailing paren

api/src/main/java/bio/terra/tanagra/service/search/Filter.java
```diff
@@ -74,5 +76,66 @@ public static ArrayFunction create(List<Filter> operands, Operator operator) {
...
+   *
+   * <p>Or in words ∃x∈boundEntity( x.boundAttribute = outerAttribute ∧ filter): there exists {@link
+   * #boundAttribute()} that's a member of {@link #boundAttribute()}'s Entity where {@link
```
calbach, line 90: I'm having a bit of difficulty following this, so I may be wrong, but should the first term on this line be `#boundEntity()` instead?

api/src/main/java/bio/terra/tanagra/service/search/Filter.java
```diff
@@ -74,5 +76,66 @@ public static ArrayFunction create(List<Filter> operands, Operator operator) {
...
+  @AutoValue
+  abstract class RelationshipFilter implements Filter {
+
+    /**
+     * The attribute of the newly introduced variable's entity that is equal to {@link
+     * #outerAttribute}.
+     */
+    public abstract AttributeVariable boundAttribute();
+
+    /** The attribute of an outer variable that's equal to {@link #boundAttribute}. */
```
calbach, line 102: Are there any restrictions on what entities are valid for this attribute? For example, if we have a heavily nested structure, can this attribute be defined multiple levels above this relationship filter, or just in the immediate surrounding context?

Is there any validation (implicit or explicit) here if I provide a reference to an entity which would not be valid in this position within the tree?

Things are getting a bit abstract so my terminology may be slightly off here.

api/src/test/java/bio/terra/tanagra/service/search/SqlVisitorTest.java
```diff
@@ -113,6 +132,22 @@ void filterArrayFunctionThrowsIfOperandsEmpty() {
         () -> Filter.ArrayFunction.create(ImmutableList.of(), Filter.ArrayFunction.Operator.AND));
   }
 
+  @Test
+  void filterRelationship() {
+    assertEquals(
+        "p.id IN (SELECT a.person_id FROM foo.address AS a WHERE a.zip_code = 12345)",
```
calbach, line 138: demonstrating a case where the outer variable is used within the filter may be good coverage to have as well

calbach (commented):
Looks good, these are mostly questions for my own understanding

api/src/main/java/bio/terra/tanagra/service/search/Variable.java
wchamber: I'm thinking of this in terms of https://en.wikipedia.org/wiki/Tuple_relational_calculus , where this is a tuple variable.
I think "variable" is a better fit. So far, variables always relate to a given entity, but having variables allows us to refer to multiple instances of the same entity in a given query - we need this for temporal queries, e.g. we want to relate 2 Visits x & y by their visit date ( x.visit_date = y.visit_date).
Later, I think variables will also be useful for aggregations, e.g. at least 5 procedures match some condition.
`WHERE (SELECT COUNT(procedure) as procedure_count FROM procedure WHERE ...) x
  AND x.procedure_count > 5`
So here x is binding to some "aggregated entity." but I'm not there yet.

But I'm really struggling to document this in a helpful way. Happy to talk more offline and iterate.

api/src/test/java/bio/terra/tanagra/service/search/SqlVisitorTest.java
wchamber: done.

api/src/main/java/bio/terra/tanagra/service/search/Filter.java
```diff
@@ -74,5 +76,67 @@ public static ArrayFunction create(List<Filter> operands, Operator operator) {
...
+      public abstract Builder outerAttribute(AttributeVariable outerAttribute);
+
+      public abstract AttributeVariable outerAttribute();
+
+      public abstract Builder filter(Filter filter);
+
+      public RelationshipFilter build() {
+        Preconditions.checkArgument(
+            !boundAttribute().variable().equals(outerAttribute().variable()),
+            "Cannot bind a new variable to the outer attribute's variable. ");
```
calbach, line 135: If I got this error, I might be stuck. Is there any guidance that can be provided in this exception as to how the user should correct their usage?

also nit: trailing space

api/src/main/java/bio/terra/tanagra/service/search/Filter.java
```diff
@@ -74,5 +76,67 @@ public static ArrayFunction create(List<Filter> operands, Operator operator) {
...
+   * <p>This introduces a new {@link Variable} and implicitly binds it to an entity in {@link
+   * #boundAttribute()}. The bound and related entities should have a relationship across the bound
+   * attributes. The outer variable may be any variable introduced higher in the outer filter scope.
+   *
+   * <p>This is equivalent to the pseudo-SQL ":outerAttribute IN (SELECT :boundAttribute FROM
+   * :boundAttribute.entity AS :boundAttribute.variable WHERE :filter)".
+   *
+   * <p>Or in words ∃x∈boundEntity( x.boundAttribute = outerAttribute ∧ filter): there exists an
+   * instance 'x' of {@link #boundAttribute()}'s Entity where x's {@link #boundAttribute()} equals
+   * {@link #outerAttribute()} and {@link #filter}.
```
calbach, line 92: Thanks for the updates. One thing I'm still a little bit confused by:

Here it says "boundAttribute equals outerAttribute and filter". I'm having difficulty following what it means for `boundAttribute` to "equal" a filter.

api/src/main/java/bio/terra/tanagra/service/search/Variable.java
calbach: This seems fine and probably correct, mostly just probing a bit for my own comprehension. Discussion not blocking, I'll add to agenda for when we next talk

calbach (approved):

api/src/main/java/bio/terra/tanagra/service/search/Filter.java
wchamber: parsing ambiguity, added some parens.

api/src/main/java/bio/terra/tanagra/service/search/Filter.java
wchamber: thanks, done.

api/src/main/java/bio/terra/tanagra/service/search/AttributeVariable.java
arychj: How about something like AttributeReference or ReferencedAttribute?

api/src/main/java/bio/terra/tanagra/service/search/Expression.java
```diff
@@ -35,13 +35,13 @@ public static Literal create(DataType dataType, String value) {
...
-  /** An {@link Expression} that's an {@link bio.terra.tanagra.service.search.Attribute} */
+  /** An {@link Expression} that's an {@link AttributeVariable}. */
   @AutoValue
   abstract class AttributeExpression implements Expression {
-    public abstract Attribute attribute();
+    public abstract AttributeVariable attributeVariable();
 
-    public static AttributeExpression create(Attribute attribute) {
-      return new AutoValue_Expression_AttributeExpression(attribute);
+    public static AttributeExpression create(AttributeVariable attributeVariable) {
```
arychj, line 43: Is it necessary for all `AttributeExpressions` to be explicitly aliased to variables, or can expression be created with just an attribute and have the variable be implicitly generated (similar to the previous behaviour 'person.height` -> `person_89ab.height`?)

wchamber: It's currently necessary in the code/API as written. It's not necessary to keep that requirement, and if it's useful for clients we can add support for that. But it was easiest to start without implementing implicit variables.

PR #7, opened 2021-07-30
Add an initial API schema for filter building

wchamber:
Adds the initial openapi schema for a dummy endpoint on entity filtering.
No implementation yet, let me know if you'd rather wait to see it working before adding the schema.

### DIFF
--- a/api/src/main/resources/api/service_openapi.yaml
+++ b/api/src/main/resources/api/service_openapi.yaml
@@ -20,8 +20,149 @@ paths:
         500:
           description: Service is broken
 
+  # TODO implement "real" export endpoints
+  '/entityFilter/sqlQuery':
+    post:
+      # TODO authorization
+      security: []
+      summary: Returns an SQL query for listing all of a Entity's primary keys.
+      operationId: entitySqlQuery
+      tags: [Entity]
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/EntityFilter'
+      responses:
+        200:
+          $ref: '#/components/responses/SqlQueryResponse'
+
 components:
+  responses:
+    SqlQueryResponse:
+      description: A raw SQL query.
+      content:
+        application/json:
+          schema:
+            $ref: '#/components/schemas/SqlQuery'
+
   schemas:
+    AttributeVariable:
+      description: An attribute of a variable for an entity.
+      type: object
+      required: [name, variable]
+      properties:
+        name:
+          description: The name of the attribute
+          type: string
+        variable:
+          description: The variable value
+          type: string
+
+    AttributeValue:
+      type: object
+      description: A possible value of an attribute. Exactly one field must be specified.
+      properties:
+        int64Val:
+          type: integer
+          format: int64
+        stringVal:
+          type: string
+        boolVal:
+          type: boolean
+
+    EntityFilter:
+      description: A filter to apply to the given entity.
+      type: object
+      properties:
+        entity:
+          description: The name of the entity
+          type: string
+        underlay:
+          description: The name of the underlay containing this entity.
+          type: string
+        entityVariable:
+          description: The variable to use for the top level entity in the filter.
+          type: string
+        filter:
+          $ref: '#/components/schemas/Filter'
+
+    SqlQuery:
+      type: object
+      description: A raw SQL query.
+      properties:
+        query:
+          type: string
+
+    # Filters
+    Filter:
+      type: object
+      description: |
+        Base Filter type that contains one-of any Filter subtype.
+        OpenApi inheritance and one-of code generation support are not great, so we roll our own.
+      properties:
+        arrayFilter:
+          $ref: '#/components/schemas/ArrayFilter'
+        binaryFilter:
+          $ref: '#/components/schemas/BinaryFilter'
+        relationshipFilter:
+          $ref: '#/components/schemas/RelationshipFilter'
+        # TODO allow aggregations in filters.
+
+    ArrayFilter:
+      type: object
+      description: A filter function for operating on N operands.
+      properties:
+        operands:
+          type: array
+          items:
+            $ref: '#/components/schemas/Filter'
+        operator:
+          $ref: '#/components/schemas/ArrayFilterOperator'
+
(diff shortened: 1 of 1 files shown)


api/src/main/resources/api/service_openapi.yaml
```diff
@@ -20,8 +20,149 @@ paths:
         500:
           description: Service is broken
 
+  # TODO implement "real" export endpoints
+  '/entityFilter/sqlQuery':
```
calbach, line 24: I like POST here because of the detailed body parameterization, and the fact that it's not a standard REST endpoint. Generally I like to suggest following this style guide: https://cloud.google.com/apis/design/custom_methods , though we've only really partially standardized on it in AoU (for example, we don't use `:` in our method names).

If you like that guide, and think it makes sense for Tanagra, then I would suggest something like `/.../generateSqlQuery`, i.e. including a verb. This can also make for a consistent operation ID that looks more like a method name `entitySqlQuery` -->  `generateSqlQuery`

Let me know if there are any other style guides etc that you have from PF that might be applying here.

api/src/main/resources/api/service_openapi.yaml
```diff
@@ -20,8 +20,149 @@ paths:
...
+        underlay:
+          description: The name of the underlay containing this entity.
+          type: string
+        entityVariable:
+          description: The variable to use for the top level entity in the filter.
+          type: string
+        filter:
+          $ref: '#/components/schemas/Filter'
+
+    SqlQuery:
```
calbach, line 91: Possibly not in this version yet, but should this give some indication of the specific database instance, or at least database type that it has been generated for?

Also, probably not for this version, do we need parameters here?

calbach (commented):
+1 to sending the API changes first in isolation - thanks. Understood details might change slightly when you get into implementation.

api/src/main/resources/api/service_openapi.yaml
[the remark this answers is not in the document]
wchamber: 1. Maybe. It very much depends on what the frontend prefers - shared resources or dynamicly retrieved from endpoints.
2. Probably around this same path, something like `:count` for counts, or `:ids`, or an "export file to google bucket" etc.
3. Totally up to product requirements. Sometimes the query, sometimes the results, I think. I'm thinking of the write to a Terra notebook as a "dataset export"
4. I don't think there's a lot, just counting, demographic counts, getting the ids, and generating queries. Eventually, with stateful artifacts, we'll need CRUD endpoints.

I'm not really familiar with http path best practices to really say. What do you think we should be doing here?
We could say this is about /entities/ and `filters` is a thing you can do on entities, along with `search` (which I think might have a few verbs under it). I don't know if dataset endpoints are also `entities` or just related enough.

We could do something like `/entities/{entityName}/filters:generateSqlQuery` or even `/underlays/{underlayName}/entities/{entityName}/filters:generateSqlQuery` but I don't know if that's an improvement.

wchamber: Thanks, that looks like a good style guide. I don't think we're using anything as perscriptive in PF.
PTAL and help me map to it.

api/src/main/resources/api/service_openapi.yaml
wchamber: Yep, added a few TODOs.
I'm not sure we keep this as a meaningful kind of "export"
I'm not sure that parameters help any, because if you don't trust the query you get from us, why should you trust that we have parameterized it correctly.

api/src/main/resources/api/service_openapi.yaml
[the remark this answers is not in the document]
calbach: I suspect that ultimately, the API could look something like this - assuming a world where Tanagra is hosting multiple datasets / dataset versions:

### Main query endpoints
POST /datasets/{dataset_id}/entities:count # raw count
POST /datasets/{dataset_id}/entities:aggregate # chart data
POST /datasets/{dataset_id}/entities:generateSqlQuery # this PR

### Meta-schema endpoints
GET /datasets/{dataset_id}/entities
GET /datasets/{dataset_id}/entities/{entity_id}
...

Possibly more CRUD for the schema, if we wanted to support admin APIs.

Some of this might depend on what nouns eventually emerge for the persistent storage, e.g. if we're calling these things "cohorts" or something else.

calbach: Yep, I think this is already conformant from what I can see - the only discrepancy is the hierarchy per other comment. Would typically expect to see the pattern of `/noun1s/{noun1_id}/noun2s/{noun2_id}` etc. FOr now I might just flatten this to `/entities:generateSqlQuery`, if that seems like the most relevant noun

calbach (approved):

api/src/main/resources/api/service_openapi.yaml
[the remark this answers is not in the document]
wchamber: Thanks, sounds good.
So far I'm going with `underlay` instead of `datasets` following a brainstomr suggestion from the design doc. I'm not attached to that, but I'd like to avoid using "dataset" which is a very overloaded term.

api/src/main/resources/api/service_openapi.yaml
```diff
@@ -20,8 +20,166 @@ paths:
         500:
           description: Service is broken
 
+  # TODO implement "real" export endpoints
+  '/underlays/{underlay}/entities/{entity}/filters:generateSqlQuery':
```
calbach, line 24: Will "filters" be a real noun in the API, aside from being a model that's passed around in requests? If not - then probably unnecessary in the URL

calbach (approved):

api/src/main/resources/api/service_openapi.yaml
wchamber: So far, I'm thinking of keeping it as a replacement for 'cohorts' and 'concept sets.' but if not, then some other noun.

calbach: I see - in this case, the `entities/{entity_id}` infix is probably unnecessary. While the primary entity is an attribute and key trait of a cohort / filter, it is not really a "parent" in my view - it also will typically reference many other entities.

PR #8, opened 2021-07-30
Add protobufs for an underlay schema

wchamber:
These protobufs are the initial pass defining what an underlay schema consists of.
Not included, a schemas for hierachical query support or indexes.

Let me know if you'd rather wait to see this being used.


api/src/main/proto/bio/terra/tanagra/underlay/mapping.proto
```diff
@@ -0,0 +1,74 @@
...
+syntax = "proto3";
+
+package bio.terra.tanagra.underlay;
+option java_package = "bio.terra.tanagra.proto.underlay";
+option java_multiple_files = true;
+
+import "bio/terra/tanagra/underlay/entities.proto";
+import "bio/terra/tanagra/underlay/tables.proto";
+
+// An id to uniquely idnetify a column of a table within an underlay.
```
calbach, line 12: identify

api/src/main/proto/bio/terra/tanagra/underlay/mapping.proto
```diff
@@ -0,0 +1,74 @@
+// Messages mapping logical entities to physical tables
```
calbach, line 1: How should I be reading this: does this mapping work for BigQuery datasets only, or is this meant to support different backends? If it is specific to bigquery (and we might one day support others), should this be indicated in the package structure somewhere?

**Update**: Reading on - it seems this is meant to be generic across potentially different backends - will review with this in mind.

api/src/main/proto/bio/terra/tanagra/underlay/mapping.proto
```diff
@@ -0,0 +1,74 @@
...
+option java_package = "bio.terra.tanagra.proto.underlay";
+option java_multiple_files = true;
+
+import "bio/terra/tanagra/underlay/entities.proto";
+import "bio/terra/tanagra/underlay/tables.proto";
+
+// An id to uniquely idnetify a column of a table within an underlay.
+message ColumnId {
+  // The name of the dataset.
+  string dataset = 1;
```
calbach, line 15: opt: A more portable term for this is probably `schema`. I don't really love this term though, so I'm fine leaving dataset, even if it might be slightly overfitting to BigQuery terminology (AFAIK)

api/src/main/proto/bio/terra/tanagra/underlay/entities.proto
```diff
@@ -0,0 +1,36 @@
...
+  // The type of data for the attribute.
+  DataType data_type = 2;
+}
+
+// A relationship between 2 entities.
+message Relationship {
+  // The name of the relationship.
+  string name = 1;
+  // The name of the first entity.
+  string entity1 = 2;
```
calbach, line 33: Is the order of entity1 vs entity2 meaningful, e.g. directional? Do we presume many:many here, or do we not care?

If the order matters, may want to consider renaming

api/src/main/proto/bio/terra/tanagra/underlay/underlay.proto
```diff
@@ -0,0 +1,32 @@
...
+import "bio/terra/tanagra/underlay/mapping.proto";
+import "bio/terra/tanagra/underlay/tables.proto";
+
+// The description of an underlay.
+message Underlay {
+  // The name of the underlay.
+  string name = 1;
+
+  // The entities that make up this underlay.
+  repeated Entity entities = 2;
```
calbach, line 19: Let me know if I got this correct:

- Entities/relationships: define an abstract data model using Tanagra's generic schema
- Datasets: define the schema of the source dataset
- Mappings: define the translation layer between the source schema and Tanagra's generic schema

As a developer who is trying to add a new underlay, the following might hold:
- Entities/relationships: copy or fork a similar-looking Tanagra dataset. Likely there are standard examples here for human studies, but you may need to pick/choose entities depending on data availability
- Datasets: should be straightforward for Tanagra to provide a tool to autogenerate this structure from a given BigQuery dataset
- Mappings: most of the manual work; of course Tanagra can also provide standard examples / mappings for common standard data models like OMOP

calbach (approved):
Looks good, thanks.

api/src/main/proto/bio/terra/tanagra/underlay/mapping.proto
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// How to map an entity.
+message EntityMapping {
+  // The name of the entity being mapped.
+  string entity = 1;
+
+  // The primary key for the entity. This is used to uniquely identify an entity
+  // and the corresponding table is treated as the entity's primary table.
+  ColumnId primaryKey = 2;
```
calbach, line 37: primary_key

api/src/main/proto/bio/terra/tanagra/underlay/entities.proto
wchamber: So far, I'm not sure that we care. Subject to change; I could imagine us having a RelationshipType { one_to_one, one_to_many, many_to_many}.

api/src/main/proto/bio/terra/tanagra/underlay/mapping.proto
wchamber: Yep, added more to this comment.

api/src/main/proto/bio/terra/tanagra/underlay/mapping.proto
wchamber: It matches for Postgres & MySQL as well, so sticking with it for now.

api/src/main/proto/bio/terra/tanagra/underlay/underlay.proto
wchamber: Yep, you nailed it. Stealing your summaries to add to the file comments.
Eventually, I'll try to distill this in a README[s].

PR #9, opened 2021-08-02
Add initial Underlay package support

wchamber:
- Add java classes for physical database schemas
- Add Underlay for access underlay schema
- Add initial code to convert from protos to java classes
- Add configuration & service for initial support of loading underlays 
- Create an artificial "nautical" underlay schema for testing.
- Add more quality suppressions for NPathComplexity. That's a matter for code review, not static analysis. (xml diff is totally useless, don't understand why)

Sorry for the big chunk of lines, but a lot of it is pretty mechanical.


api/src/test/resources/underlays/nautical.pbtext
```diff
@@ -0,0 +1,305 @@
+# proto-file: proto/bio/terra/tanagra/underlay/underlay.proto
```
calbach, line 1: Given the length of this simple example, it seems possible that we may eventually want to support splitting across multiple files

calbach (commented):
Nice tests. Looks good overall - I probably need to give it another pass over after this round of comments - a lot to cover here

api/src/main/java/bio/terra/tanagra/service/underlay/UnderlayConversion.java
```diff
@@ -0,0 +1,169 @@
...
+    return Underlay.builder()
+        .name(underlayProto.getName())
+        .entities(entities)
+        .attributes(attributes)
+        .primaryKeys(primaryKeys)
+        .simpleAttributesToColumns(simpleAttributesToColumns)
+        .build();
+  }
+
+  private static Entity convert(
```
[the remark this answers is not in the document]
wchamber: It's a little interesting, but I'd like to hear more to be sold on the benefit for the mental overhead.
This mapping code *is* boilerplate - kind of verbose, but not hard or expensive to write or maintain. The mapstruct stuff looks a little more succinct, but I'm not sure it's a 10x time saver. I can't imagine not wanting tests for conversions, so I'm not sure I'd buy there being much benefit in terms of bugs.
 As soon as you have to do a nontrivial transformation, the plain java seems to me to be much easier to read.

api/src/test/resources/underlays/nautical.pbtext
wchamber: Yep. I could imagine splitting on the same lines as the protos (entities/tables/mapping), but at the same time I'd like to wait to have more experience with building/generating these and thinking about tooling we might like to have before getting too much into ergonomics.

api/src/main/java/bio/terra/tanagra/service/underlay/Underlay.java
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * An underlay dataset used to power a Tanagra experience.
+ *
+ * <p>Each underlying dataset that a user wants to explore with Tanagra is represented by an
+ * "underlay." That underlay represents what logical entities are being modeled, what physical
+ * tables and columns are in the underlying dataset, and the mapping between Tanagra concepts like
+ * entities and searches and the physical datasets.
+ *
+ * <p>An Underlay instance is what powers a Tanagra search experience for an extenral backing
```
calbach, line 23:
```suggestion
 * <p>An Underlay instance is what powers a Tanagra search experience for an external backing
```

api/src/main/java/bio/terra/tanagra/service/underlay/UnderlayConversion.java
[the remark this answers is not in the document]
calbach: From my experience, the primary value adds are (1) generation-time exhaustive mapping enforcement (similar to enum mapper, but more general and earlier), i.e. when you add a new field and forget to map it, things should break immediately and (2) relatively succinct nested object null-handling, which can become very verbose in these hand-crafted converters, though that class of issues can be avoided somewhat with a clean interface design. And yeah - you would still want tests.

Of course, it definitely needs to be weighed against comprehension and magic codegen cost

api/src/test/resources/underlays/nautical.pbtext
calbach, line 1: Another thought: yaml may be a more accessible / common config format here. YAML can be converted to JSON, which can be converted to proto using the proto3 JSON encoding, so in theory that pathway could also work with little additional effort.

calbach (approved):

api/src/test/resources/underlays/nautical.pbtext
wchamber: Added an equivalent yaml file and configuration property to allow it.

calbach: Thanks! Is it pushed? I'm not seeing it in the branch

wchamber: Merge error, ptal now!

api/src/test/resources/underlays/nautical.yaml
```diff
@@ -0,0 +1,189 @@
+# An Underlay yaml file describing a fictional relational domain of sailors, boats and
```
calbach, line 1: This is great, thanks. I think this is a lot more approachable

calbach (approved):

PR #10, opened 2021-08-06
Refactor relationship filters

wchamber:
Refactor relationship filters to be interms of entities instead of attributes. Rely on underlay resolution to find the entities to use for creating queries. This better reflects the logical model and will let the API operate in terms of entities without knowing what attributes are used to relate entities.
Refactor the UnderlaySqlResolver to make it a direct part of SqlVisitor with a dependency on the Underlay. This keeps the particulars of SQL generation to the appropriate search entity.


api/src/test/resources/underlays/nautical.pbtext
```diff
@@ -299,7 +299,7 @@ relationship_mappings: {
     }
     foreign_key: {
       dataset: "nautical"
-      table: "boats"
+      table: "reservations"
```
calbach, line 302: did the new validation catch this?

calbach (approved):

api/src/test/resources/underlays/nautical.pbtext
wchamber: yep!

api/src/main/java/bio/terra/tanagra/service/underlay/Underlay.java
```diff
@@ -27,65 +27,50 @@
...
+        relationships().values().stream()
+            .filter(
+                relationship ->
+                    (relationship.entity1().equals(x) && relationship.entity2().equals(y))
+                        || (relationship.entity1().equals(y) && relationship.entity2().equals(x)))
+            .collect(Collectors.toList());
+    if (matching.isEmpty()) {
+      return Optional.empty();
+    }
+    // TODO consider adding relationship names if we need to support multiple relationships between
```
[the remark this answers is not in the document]
wchamber: Good to know, thanks. I'll leave as a TODO for now, but will circle back.

PR #11, opened 2021-08-06
Add conversions from API models to internal classes

wchamber:
To be wired up to controllers and services in a follow up PR.

Also fix the github workflow to apply quality checks to the test source set.


api/src/main/java/bio/terra/tanagra/service/query/api/ConversionUtils.java
```diff
@@ -0,0 +1,23 @@
...
+  /** Returns if exactly one of the objects is non-null. */
+  public static boolean exactlyOneNonNull(Object... objects) {
+    return Stream.of(objects).filter(Objects::nonNull).count() == 1;
+  }
+
+  public static Variable createVariable(String name) {
+    if (name.isEmpty()) {
+      throw new BadRequestException("A variable must have a non-empty name.");
+    }
+    return Variable.create(name);
```
calbach, line 21: nit: why this utility, rather than just adding this validation to `Variable.create` ? Reading further, it looks like you also did that. Do you need this util then?

api/src/main/java/bio/terra/tanagra/service/query/api/FilterConverter.java
```diff
@@ -0,0 +1,139 @@
...
+      return convert(apiFilter.getBinaryFilter(), scope);
+    } else if (apiFilter.getRelationshipFilter() != null) {
+      return convert(apiFilter.getRelationshipFilter(), scope);
+    }
+    throw new BadRequestException("Unknown Filter type: " + apiFilter.toString());
+  }
+
+  @VisibleForTesting
+  Filter.BinaryFunction convert(ApiBinaryFilter apiBinary, VariableScope scope) {
+    Preconditions.checkNotNull(
```
calbach, line 52: should these be `BadRequestException` as well? This is basically request validation which should return a 400 - right?

api/src/main/java/bio/terra/tanagra/service/query/api/VariableScope.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+class VariableScope {
+
+  /** The VariableScope that encloses this, or null if there is none. */
+  @Nullable private final VariableScope enclosing;
+
+  /** The variable names to {@link EntityVariable}s bound at this scope. */
+  private final Map<String, EntityVariable> variables = new HashMap<>();
+
+  VariableScope(VariableScope enclosing) {
```
calbach, line 25: nit: `@Nullable`?

api/src/test/java/bio/terra/tanagra/service/query/api/VariableScopeTest.java
```diff
@@ -0,0 +1,36 @@
...
+public class VariableScopeTest {
+  private static final Entity FOO = Entity.builder().underlay("u").name("foo").build();
+  private static final Entity BAR = Entity.builder().underlay("u").name("bar").build();
+
+  private static final EntityVariable X_FOO = EntityVariable.create(FOO, Variable.create("x"));
+  private static final EntityVariable Y_FOO = EntityVariable.create(FOO, Variable.create("y"));
+  private static final EntityVariable Z_FOO = EntityVariable.create(FOO, Variable.create("z"));
+
+  @Test
+  void scope() {
```
calbach, line 22: May be worth covering a variable shadowing case here to at least fix the semantics. I'm not sure whether or not it should be disallowed.

api/src/main/java/bio/terra/tanagra/service/query/api/FilterConverter.java
wchamber: Yep, thanks.

api/src/main/java/bio/terra/tanagra/service/query/api/VariableScope.java
wchamber: done

api/src/test/java/bio/terra/tanagra/service/query/api/VariableScopeTest.java
wchamber: Thanks, yep, added a couple tests.

wchamber:
cc @arychj @freemabd

api/src/main/java/bio/terra/tanagra/service/query/api/ConversionUtils.java
calbach: Got it. I think that distinction is probably correct - I just missed it. It's correct for it to come from the conversion layer.

calbach (approved):

PR #13, opened 2021-08-09
Add basic container building & publishing with Jib & GH actions

wchamber:
Do the easiest first pass on creating a docker container for the service with Jib. Use a simple manual github action to publish images versions to GCR.

Initial image at gcr.io/broad-dsp-gcr-public/tanagra-api

### DIFF
--- a/.github/workflows/api-publish.yaml
+++ b/.github/workflows/api-publish.yaml
@@ -11,7 +11,7 @@ on:
 
 env:
   REGISTRY: ghcr.io
-  IMAGE_NAME: tanagra-api:${{ github.event.inputs.version }}
+  IMAGE_NAME: ${{ env.REGISTRY }}/DataBiosphere/tanagra-api:${{ github.event.inputs.version }}
 
 jobs:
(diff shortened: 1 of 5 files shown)


.github/workflows/api-publish.yaml
```diff
@@ -1,14 +1,18 @@
-# Dummy github action to allow iteration.
+# This workflow builds a docker image with jib and pushes it to Github Container Repository.
```
calbach, line 1: GCR == Google Container Registry, i.e. gcr.io. I was surprised when you mentioned pushing containers to GitHub, but perhaps that was a mixup.

Apparently there is also a GitHub container registry @ `ghcr.io`, though I've never actually seen an image from here in the wild.

wchamber: Oops, stale comment. I looked at the github container registry, but DataBiosphere doesn't currently allow anyone to publish public packages there, so went with GCR.

calbach (approved):
Ah, that explains it, thanks.

PR #14, opened 2021-08-09
Add an intial endpoint OpenApi schema for creating SQL for datasets

wchamber:
Start with a simple API for creating SQL queries for datasets.
Initially, this only supports concept sets by having the client compose the filter relationships between entities. I'm not sure that will hold, but it's the simplest place to start. (e.g. to select a 'Conditions' concept set on a 'Person' cohort, the 'Person' entity filter needs to be added as a relationship filter on the 'Conditions' entity.) This also doesn't create an object for the selected attributes list, which may also need to change.

### DIFF
--- a/api/src/main/resources/api/service_openapi.yaml
+++ b/api/src/main/resources/api/service_openapi.yaml
@@ -36,7 +36,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/EntityDataset'
+              $ref: '#/components/schemas/GenerateSqlQueryDatasetRequest'
         responses:
           200:
             $ref: '#/components/responses/SqlQueryResponse'
@@ -119,10 +119,10 @@ components:
       type: object
       properties:
         entityVariable:
-          description: The variable to use for the top level entity in the filter.
+          description: The variable to use for the primary entity in the filter.
           type: string
         attributes:
-          description: The name of the attributes to select for the dataset.
+          description: The name of the attributes of the primary entity to select for the dataset.
           type: array
           items:
             type: string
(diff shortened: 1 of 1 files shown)


wchamber:
cc @freemabd @arychj

api/src/main/resources/api/service_openapi.yaml
```diff
@@ -20,6 +20,28 @@ paths:
...
+      security: [ ]
+      summary: Returns an SQL query for selecting a dataset of an Entity's attributes.
+      operationId: generateSqlQuery
+      tags: [EntitiesDataset]
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/EntityDataset'
```
calbach, line 39: In this case, I would probably make a dedicated request schema for this method for extensibility. Usually I would just make the request/responses parallel, e.g. `GenerateSqlQueryRequest`

api/src/main/resources/api/service_openapi.yaml
```diff
@@ -92,6 +114,21 @@ components:
...
 
+    EntityDataset:
+      description: A selection of entity attributes and filters on them.
+      type: object
+      properties:
+        entityVariable:
+          description: The variable to use for the top level entity in the filter.
+          type: string
+        attributes:
+          description: The name of the attributes to select for the dataset.
```
calbach, line 125: to be clear, these need to be attributes of the primary entity, right? If not, how would I specify a foreign table attribute?

api/src/main/resources/api/service_openapi.yaml
wchamber: Good idea, thanks. Went with `GenerateSqlQueryDatasetRequest` since we'll have a few amount of sql generations.

api/src/main/resources/api/service_openapi.yaml
wchamber: Yes, primary entity. added more to the comment.

calbach (approved):

PR #15, opened 2021-08-09
Add get and list OpenAPI endpoints for underlays and entities.

wchamber:
Some initial standard REST endpoints for underlays and entities.
Include pagination parameters.
Will add more "get me the udnerlay metadata" endpoints for entity filters in another PR.

### DIFF
--- a/api/src/main/resources/api/service_openapi.yaml
+++ b/api/src/main/resources/api/service_openapi.yaml
@@ -376,7 +376,6 @@ components:
         entityDataset:
           $ref: '#/components/schemas/EntityDataset'
 
-
     NextPageToken:
       type: string
       description: |
--- a/api/src/main/resources/api/service_openapi.yaml
+++ b/api/src/main/resources/api/service_openapi.yaml
@@ -22,6 +22,9 @@ paths:
 
   '/underlays':
     get:
+      parameters:
+        - $ref: '#/components/parameters/PageSize'
+        - $ref: '#/components/parameters/PageToken'
       # TODO authorization
       security: []
       summary: Lists the underlays.
@@ -46,6 +49,8 @@ paths:
 
   '/underlays/{underlayName}/entities':
     parameters:
+      - $ref: '#/components/parameters/PageSize'
+      - $ref: '#/components/parameters/PageToken'
       - $ref: '#/components/parameters/Underlay'
     get:
       # TODO authorization
@@ -110,6 +115,24 @@ components:
       schema:
         type: string
 
+    PageSize:
+      name: pageSize
+      in: query
+      description: |
+        The maximum number of results to return in this page. The server may return less.
+        If 0 or unspecified, the server will decide the number of results.
+      schema:
+        type: integer
+        minimum: 0
+        default: 0
+
+    PageToken:
+      name: pageToken
+      in: query
+      description: An opaque pagination token to retrieve a specific page of results
+      schema:
+        type: string
+
   responses:
     GetEntityResponse:
       description: Response to get an entities.
@@ -130,14 +153,14 @@ components:
       content:
         application/json:
           schema:
-            $ref: '#/components/schemas/EntitiesList'
+            $ref: '#/components/schemas/ListEntitiesResponse'
 
     ListUnderlaysResponse:
       description: Response to list the underlays.
       content:
         application/json:
           schema:
-            $ref: '#/components/schemas/UnderlaysList'
+            $ref: '#/components/schemas/ListUnderlaysResponse'
 
     SqlQueryResponse:
       description: A raw SQL query.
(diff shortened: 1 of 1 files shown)


wchamber:
cc @freemabd @arychj

api/src/main/resources/api/service_openapi.yaml
```diff
@@ -20,6 +20,57 @@ paths:
...
+      # TODO authorization
+      security: []
+      summary: Lists the underlays.
+      operationId: listUnderlays
+      tags: [ Underlays ]
+      responses:
+        200:
+          $ref: '#/components/responses/ListUnderlaysResponse'
+
+  '/underlays/{underlay}':
```
calbach, line 34: nit: usually I would call this parameter either an `underlayId` or `underlayName`

api/src/main/resources/api/service_openapi.yaml
```diff
@@ -20,6 +20,57 @@ paths:
...
+      - $ref: '#/components/parameters/Underlay'
+    get:
+      # TODO authorization
+      security: []
+      summary: Get info on an underlay.
+      operationId: getUnderlay
+      tags: [ Underlays ]
+      responses:
+        200:
+          $ref: '#/components/responses/GetUnderlayResponse'
```
calbach, line 45: Per https://cloud.google.com/apis/design/standard_methods#get  - I recommend returning the entity here rather than a wrapper. In my experience this has worked: if you find yourself having to include other things in the response object (aside from the resource), then probably you've deviated from a standard REST method.

Here and below for GetEntity

api/src/main/resources/api/service_openapi.yaml
```diff
@@ -92,6 +181,24 @@ components:
...
+    DataType:
+      description: Enum of data types
+      type: string
+      enum: ['INT64', 'STRING', 'BOOL']
+
+    Entity:
+      type: object
+      description: A relational logical entity supported by an Underlay.
+      properties:
+        name:
```
calbach, line 193: Probably not for this PR: I'm wondering whether entity name and display name will always be the same, or if we'll need to split them out. I guess it will largely depend on how dynamic the frontend is.

Given you're including this in the API now, we will likely need to enforce some restrictions around them (length, characters)

wchamber: Yeah. I can see us having a display name or description for rendering, but having a short meaningful entity name instead of an opaque id seems useful.
I'll add entity name restrictions in a follow up PR.

api/src/main/resources/api/service_openapi.yaml
wchamber: I thought that's what I was doing. Don't I need the intervening "responses" object with `content: application/json:` around the schema? I could inline the `GetUnderlayResponse` object, I don't really know if there's a good reason to prefer or avoid that.

calbach: Sorry, I misread this - I'm less accustomed to OpenAPI 3. As long as the response Entity is not wrapped - we should be good.

api/src/main/resources/api/service_openapi.yaml
```diff
@@ -60,6 +111,34 @@ components:
...
+        application/json:
+          schema:
+            $ref: '#/components/schemas/Underlay'
+
+    ListEntitiesResponse:
+      description: Response to list the entities.
+      content:
+        application/json:
+          schema:
+            $ref: '#/components/schemas/EntitiesList'
```
calbach, line 133: I feel obligated to link this: https://cloud.google.com/apis/design/design_patterns#list_pagination

However, I don't know that I would go through the trouble for these metadata APIs

calbach (approved):

api/src/main/resources/api/service_openapi.yaml
wchamber: Oh, good point. Hmmm, yeah it seems a little overkill, but consistency is a good argument. Eh, I'll start with it in and if implementation time it doesn't seem worth it I'll rip it out again.
PTAL

api/src/main/resources/api/service_openapi.yaml
```diff
@@ -139,6 +269,17 @@ components:
...
 
+    ListEntitiesResponse:
+      type: object
+      description: A response for a list of Entities.
+      properties:
+        entities:
+          type: array
+          items:
+            $ref: '#/components/schemas/Entity'
+        next_page_token:
```
calbach, line 280: camel case

api/src/main/resources/api/service_openapi.yaml
```diff
@@ -235,3 +375,33 @@ components:
...
+
+    ListUnderlaysResponse:
+      type: object
+      description: A list of Underlays.
+      properties:
+        underlays:
+          type: array
+          items:
+            $ref: '#/components/schemas/Underlay'
+        next_page_token:
```
calbach, line 406: camel

calbach (approved):

PR #17, opened 2021-08-12
Add maven publishing for Tanagra service

wchamber:
gradle tasks for maven publishing and pushing to artifactory.
Add a github workflow for pushing to the Broad's artifactory.
Remove github action permissions on api-publish since we're not using GHCR and they're unnecessary.

I'll publish the first package after this is merged and I can test the github action. Will add fixes as needed.


calbach (approved):
Looks good

PR #18, opened 2021-08-12
Add some naming restrictions for Tanagra objects

wchamber:
Add some simple name validation restrictions for Entities, Attributes, and Relationships. We need to refere to these names in APIs and probably don't want to allow anything of any length.

### DIFF
--- a/api/src/main/java/bio/terra/tanagra/service/search/NameUtils.java
+++ b/api/src/main/java/bio/terra/tanagra/service/search/NameUtils.java
@@ -1,0 +1,33 @@
+package bio.terra.tanagra.service.search;
+
+import com.google.common.base.Preconditions;
+import java.util.regex.Pattern;
+
+/**
+ * Utilities for checking Tanagra names.
+ *
+ * <p>Some names traverse Tanagra from HTTP requests to SQL generation. For these names, it's
+ * important that they are relatively friendly and simple.
+ */
+final class NameUtils {
+  private NameUtils() {}
+
+  // Start with simple lower case letters, numbers, and underscores of less than 32 characters.
+  // We could add more valid characters, or allow different characters for different places, but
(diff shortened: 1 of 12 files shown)


api/src/main/java/bio/terra/tanagra/service/search/NameUtils.java
```diff
@@ -0,0 +1,33 @@
...
+import com.google.common.base.Preconditions;
+import java.util.regex.Pattern;
+
+/**
+ * Utilities for checking Tanagra names.
+ *
+ * <p>Some names traverse Tanagra from HTTP requests to SQL generation. For these names, it's
+ * important that they are relatively friendly and simple.
+ */
+final class NameUtils {
```
calbach, line 12: Should this be applied to variables?

calbach (approved):
I like it, thanks.

api/src/main/java/bio/terra/tanagra/service/search/NameUtils.java
wchamber: Maybe. Variables are only user specified and are used directly in SQL, so they're a little differnt. OTOH, having a consistent simple naming schema might be valuable. I could go either way, WDYT?

calbach: I can definitely see them needing to diverge for that reason (or else necessitating a conversion layer for Tanagra variable -> SQL variable). I'd keep them semantically separate, e.g. in API documentation - but your call whether you want to just reuse this code, or split out a different utility.

wchamber: reverting variable to keep it separate for now.

calbach (approved):

PR #20, opened 2021-08-13
Add support for attributes based on lookup columns

wchamber:
Support attributes that are related normalized with a lookup table by id reference.
Add an AttributeMapping interface to support the 2 types of mappings so far: simple columns and lookup columns.
Add an example "boat type" fact table for boats for tests.


api/src/main/proto/bio/terra/tanagra/underlay/mapping.proto
```diff
@@ -51,9 +51,23 @@ message AttributeMapping {
     ColumnId column_id = 1;
   }
 
+  // A mapping where an attribute refers to a normalized column in a fact table
```
calbach, line 54: Can you expand on the terminology "fact" here? I'm not quite following this. What makes this a fact, vs just a related table?

wchamber: "fact" probably wasn't right. How about "lookup"? Tried to add some more comments, PTAL.

calbach (approved):
Thanks, this wording makes more sense to me

PR #21, opened 2021-08-13
Create an initial OMOP underlay for development testing.

wchamber:
Use an existing dataset from `all-of-us-ehr-dev` to start.


api/src/main/resources/underlays/omop_test.yaml
```diff
@@ -0,0 +1,437 @@
...
+      - name: standard_concept
+        dataType: STRING
+      - name: concept_code
+        dataType: INT64
+relationships:
+  - name: person_condition_occurrence
+    entity1: person
+    entity2: condition_occurrence
+  - name: person_gender_concept_id
+    entity1: concept
```
calbach, line 60: Hmm I'm surprised by the inversion of order on this, as compared to the first entry. Is this meaningful? If not, maybe just a convention to mirror the order in the name could make sense

api/src/main/resources/underlays/omop_test.yaml
```diff
@@ -0,0 +1,437 @@
+# An initial OMOP underlay schema for development.
+
+name: omop_test
+entities:
```
calbach, line 4: @freemabd would be curious to get your feedback on this

calbach (commented):
Nice! This is a big step towards the initial milestone

api/src/main/resources/underlays/omop_test.yaml
freemabd: This seems logical but very tedious and easy to make a mistake. I assuming this was this done manually? Could this be automated in some way using some sort of Entity object and the BQ OMOP schema files? Would it also make sense to break these into separate yaml files? This file could get pretty large.

api/src/main/resources/underlays/omop_test.yaml
```diff
@@ -0,0 +1,437 @@
...
+  - entity: condition_occurrence
+    primaryKey:
+      dataset: omop_test
+      table: condition_occurrence
+      column: condition_occurrence_id
+  - entity: concept
+    primaryKey:
+      dataset: omop_test
+      table: concept
+      column: concept_id
```
freemabd, line 135: Does an entityMapping only define the primaryKey? If so could this be specified in the entity itself?

api/src/main/resources/underlays/omop_test.yaml
wchamber: This was done manually for this v0.
We could write some pretty simple code to take BQ schema and make the dataset/table definitions as well as the simple column attributes for files. What doesn't work was well is things like person.gender, where that's not a direct column on the person table (person.gender_concept_id is). We can develop more tools as we scale this out to make it less manual. There's already checking in place to find errors.
We can split things into multiple yaml files in a later iteration too.

api/src/main/resources/underlays/omop_test.yaml
wchamber: right now it matters for the related `relationshipMapping` where the entity1 is expected to be the entity twith the table as the primary key. Did some renaming.

api/src/main/resources/underlays/omop_test.yaml
wchamber: Right now it's the only thing it defines. We could change the structure of more of this to include the mappings within the logical relationships - right now the logical entities/attributes/relationships and their related mappings are specified separately, but we could have a nested struture instead.

api/src/main/resources/underlays/omop_test.yaml
calbach: Got it. may be worth renaming from entity1 and entity2 if these have semantics (probably not in this PR), maybe something like entity and foreignEntity?

calbach (approved):

api/src/main/resources/underlays/omop_test.yaml
wchamber: SG, will do in a follow up PR

PR #23, opened 2021-08-16
Implement get & list underlays endpoints.

wchamber:
Add a simple paginator implementation for in-memory lists.

### DIFF
--- a/api/src/main/java/bio/terra/tanagra/common/Paginator.java
+++ b/api/src/main/java/bio/terra/tanagra/common/Paginator.java
@@ -1,0 +1,85 @@
+package bio.terra.tanagra.common;
+
+import bio.terra.common.exception.BadRequestException;
+import bio.terra.tanagra.proto.utils.pagination.PageSizeOffset;
+import com.google.auto.value.AutoValue;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+import com.google.protobuf.InvalidProtocolBufferException;
+import java.util.Base64;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/** Implements simple pagination for sorted lists in memory. */
+public final class Paginator<T> {
+  private Paginator() {}
+
+  /**
+   * Returns a {@link Page} for the results.
+   *
+   * @param allResults the sorted list of all results being paginated.
+   * @param pageSize the maximumn number of results to have per page
+   * @param pageToken the token of the page to retrieve. If null or empty, the first page is
+   *     retrieved.
+   */
+  public static <T> Page<T> getPage(List<T> allResults, int pageSize, @Nullable String pageToken) {
+    Preconditions.checkArgument(pageSize > 0, "pageSize must be greater than 0 to paginate.");
+    if (pageToken == null || pageToken.isEmpty()) {
+      return getPageInternal(
+          allResults, PageSizeOffset.newBuilder().setOffest(0).setPageSize(pageSize).build());
+    }
+    PageSizeOffset pageSizeOffset = decode(pageToken);
+    if (pageSizeOffset.getPageSize() != pageSize) {
+      throw new BadRequestException(
+          String.format(
+              "Token page size '%d' does not match input page size '%d'.",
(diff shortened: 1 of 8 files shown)


api/src/main/proto/bio/terra/tanagra/utils/pagination.proto
```diff
@@ -0,0 +1,17 @@
...
+package bio.terra.tanagra.utils;
+option java_package = "bio.terra.tanagra.proto.utils.pagination";
+option java_multiple_files = true;
+
+// A simple pagination message with the page size and offset for the page.
+message PageSizeOffset {
+  // The number of results to return per page.
+  int32 page_size = 1;
+  // The offset into the result for the page.
+  int32 offest = 2;
```
calbach, line 16: offset

api/src/main/java/bio/terra/tanagra/common/Paginator.java
```diff
@@ -0,0 +1,82 @@
...
+    PageSizeOffset nextPageSizeOffset = pageSizeOffset.toBuilder().setOffest(pageLimit).build();
+    List<T> results = allResults.subList(offset, pageLimit);
+    return Page.create(results, encode(nextPageSizeOffset));
+  }
+
+  private static String encode(PageSizeOffset pageSizeOffset) {
+    return Base64.getEncoder().encodeToString(pageSizeOffset.toByteArray());
+  }
+
+  private static PageSizeOffset decode(String pageToken) {
```
calbach, line 61: FYI: not relevant yet, since you don't really have any parameterization on your list endpoint, but one pattern I've used in the past:

- take a short hash of the request (sans pageSize, pageToken)
- add this to the page token payload
- validate this against incoming requests that include a token

This allows you to reject requests that break the pagination contract - i.e. attempt to intentionally or accidentally change filter parameters (e.g. for listEntities, say they change the underlayName) in the middle of a pagination stream.

api/src/main/java/bio/terra/tanagra/common/Paginator.java
```diff
@@ -0,0 +1,82 @@
...
+    if (pageToken == null || pageToken.isEmpty()) {
+      return getPageInternal(
+          allResults, PageSizeOffset.newBuilder().setOffest(0).setPageSize(pageSize).build());
+    }
+    PageSizeOffset pageSizeOffset = decode(pageToken);
+    if (pageSizeOffset.getPageSize() != pageSize) {
+      throw new BadRequestException(
+          String.format(
+              "Token page size '%d' does not match input page size '%d'.",
+              pageSizeOffset.getPageSize(), pageSize));
```
calbach, line 36: Technically you could allow this in the API, though it would depend on your implementation (this current implementation cannot support page size changes mid-stream). See other comment about `next_id`

api/src/main/proto/bio/terra/tanagra/utils/pagination.proto
```diff
@@ -0,0 +1,17 @@
...
+syntax = "proto3";
+
+package bio.terra.tanagra.utils;
+option java_package = "bio.terra.tanagra.proto.utils.pagination";
+option java_multiple_files = true;
+
+// A simple pagination message with the page size and offset for the page.
+message PageSizeOffset {
+  // The number of results to return per page.
+  int32 page_size = 1;
```
calbach, line 14: This is a good starting point and works well for rarely changing (effectively static) resources like underlay / entitity. For mutable endpoints (for Tanagra, mostly this pertains to artifacts), likely you'll want something more stable, e.g. `next_id`, or for performance, possibly something more tailored the endpoint that allows you to take advantage of the index. Alternatively, having some other way of dealing with insert/deletes of resources mid-pagination stream

api/src/main/java/bio/terra/tanagra/common/Paginator.java
wchamber: Thanks, added a field to the proto and updated the Paginator.

calbach: Nice!

calbach (approved):

PR #25, opened 2021-08-17
Add an api for retrieving how an entity filter can be constructed.

wchamber:
This endpoint is meant to be used by the frontend to shape the user experience of creating an entity filter.
Needs additional underlay proto support to specify possible values.
The entity search hint is a little speculative because I'm still not quite sure what we want entity search to look like, but it's meant to be analogous to the AoU CB domain search pages.


api/src/main/resources/api/service_openapi.yaml
```diff
@@ -75,6 +75,21 @@ paths:
...
+      - $ref: '#/components/parameters/Underlay'
+      - $ref: '#/components/parameters/Entity'
+    get:
+      # TODO authorization
+      security: []
+      summary: |
+        Returns what attributes and relationships can be used to create an entity filter and hints
+        for their values.
+      operationId: getFilters
+      tags: [EntitiesFilters]
```
calbach, line 89: I just now realized that `Entity` just has a `schema: string` field on it. Should this information here instead just be embedded onto the entity model itself? What layers of abstraction would this break?

wchamber: Sorry, I'm not sure what you mean. The `#/components/parameters/Entity` is the `{entityName}` parameter with the schema: string field. There's also a `#/components/schemas/Entity` that has `name` and `attributes` properties.
I don't know what information you're asking about.

calbach: As in , this filter metadata. An alternative to this hierarchical API would be to simply return this filter information on the `Entity` model itself. Do you imagine cases where you'd want to grab the entity and its filter details separately?

wchamber: Ah, hmmm. I'm not sure if we'd ever want to grab them separately, or not care about one or the other.
Putting them on the model itself would start incorporating search concerns with the logical model. That seems fine-ish for how to filter entities/attributes, but I don't know if we'd keep doing that for other Tanagra searches. Like if we wanted to let the users choose what to group by on demographic count searches, would we also add a 'is_groupable' field to 'Attribute'?
I'm not sure to how to weigh some modeling purity vs a client use case. Is it annoying to do multiple backend calls? Are we going to want to just have a "tell me everything about this entity and its attributes" endpoint eventually anyway.
Don't know, happy to start in one direction then change later with client input as well.

calbach: I think we have a lot of flexibility with what we decide in this case because these are static / read-only metadata endpoints. If we were to have full CRUD for these models, things would get a bit more complicated and I'd have to think harder (I might still suggest embedding on the entity resource, though). The client will almost certainly be fetching this all at once. 

My gut says that just encoding this onto the Entity model is probably the best path. I could be wrong about this though, and I'm fine with the way you have it, if you feel strongly on this point.

calbach (approved):

api/src/main/resources/api/service_openapi.yaml
wchamber: SG, PTAL

api/src/main/resources/api/service_openapi.yaml
```diff
@@ -439,6 +515,16 @@ components:
...
 
+    Relationship:
+      type: object
+      description: A relationship to another entity.
+      properties:
+        name:
+          type: string
+          description: The name of the relationship.
+        relatedEntity:
+          $ref: '#/components/schemas/Entity'
```
calbach, line 526: `entity.relationships[].relatedEntity.relationships[].relatedEntity...`

I would probably just use the entity name here to avoid these kinds of issues and potential cycles

calbach (approved):

api/src/main/resources/api/service_openapi.yaml
wchamber: SG, does seem simpler. Will also change the proto. But it does mean you may need to call this end point multiple times during construction.

PR #26, opened 2021-08-17
Add underlay proto definition for what's supported in an entity filter.

wchamber:
This will power the frontend to shape the user experience of creating an entity filter.
Move ColumnId and AttributeId next to their respective identified messages.

### DIFF
--- a/api/src/main/proto/bio/terra/tanagra/underlay/search.proto
+++ b/api/src/main/proto/bio/terra/tanagra/underlay/search.proto
@@ -25,13 +25,11 @@ message FilterableAttribute {
   // The attribute that can be used in the entity filter.
   AttributeId attribute = 1;
 
-  // Hints are usually but not necessarily exclusive.
-
-  optional EntitySearchHint entity_search_hint = 2;
-
-  optional EnumHint enum_hint = 3;
-
-  optional IntegerBoundsHint integer_bounds_hint = 4;
+  oneof hint {
+    EntitySearchHint entity_search_hint = 2;
+    EnumHint enum_hint = 3;
+    IntegerBoundsHint integer_bounds_hint = 4;
+  }
 }
 
 // A hint for a filterable attribute where the values are the primary id of an
@@ -74,8 +72,6 @@ message IntegerBoundsHint {
 message FilterableRelationship {
   // The name of the relationship.
   string relationship_name = 1;
-  // The name of the related entity in the relationship.
-  string related_entity = 2;
   // The schema of the filter available on the related entity.
-  EntityFiltersSchema entity_filters_schema = 3;
+  EntityFiltersSchema entity_filters_schema = 2;
 }
--- a/api/src/main/proto/bio/terra/tanagra/underlay/search.proto
+++ b/api/src/main/proto/bio/terra/tanagra/underlay/search.proto
@@ -1,0 +1,81 @@
+// Messages describing how to construct Tanagra entity searches.
+
+syntax = "proto3";
+
+package bio.terra.tanagra.underlay;
+option java_package = "bio.terra.tanagra.proto.underlay";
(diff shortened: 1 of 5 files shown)


api/src/main/proto/bio/terra/tanagra/underlay/search.proto
```diff
@@ -0,0 +1,81 @@
...
+
+package bio.terra.tanagra.underlay;
+option java_package = "bio.terra.tanagra.proto.underlay";
+option java_multiple_files = true;
+
+import "bio/terra/tanagra/underlay/entities.proto";
+import "bio/terra/tanagra/underlay/mapping.proto";
+
+// The schema of how an entity filter can be constructed for a given entity.
+message EntityFiltersSchema {
```
calbach, line 13: Same question as on the other PR: are these properties of the filters, or properties of the entity attributes themselves? Is the schema sufficient to express what search operations are possible, or do we need to constraint / expand upon that with this additional structure?

wchamber: I think these are properties of the filters. For example, we might not want to suggest users filter on every attribute on the entity - e.g. we don't need selectors for both `person.gender` and `person.gender_concept_id`, we would just have an EnumHint for `person.gender_concept_id`.
We might need to further expand. Right now the supported operations on the DataType are implicit (not that we've defined many operations yet, just `EQUALS` and `LESS_THAN`). We might want some special string term search operator that's not available on every `STRING` entity, at which point we might want to enumerate the supported attributes, or add sub types for DataTypes. Not sure.

api/src/main/proto/bio/terra/tanagra/underlay/search.proto
```diff
@@ -0,0 +1,81 @@
...
+// A hint for a filterable attribute that the values are within certain bounds.
+message IntegerBoundsHint {
+  optional int64 min = 1;
+  optional int64 max = 2;
+}
+
+// A relationship on an entity that may be used in an entity filter.
+message FilterableRelationship {
+  // The name of the relationship.
+  string relationship_name = 1;
```
calbach, line 76: Does this need to exactly match the relationship definition in the entity model? If so, should this just act as a lookup into that relationship config? thereby making the `related_entity` here duplicative/unnecessary?

api/src/main/proto/bio/terra/tanagra/underlay/search.proto
```diff
@@ -0,0 +1,81 @@
...
+  repeated FilterableRelationship relationships = 3;
+}
+
+// An attribute on an entity that may be used for filtering with hints on how to find
+// interesting values.
+message FilterableAttribute {
+  // The attribute that can be used in the entity filter.
+  AttributeId attribute = 1;
+
+  // Hints are usually but not necessarily exclusive.
```
calbach, line 28: I'm curious - what is the exception you are imagining here? An attribute which an overloaded purpose?

api/src/main/proto/bio/terra/tanagra/underlay/search.proto
calbach: Thanks, I think I'm convinced.

calbach (approved):

api/src/main/proto/bio/terra/tanagra/underlay/search.proto
wchamber: ... I guess I don't have an example. I can make it a oneof for now.

api/src/main/proto/bio/terra/tanagra/underlay/search.proto
wchamber: Yep, good catch.

calbach (approved):

PR #27, opened 2021-08-30
Add backticks around project & database ids for BQ

wchamber:
While backticks aren't necessary in BQ SQL APIs, they're still needed on some BQ consoles. Adding backticks around always for now.
When we support non BQ tables, we can adjust how these kinds of specializations happen.

### DIFF
--- a/api/src/main/java/bio/terra/tanagra/service/search/SqlVisitor.java
+++ b/api/src/main/java/bio/terra/tanagra/service/search/SqlVisitor.java
@@ -198,9 +198,9 @@ public class SqlVisitor {
     }
 
     private String resolveTable(Table table) {
-      // projectId.datasetId.table
+      // `projectId.datasetId.table`
       return String.format(
-          "%s.%s.%s", table.dataset().projectId(), table.dataset().datasetId(), table.name());
+          "`%s.%s.%s`", table.dataset().projectId(), table.dataset().datasetId(), table.name());
     }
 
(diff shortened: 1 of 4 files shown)


api/src/main/java/bio/terra/tanagra/service/search/SqlVisitor.java
```diff
@@ -198,9 +198,9 @@ public String resolveTable(EntityVariable entityVariable) {
     }
 
     private String resolveTable(Table table) {
-      // projectId.datasetId.table
+      // `projectId.datasetId`.table
       return String.format(
-          "%s.%s.%s", table.dataset().projectId(), table.dataset().datasetId(), table.name());
+          "`%s.%s`.%s", table.dataset().projectId(), table.dataset().datasetId(), table.name());
```
calbach, line 203: Hmm.. perhaps I'm wrong - but isn't it typical to include the `table` in the backticks as well?

calbach (approved):

api/src/main/java/bio/terra/tanagra/service/search/SqlVisitor.java
wchamber: Maybe, but if you do, you have to include the full table qualification when the table isn't aliased i.e. ```
`my-project.dataset.foo` where foo.x =0
``` doesn't work.

PR #28, opened 2021-08-30
Implement piping entity filter schema to get/list entity API

wchamber:
- Unnest the entity filter schema in the proto definition. Each entity now has a single filter schema.
- fix controller test package location
- Split UnderlayConversion into more helper methods


api/src/main/java/bio/terra/tanagra/service/search/Relationship.java
```diff
@@ -14,6 +14,37 @@
   /** The second entity in the relationship. */
   public abstract Entity entity2();
 
+  /**
+   * Returns whether the {@code x} and {@code y} entities are the two entities of the relationship,
+   * in any order.
+   */
+  public boolean unorderedEntitiesAre(Entity x, Entity y) {
```
calbach, line 21: nit: are what? Perhaps it's meant to be implied by the `Relationship` noun, but I don't quite follow. Maybe something like `relatesEntitiesUnordered(x, y)` ? Or just `hasEntitiesUnordered(x, y)` for consistency with the other method?

api/src/main/java/bio/terra/tanagra/service/underlay/Underlay.java
```diff
@@ -43,17 +45,17 @@
    */
   public abstract ImmutableMap<Relationship, ForeignKey> foreignKeys();
 
+  /** Map from entities to the filters schema that supports the entity. */
```
calbach, line 48: nit: I would add ", if any", assuming that it is legal for an entity to omit a filter schema

api/src/main/java/bio/terra/tanagra/service/underlay/Underlay.java
```diff
@@ -67,6 +69,15 @@
     return Optional.of(matching.get(0));
   }
 
+  /** Returns all the relationships that the entity is a member of. */
+  public Set<Relationship> getRelationshipsOf(Entity entity) {
+    return relationships().values().stream()
+        .filter(
+            relationship ->
+                relationship.entity1().equals(entity) || relationship.entity2().equals(entity))
```
calbach, line 77: nit: seems this can be replaced by `relationship.hasEntity(entity)`

api/src/main/resources/api/service_openapi.yaml
```diff
@@ -525,6 +525,9 @@ components:
         relatedEntity:
           type: string
           description: The name of the related entity.
+        filterable:
+          type: boolean
+          description: Whether this relationship should be used when creating relationships.
```
calbach, line 530: nit: is this comment correct? seems circular

calbach (approved):

api/src/main/java/bio/terra/tanagra/service/search/Relationship.java
wchamber: thanks, went with `hasEntitiesUnordered`

api/src/main/java/bio/terra/tanagra/service/underlay/Underlay.java
wchamber: good point, added.

api/src/main/java/bio/terra/tanagra/service/underlay/Underlay.java
wchamber: done

api/src/main/resources/api/service_openapi.yaml
wchamber: thanks, fixed up

PR #29, opened 2021-08-31
Implement entity instance search SQL generation

wchamber:
Implements the entity instance search endpoint for generating SQL.
Add a builder for Selection.SelectExpression.
Fix parametr names in open api paths. Noticed through manual usage of the swagger UI.

### DIFF
--- a/api/src/main/java/bio/terra/tanagra/service/query/EntityDataset.java
+++ b/api/src/main/java/bio/terra/tanagra/service/query/EntityDataset.java
@@ -1,0 +1,38 @@
+package bio.terra.tanagra.service.query;
+
+import bio.terra.tanagra.service.search.Attribute;
+import bio.terra.tanagra.service.search.EntityVariable;
+import bio.terra.tanagra.service.search.Filter;
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+
+/** A query for a dataset of entity instances. */
+@AutoValue
+public abstract class EntityDataset {
+  /** The primary entity and variable that the dataset is being created from. */
+  public abstract EntityVariable primaryEntity();
+
+  /** The attributes selected to make the columns of the dataset. */
+  public abstract ImmutableList<Attribute> selectedAttributes();
+
+  /** The filter to apply to the primary entity. */
+  public abstract Filter filter();
+
+  public static Builder builder() {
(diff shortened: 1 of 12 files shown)


api/src/main/java/bio/terra/tanagra/service/query/EntityDataset.java
```diff
@@ -0,0 +1,58 @@
...
+
+import bio.terra.tanagra.service.search.Attribute;
+import bio.terra.tanagra.service.search.EntityVariable;
+import bio.terra.tanagra.service.search.Filter;
+import com.google.auto.value.AutoValue;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+
+/** A query for a dataset of entity instances. */
```
calbach, line 11: Terminology: what does dataset refer to here? Is this using it in the "BigQuery dataset" sense? Or is this like the "dataset" in "PDB"?

api/src/test/java/bio/terra/tanagra/service/query/QueryServiceTest.java
```diff
@@ -1,7 +1,6 @@
 package bio.terra.tanagra.service.query;
 
-import static bio.terra.tanagra.service.underlay.NauticalUnderlayUtils.SAILOR;
-import static bio.terra.tanagra.service.underlay.NauticalUnderlayUtils.SAILOR_RATING;
+import static bio.terra.tanagra.service.underlay.NauticalUnderlayUtils.*;
```
calbach, line 3: nit: static import `*` seems potentially problematic, general advice I've followed is to avoid all `*` imports in java

calbach (approved):
Sorry for the delay. LGTM

api/src/main/java/bio/terra/tanagra/service/query/EntityDataset.java
wchamber: Dataset, like what gets exported for researchers. Very overloaded, would love a better name.

api/src/test/java/bio/terra/tanagra/service/query/QueryServiceTest.java
wchamber: SGTM. Some other test files doing that, I'll update incrementally or all at once later.

PR #32, opened 2021-09-07
Implement initial BigQuery job execution

wchamber:
This adds some initial support for Tanagra to execute queries against BigQuery. With this, I can (with my user credentials) retrieve results through a Tanagra API.

Add database access classes and interfaces for querying databases.
Implement BigQuery database access.
Implements entity instances search API to execute queries.

More todo in later PRs:
Add BigQuery integration tests once we have a test environment with BigQuery.
Implement JDBC database access.
Add support for paginating database query results.
Add QueryService and controller tests for entity instance searches.


api/src/main/java/bio/terra/tanagra/service/databaseaccess/CellValue.java
```diff
@@ -0,0 +1,34 @@
...
+ * <p>This interface allows us to read data from different databases in a simple but uniform way.
+ * Different database types should implement this for returning values.
+ */
+public interface CellValue {
+
+  /** The type of data in this cell. */
+  DataType dataType();
+
+  /** Returns whether the cell's value is null. */
+  boolean isNull();
```
calbach, line 17: Any consideration towards using optional in the various interfaces rather having the `isNull` field? I'm not sure I have a strong opinion.

api/src/main/java/bio/terra/tanagra/service/databaseaccess/bigquery/BigQueryExecutor.java
```diff
@@ -0,0 +1,69 @@
...
+    QueryJobConfiguration queryJobConfiguration =
+        QueryJobConfiguration.newBuilder(queryRequest.sql()).build();
+    // TODO add pagination. Right now we always only return the first page of results.
+    TableResult tableResult = getResults(queryJobConfiguration);
+
+    Iterable<RowResult> rowResults =
+        Iterables.transform(
+            tableResult.getValues(),
+            (FieldValueList fieldValueList) ->
+                new BigQueryRowResult(fieldValueList, queryRequest.columnHeaderSchema()));
```
calbach, line 42: Seems like the header schema is mostly passed through from the reques t- at what point would we expect to see an error from a mismatch between the expected header schema and the actual row results?

api/src/main/java/bio/terra/tanagra/service/databaseaccess/bigquery/BigQueryCellValue.java
```diff
@@ -0,0 +1,46 @@
+package bio.terra.tanagra.service.databaseaccess.bigquery;
+
+import bio.terra.tanagra.service.databaseaccess.CellValue;
+import bio.terra.tanagra.service.databaseaccess.ColumnSchema;
+import bio.terra.tanagra.service.search.DataType;
+import com.google.cloud.bigquery.FieldValue;
+
+/** A {@link CellValue} for BigQuery's {@link FieldValue}. */
+@SuppressWarnings("PMD.PreserveStackTrace")
```
calbach, line 9: nit: can it be at the method? Is class the smallest granularity for this one?

calbach (approved):

api/src/main/java/bio/terra/tanagra/service/databaseaccess/CellValue.java
wchamber: I admit I didn't think much about that and was just looking at BQ's FieldValue. There's also returning `@Nullable Long/String/etc`.
I think I like optionals most slightly, going with that, thanks.

api/src/main/java/bio/terra/tanagra/service/databaseaccess/bigquery/BigQueryCellValue.java
wchamber: Yep, thanks!

api/src/main/java/bio/terra/tanagra/service/databaseaccess/bigquery/BigQueryExecutor.java
wchamber: - yes, indirectly from the attributes and column mapping. (and also the result columns could be from the selection, e.g. count(x) -> int)
- We get the error when the value is retrieved. We could do it when we do the wrapping of FieldValue -> CellValue, which is a little earlier but not much better. I don't know we can figure it out earlier than that.

The assert I was referring to is only for the number of columns & number of FieldValues, in the BigQueryRowResult constructor (sorry, misleading). https://github.com/DataBiosphere/tanagra/blob/1bcb8c6d91623e8b652fb3a44494cbae547d00c6/api/src/main/java/bio/terra/tanagra/service/databaseaccess/bigquery/BigQueryRowResult.java#L14-L19

PR #35, opened 2021-09-20
Add beam utility for finding reachability of a graph

wchamber:
Use JUnit 4 in workflow tests as the Apache Beam TestPipeline makes use of it.
To be used to make hierarchical relationship indexes job in another PR.


workflow/src/main/java/bio/terra/tanagra/workflow/GraphUtils.java
```diff
@@ -0,0 +1,131 @@
...
+
+      PCollection<KV<T, T>> nMinus1AllPaths = allPaths.get(n - 1);
+
+      PCollection<KV<T, T>> newAllPaths =
+          PCollectionList.of(nMinus1AllPaths)
+              .and(nExactPaths)
+              .and(concatenate(nExactPaths, nMinus1AllPaths, "allPaths temp N" + n))
+              .apply("flatten N" + n, Flatten.pCollections());
+      int newLongestPathLength = 2 * n - 1;
+      if (newLongestPathLength >= maxPathLength) {
```
calbach, line 81: nit: maybe rename `maxPathLength` to make it clearer that this is to bound computation, and is not an exact cutoff to the answer e.g. `maxPathLengthHint`?

workflow/src/main/java/bio/terra/tanagra/workflow/GraphUtils.java
```diff
@@ -0,0 +1,131 @@
...
+    Map<Integer, PCollection<KV<T, T>>> allPaths = new HashMap<>();
+    // exactPaths[n] is the set of vertices where there exists a path from V0 -> V1 in exactly N
+    // edges.
+    Map<Integer, PCollection<KV<T, T>>> exactPaths = new HashMap<>();
+
+    // The input edges are both allPaths[1] and exactPaths[1].
+    allPaths.put(1, edges);
+    exactPaths.put(1, edges);
+
+    // Iteratively build up allPaths in O(log_2(maxPathLength)) joins. We want to avoid doing many
```
calbach, line 51: Nice algorithm. Is there a name for this or was this something you just fit to purpose for this? May be worth linking a wikipedia article or something for visual learners, if there is any such documentation

workflow/src/main/java/bio/terra/tanagra/workflow/GraphUtils.java
```diff
@@ -0,0 +1,131 @@
...
+
+      PCollection<KV<T, T>> newAllPaths =
+          PCollectionList.of(nMinus1AllPaths)
+              .and(nExactPaths)
+              .and(concatenate(nExactPaths, nMinus1AllPaths, "allPaths temp N" + n))
+              .apply("flatten N" + n, Flatten.pCollections());
+      int newLongestPathLength = 2 * n - 1;
+      if (newLongestPathLength >= maxPathLength) {
+        // Stop when we have computes at least maxPathLength length paths.
+        return newAllPaths;
```
calbach, line 83: opt nit: technically we could also exit when the output set stops changing, but this operation might be the same cost or similar to just running out the rest of the cycles, especially since it's a logarithmic number of steps.

workflow/src/main/java/bio/terra/tanagra/workflow/GraphUtils.java
```diff
@@ -0,0 +1,131 @@
...
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.PCollectionList;
+import org.apache.beam.sdk.values.TupleTag;
+
+/** Utilities for working with graphs in Apache Beam. */
+public final class GraphUtils {
+  private GraphUtils() {}
+
+  /**
+   * Computes the transitive closure of a directed acyclic graph defined by {@code edges}.
```
calbach, line 25: Out of curiosity, what happens if this is not acyclic? Is that actually a prerequisite of this function?

workflow/src/test/java/bio/terra/tanagra/workflow/GraphUtilsTest.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  /** Convert a multimap to an equivalent list of KVs. */
+  private static <T> List<KV<T, T>> convertToKvs(Multimap<T, T> multimap) {
+    return multimap.entries().stream()
+        .map(entry -> KV.of(entry.getKey(), entry.getValue()))
+        .collect(Collectors.toList());
+  }
+
+  @Test
```
calbach, line 40: opt: may be worth including some smaller edge cases for the logarithmic approach, e.g. single-edge graph, with a maxPathLen of 1

calbach (approved):
Nice!

workflow/src/main/java/bio/terra/tanagra/workflow/GraphUtils.java
wchamber: Nope, though it will generate a lot of duplicates. Might want to add calls to `Distinct` to deduplicate if that happens often.

workflow/src/main/java/bio/terra/tanagra/workflow/GraphUtils.java
wchamber: Beam doesn't support that "conditional iteration" yet, so unfortunately we can't exit when the output set doesn't change. Added a comment and link

workflow/src/test/java/bio/terra/tanagra/workflow/GraphUtilsTest.java
wchamber: Good idea, added.

workflow/src/main/java/bio/terra/tanagra/workflow/GraphUtils.java
wchamber: Found internally in a google doc initially, but added a link to a paper reference. No great graphics unfortunately.

workflow/src/main/java/bio/terra/tanagra/workflow/GraphUtils.java
wchamber: thanks, done.

PR #36, opened 2021-09-20
Add initial index job for flattening hierachical relationships

wchamber:
Adds an Apache Beam job for creating a BQ table as an index for a flattened hierachical relationship of ancestor -> all descendants.
Provides some notes on an end-end working example in the `broad-tanagra-dev` project.
This does not begin to address making this easy to make further deployments of. I'd like to have a few more examples before creating some common ways to make these easeir to re-use.
Example successful dataflow run: https://console.cloud.google.com/dataflow/jobs/us-central1/2021-09-20_13_48_33-8166662199084750131?project=broad-tanagra-dev


workflow/src/main/java/bio/terra/tanagra/workflow/FlattenHierarchy.java
```diff
@@ -0,0 +1,154 @@
...
+  private static TableSchema flattenedHierarchySchema(FlattenHierarchyOptions options) {
+    return new TableSchema()
+        .setFields(
+            List.of(
+                new TableFieldSchema()
+                    .setName(options.getOutputAncestorColumn())
+                    .setType("INTEGER"),
+                new TableFieldSchema()
+                    .setName(options.getOutputDescendantColumn())
+                    .setMode("REPEATED")
```
calbach, line 93: TO be clear about what this is doing, we're going to expand the entire subtree for each concept, and put it into this singular column - correct? Or is this only the leaves?

What are the limits on this in BigQuery (or if not there, what is a reasonable limit in general?). I believe that many of these trees are quite large (20K concepts, maybe more)

workflow/src/main/java/bio/terra/tanagra/workflow/FlattenHierarchy.java
```diff
@@ -0,0 +1,154 @@
...
+                                context
+                                    .getPipelineOptions()
+                                    .as(FlattenHierarchyOptions.class)
+                                    .getOutputDescendantColumn(),
+                                descendants);
+                    context.output(row);
+                  }
+                }))
+        .apply(
+            BigQueryIO.writeTableRows()
```
calbach, line 146: Neat

workflow/src/main/java/bio/terra/tanagra/workflow/FlattenHierarchy.java
```diff
@@ -0,0 +1,154 @@
+package bio.terra.tanagra.workflow;
+
+import avro.shaded.com.google.common.collect.Lists;
```
calbach, line 3: This dep looks suspect, is this something for Beam compatibility, or just an accidental import?

workflow/src/main/java/bio/terra/tanagra/workflow/FlattenHierarchy.java
wchamber: Yep, that's correct. `ancestor, array[descendants]`
The BigQuery limit is 100MB per row. The synpuf dataset has 100k children for the largest  condition snomed concept, 'Clinical Finding'.
Maybe not such a great idea to group it this way. The idea with this table is to support the 'DESCENDANT_OF' operator. With this structuring, after the scan on the ancestor column, we only need to look at a single descendant column.
 Some alternatives, would love to hear other suggestions:
single `ancestor, descendant` per row, so many duplicates of a given ancestor.
descendant to array[ancestor]. at lookup time, we need to scan the entire descendants table to look for ancestors. 
thoughts?

calbach: Also @freemabd , what is the largest concept tree the cohort builder currently supports? IIRC you actually disable some of the larger nodes, and do not let the user select them (I think "clinical finding" probably falls in this category)

wchamber: `ancestor, descendant` SGTM.
Clustering will have to wait a bit since it's disabled without a time partition due to an out of date biquery restriction :/ But there's a PR just out, so there's hope.
https://issues.apache.org/jira/browse/BEAM-10652

freemabd: > Also @freemabd , what is the largest concept tree the cohort builder currently supports? IIRC you actually disable some of the larger nodes, and do not let the user select them (I think "clinical finding" probably falls in this category)

Definitely, SNOMED Conditions tree. Clinical finding is the very top level of that tree, but everything under that node is selectable. So the largest nodes are probably "Disease" and "Finding by site" with ~28K descendants

PR #37, opened 2021-09-21
Implement a DESCENDANT_OF_INCLUSIVE filter operator

wchamber:
- Add proto configuration definition for hierarchical attributes and descendant tables. This is the first example of underlay configuration for an index table.
- Implement SqlVisitor to use the referenced descendants index table for DESCENDANT_OF_INCLUSIVE
- Add the operator to the API and implement conversion

Work for further PRs:
Parent/child relationship table proto definitions for hierarchical browsing.
DESCENDANT_OF non inclusive operator, or consider "smart" inclusion.
exposing what attributes support the descendant_of operator through the metadata API to the FE.


api/src/main/java/bio/terra/tanagra/service/query/api/FilterConverter.java
```diff
@@ -75,11 +79,25 @@ public Filter convert(ApiFilter apiFilter, VariableScope scope) {
...
+        return Filter.BinaryFunction.Operator.DESCENDANT_OF;
       default:
         throw new BadRequestException("Unknown BinaryFilterOperator: " + apiOperator.toString());
     }
   }
 
+  /** Checks if the operator maybe be used with the attribute or else throws. */
+  private void checkAttributeOperatorMatch(
+      Attribute attribute, Filter.BinaryFunction.Operator operator) {
+    if (operator.equals(Filter.BinaryFunction.Operator.DESCENDANT_OF)
```
calbach, line 92: opt nit: As my own rule of thumb I generally reverse the order of this kind of comparison to offer more null-safety, i.e. CONSTANT.equals(parameter)

api/src/main/java/bio/terra/tanagra/service/query/api/FilterConverter.java
```diff
@@ -75,11 +79,25 @@ public Filter convert(ApiFilter apiFilter, VariableScope scope) {
...
+        && underlay.hierarchies().get(attribute) == null) {
```
calbach, line 93: nit: containsKey? Or is having null in the map significant?

api/src/main/java/bio/terra/tanagra/service/search/SqlVisitor.java
```diff
@@ -130,12 +133,54 @@ public String visitBinaryComparision(BinaryFunction binaryFunction) {
...
+      }
+      Expression.AttributeExpression attributeExpression =
+          (Expression.AttributeExpression) binaryFunction.left();
+      Hierarchy hierarchy =
+          searchContext
+              .underlay()
+              .hierarchies()
+              .get(attributeExpression.attributeVariable().attribute());
+      if (hierarchy == null) {
+        throw new BadRequestException(
```
calbach, line 161: To check my understanding: using DECENDANT_OF with a leaf concept is illegal - right?

api/src/test/resources/underlays/nautical.pbtext
```diff
@@ -138,6 +142,17 @@ datasets: {
       data_type: STRING
     }
   }
+  tables: {
```
calbach, line 145: So indexes will generally be tables, but not entities themselves - correct?

api/src/test/resources/underlays/nautical.pbtext
```diff
@@ -362,6 +390,24 @@ entity_filters_schemas: {
     relationship_name: "sailor_reservation"
   }
 }
+hierarchies: {
+  attribute: {
```
calbach, line 394: How will this work for OMOP? Would there just be one such hierarchy defined on the concept table? Or would there be different hierarchies depending on the specific domain table?

api/src/test/resources/underlays/nautical.pbtext
wchamber: Yep, indexes are tables to support operations on entities, but they're not exposed as separate entities.
e.g. we also want a free text search index. that will be backed initially by a postgres table and be related to an entity, but it's not a distinct entity in itself.

api/src/test/resources/underlays/nautical.pbtext
wchamber: So far, it's one hierarchy per entity.attribute. in this case `condition_occurrence.condition_concept_id`, where `procedures.procedure_concept_id` would have its own set of hierarchies. This let's hierarchical browsing be distinct to the domain.
Note that there are not multiple hierarchies per attributes - e.g. we would roll snomed and other condition vocabulary hierarchies together, they would just have different roots.

api/src/main/java/bio/terra/tanagra/service/search/SqlVisitor.java
wchamber: Ergh, that should probably be fine with this operator does - include the concept and its descendants so it's legal on leaves.
We could do this in the index generation (i.e. everything has the identity relationship), or in our SQL expansion 1(attribute = 1234 OR attribute in (select descendants of 1234))`.
In OMOP's concept_ancestor, each concept is recorded as the ancestor of itself, so that's some kind of precedent. https://www.ohdsi.org/web/wiki/doku.php?id=documentation:cdm:concept_ancestor
Maybe also a better name for the operator? `INCLUDE_DESCENDANTS`? `IS_OR_DESCENDANT_OF`? Thoughts?

calbach: I recall that the AoU cohort builder actually supports both (select a tree WITH the parent node, or without), but I can't remember if:

1. this is built into the index (i.e. criteria tables), or
2. this is something the user can decide at query time

 @freemabd could you confirm which one it is?

If it's scenario (1), your current operator name makes sense, though arguably this is too opaque. If it's (2), then perhaps we should have two operators OR we just go with what you have now (exclude parent), and leave it to the client to compose the predicate themselves.

freemabd: > I recall that the AoU cohort builder actually supports both (select a tree WITH the parent node, or without), but I can't remember if:
> 
> 1. this is built into the index (i.e. criteria tables), or
> 2. this is something the user can decide at query time

This is built into the criteria table.

wchamber: Per offline discussion with CH - for now we'll make the inclusion of the parent explicit in the name of the operator. Later we can evaluate if we want to make inclusion determined by the index, or support different operators.
@calbach PTAL

calbach (approved):

PR #38, opened 2021-09-22
Add an underlay for a dev synpuf omop dataset.

wchamber:
The mapping is copied form the existing omop_test.yaml underlay, but pointed at a different backing datset.
More sophisticated omop templating and copying will come later.

### DIFF
--- a/api/src/main/resources/application.yml
+++ b/api/src/main/resources/application.yml
@@ -16,4 +16,4 @@ spring:
 
 tanagra:
   underlay:
-    underlay-yaml-files: "underlays/omop_test.yaml"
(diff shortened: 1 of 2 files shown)


calbach (approved):
For future reference, how did you generate this? Manually? Or via BigQuery schema? Any CLIs you can include in the PR description might be useful later.

wchamber:
> For future reference, how did you generate this? Manually? Or via BigQuery schema? Any CLIs you can include in the PR description might be useful later.

I copied the existing [omop_test.yaml] (https://github.com/DataBiosphere/tanagra/blob/main/api/src/main/resources/underlays/omop_test.yaml) file and changed the database BQ project & dataset id. That file I wrote manually.

PR #40, opened 2021-09-23
Add a SearchTermFilter to the API for text search on entities.

wchamber:
This is meant to allow the text search for things like concepts.
This is a somewhat unusual filter in that we maybe don't want to encourage users to include it in their cohort building, but it's mostly used to select entities during cohort construction.

We could do another API approach, like having a dedicated `instances:searchTerm` API and not include SearchTermFilter as a part of a general Filter. That new API may look a lot like `instances:search` as we want to return values for entity instances. I'm ambivalent. LMK if a 1 pager would be helpful.

TODOs for support for this:
Underlay configuration support
Text search indexes for entities
Reconsider how EntitySearchHint should relate to this.

### DIFF
--- a/api/src/main/resources/api/service_openapi.yaml
+++ b/api/src/main/resources/api/service_openapi.yaml
@@ -351,6 +351,9 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/Relationship'
+        searchTermSupported:
+          description: Whether a SearchTermFilter can be used to filter on this entity.
+          type: boolean
 
     EntityDataset:
       description: A selection of entity attributes and filters on them.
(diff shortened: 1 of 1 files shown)


api/src/main/resources/api/service_openapi.yaml
```diff
@@ -466,6 +471,19 @@ components:
           type: string
         filter:
           $ref: '#/components/schemas/Filter'
+
+    SearchTermFilter:
```
calbach, line 475: The one risk of including this as part of the main filtering API is if the semantics of this are fuzzy, i.e. whether something constitutes a match could vary across versions of Tanagra, or versions of the indexing pipeline.

If it is indeed stable/portable, then this could actually be an interesting filter to have within a cohort or dataset selection. I'm trying to think whether or not it would be problematic. Probably OK being here, at least for now I think..

calbach (approved):

PR #42, opened 2021-09-29
Add initial pipeline for copying from BQ to CloudSQL

wchamber:
Adds an initial Apache beam pipeline for copying data from BigQuery datas to a Postgres CloudSQL instance.

Fix omop underlays where `vocabulary.vocabulary_id` should be a string. Discovered doing omop copies.
Remove `sex_at_birth` from synpuf underlay as it doesn't have that column in the base data.

TODOs for future:
Make this configuration more underlay driven.
Make an end2end testing environment and test pipelines there.
Consider using CloudSQL import from CSV for large tables.


api/src/main/resources/underlays/synpuf.yaml
```diff
@@ -108,13 +102,13 @@ datasets:
           - name: standard_concept
             dataType: STRING
           - name: vocabulary_id
-            dataType: INT64
+            dataType: STRING
```
calbach, line 105: FYI: Some of these discrepancies may be with the OMOP version, or might be adaptations made to the AoU dataset. I know sex_at_birth is definitely an AoU extension (feedback has been given to provide clearer labeling of these fields/tables)

workflow/src/main/java/bio/terra/tanagra/workflow/CloudSqlOptions.java
```diff
@@ -0,0 +1,56 @@
...
+
+  void setCloudSqlDatabaseName(String databaseName);
+
+  @Description("The username for connecting to the CloudSQL database.")
+  String getCloudSqlUserName();
+
+  void setCloudSqlUserName(String userName);
+
+  /**
+   * Create a {@link DataSourceConfiguration} for connectint to the Cloud SQL instance specified by
```
calbach, line 32: nit: connecting

workflow/src/main/java/bio/terra/tanagra/workflow/CopyBigQueryDataset.java
```diff
@@ -0,0 +1,152 @@
...
+import org.apache.beam.sdk.io.gcp.bigquery.BigQueryOptions;
+import org.apache.beam.sdk.io.jdbc.JdbcIO;
+import org.apache.beam.sdk.options.PipelineOptionsFactory;
+import org.apache.beam.sdk.transforms.SerializableFunction;
+import org.apache.commons.text.StringSubstitutor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** A batch Apache Beam pipeline for copying data from BigQuery to CloudSQL Postgres instances. */
+public class CopyBigQueryDataset {
```
calbach, line 29: nit: should this pipeline name include the destination? e.g. `CopyBigQueryDatasetToPostgres`, or is the intent to eventually have the destination be parameterizable?

workflow/src/main/java/bio/terra/tanagra/workflow/CopyBigQueryDataset.java
```diff
@@ -0,0 +1,152 @@
...
+    try (Connection connection = dataSource.getConnection();
+        Statement statement = connection.createStatement()) {
+      statement.execute(createStatement);
+      if (!connection.getAutoCommit()) {
+        connection.commit();
+      }
+    }
+  }
+
+  /** Creates an SQL string for creating a SQL table based on ta {@link Table}. */
```
calbach, line 80: nit: rm " ta "

workflow/src/main/java/bio/terra/tanagra/workflow/CopyBigQueryTable.java
```diff
@@ -0,0 +1,144 @@
...
+import org.apache.beam.sdk.values.PBegin;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.PDone;
+
+/**
+ * Copies the contents of a BigQuery table to a SQL table.
+ *
+ * <p>The SQL table must already exist.
+ */
+@AutoValue
```
calbach, line 32: Interesting that you're able to Autovalue in combination with PTransform; does the annotation basically just get applied to the generated class as well?

Relatedly, for my own benefit: do you know off-hand or have a guess about whether this can work with Hibernate entities?

calbach (approved):
Nice!

workflow/src/main/java/bio/terra/tanagra/workflow/CopyBigQueryDataset.java
wchamber: Being specific sounds good. We can rename if it does parameterize.

workflow/src/main/java/bio/terra/tanagra/workflow/CopyBigQueryTable.java
wchamber: I'm not sure which annotation you mean. For a PTransform, the `expand` is overriding the base PTransform function, unlike some of the DoFn or other function's in joins/JdbcIO calls that are based on annotations. The AutoValue generated class won't create it's own `expand`.
I mostly went with making it an autovalue for configurations and some casual reading of https://beam.apache.org/contribute/ptransform-style-guide/#fluent-builders-for-configuration , which is maybe a little more for library functions than what we're strictly doing here.

I'm not sure what you're asking with Hibernate entities.
If this is a question about Hibernate in Beam, there's some implicit serialization going on in all of these pipelines as data gets distributed to different VMs. Having not worked with Hibernate, I'd think it's POJO classes would do fine with serialization. I don't know how the Hibernate<->JDBC connection would as a part of pipeline transformations.

If the question was about inheriting annotations more generally (e.g. AutoValue classes for Hibernate `@Entity` clases), I think it behavior changes based on if the annotation itself is marked as[ java.lang.annotation.Inherited](https://docs.oracle.com/javase/7/docs/api/java/lang/annotation/Inherited.html). But AutoValue has some support for moving annotations along to the generated implementation classes https://github.com/google/auto/blob/master/value/userguide/howto.md#-copy-annotations-from-a-classmethod-to-the-implemented-classmethodfield. So, sounds feasible, but I haven't mucked with that before myself.

calbach: Sorry, rereading - my annotation question didn't make sense, my brain jumped ahead to how I could use this in combination with the `@Entity` annotation.

The latter is what I was asking about - thanks for the link.

PR #45, opened 2021-10-05
Initial commit of Tanagra UI

tjennison-google:
This commit includes
* Projet created using create-react-app
* Linting
* Framework for working with the Tanagra service
* Some barebones components to validate the project setup


ui/src/state.ts
```diff
@@ -0,0 +1,29 @@
+export class State {
```
calbach, line 1: What is the scope of this? Can it be named more specifically?

ui/src/state.ts
```diff
@@ -0,0 +1,29 @@
...
+  constructor() {
+    this.included = new Array<Group>();
+  }
+
+  included: Array<Group>;
+}
+
+export class Group {
```
calbach, line 9: It's possible you have a need for this to be a class for later, but an alternative that you can use with Typescript is instead (similar to a Go struct):

1. Just define these kinds of types as interfaces:

```
export interface Group {
  id: string
  criteria: Array<Criteria>
}
```

2. create instances via Javascript object literals

```
const g: Group = {
  id: 123,
  criteria: [...]
}
```

With this pattern I haven't had too much need for defining any custom model classes

ui/src/conceptList.tsx
```diff
@@ -0,0 +1,103 @@
...
+// Row acts as a bridge between the data format returned from the API and the
+// format expected by DataGrid.
+class Row {
+  id = 0;
+  [x: string]: string | number | boolean | undefined;
+}
+
+type ConceptListProps = {
+  underlay: string;
+  state: State;
```
calbach, line 17: Putting something called `State` into props seems slightly confusing to me since these are usually props and state are contrasting concepts in react.

ui/src/conceptList.tsx
```diff
@@ -0,0 +1,103 @@
...
+};
+
+export default function ConceptList(props: ConceptListProps) {
+  const [error, setError] = useState<Error | null>(null);
+  const [rows, setRows] = useState<Array<Row> | null>(null);
+
+  const params = useParams<{ group: string; criteria: string }>();
+  const api = useContext(EntityInstancesApiContext);
+
+  const columns: GridColDef[] = useMemo(() => {
```
calbach, line 28: opt nit: in ES6, you can omit the function brackets / return and just put an expression, even multiline as in this case

ui/src/pdb.test.tsx
```diff
@@ -0,0 +1,9 @@
+import { render, screen } from "@testing-library/react";
+import React from "react";
+import Pdb from "./pdb";
+
+test("renders learn react link", () => {
```
calbach, line 5: nit: update

ui/src/conceptList.tsx
```diff
@@ -0,0 +1,103 @@
...
+    return [
+      { field: "concept_name", headerName: "Concept Name", width: 400 },
+      { field: "concept_id", headerName: "Concept ID", width: 100 },
+      { field: "domain_id", headerName: "Domain", width: 100 },
+    ];
+  }, []);
+
+  useEffect(() => {
+    api
```
calbach, line 37: A pattern you will likely want to establish early is for handling of async cleanup on dismount (avoids this XHR completing after you've already navigated away, and trying to update some stale state or causing side-effects/errors). In Workbench, we use `AbortController` directly ([example](https://github.com/all-of-us/workbench/blob/master/ui/src/app/pages/analysis/runtime-panel.tsx#L1209-L1219)). In Hopper, they use `useAsync`, which I believe handles this as well.

ui/src/overview.tsx
```diff
@@ -0,0 +1,104 @@
...
+type OverviewProps = {
+  underlay: string;
+  state: State;
+};
+
+export default function Overview(props: OverviewProps) {
+  return (
+    <>
+      <Grid container columns={3}>
+        <Grid item xs={1} sx={{ mx: 2 }} key="included">
```
calbach, line 28: my understanding is that key is only really needed when you have multiple siblings of the same component type generated dynamically. I don't think this usage of key is necessary

ui/src/overview.tsx
```diff
@@ -0,0 +1,104 @@
...
+            {props.state.included.map((group) => (
+              <Box key={group.id}>
+                <ParticipantsGroup group={group} />
+                <Divider>
+                  <Chip label="AND" />
+                </Divider>
+              </Box>
+            ))}
+            <Box>
+              <AddCriteriaButton id="" key="" />
```
calbach, line 40: Same comment on key ^

ui/src/pdb.tsx
```diff
@@ -0,0 +1,72 @@
+import React, { useState } from "react";
+import { HashRouter, Route, Switch } from "react-router-dom";
+import { v4 as uuid } from "uuid";
+import ConceptList from "./conceptList";
+import Overview from "./overview";
+import "./pdb.css";
```
calbach, line 6: Is this using a library or a plugin? What is the intent for how stylesheets are being applied here?

ui/public/index.html
```diff
@@ -0,0 +1,55 @@
...
+    <meta
+      name="viewport"
+      content="minimum-scale=1, initial-scale=1, width=device-width"
+    />
+    <title>React PDB</title>
+  </head>
+  <body>
+    <noscript>You need to enable JavaScript to run this app.</noscript>
+    <div id="root"></div>
+    <!--
```
calbach, line 44: opt: consider removing some of the cruft in here

ui/src/conceptList.tsx
```diff
@@ -0,0 +1,103 @@
...
+                newEntity: props.filter,
+              },
+            },
+          },
+        },
+      })
+      .then(
+        (res) => {
+          const rows: Row[] = [];
+          if (!!res.instances) {
```
calbach, line 60: nit: technically !! is unnecessary here, as this expression would already be coerced to a boolean given the context

calbach (commented):
Thanks, looks good. Mostly style comments, but a few structural questions.

ui/README.md
```diff
@@ -0,0 +1,8 @@
+Runs the app in the development mode.\
```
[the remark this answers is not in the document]
tjennison-google: Yep, deleted one line too many.

ui/package.json
```diff
@@ -0,0 +1,68 @@
...
+    "@emotion/styled": "^11.3.0",
+    "@material-ui/core": "^4.12.3",
+    "@material-ui/icons": "^4.11.2",
+    "@mui/material": "^5.0.0",
+    "@mui/x-data-grid": "^5.0.0-beta.2",
+    "@openapitools/openapi-generator-cli": "^2.4.1",
+    "@testing-library/jest-dom": "^5.14.1",
+    "@testing-library/react": "^11.2.7",
+    "@testing-library/user-event": "^12.8.3",
+    "@types/jest": "^26.0.24",
```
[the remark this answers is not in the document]
tjennison-google: create-react-app does this intentionally: https://github.com/facebook/create-react-app/issues/6180#issuecomment-453640473

ui/package.json
```diff
@@ -0,0 +1,68 @@
...
+    "typescript": "^4.4.2",
+    "uuid": "^8.3.2",
+    "web-vitals": "^1.1.2"
+  },
+  "scripts": {
+    "start": "react-scripts start",
+    "build": "react-scripts build",
+    "test": "react-scripts test",
+    "eject": "react-scripts eject",
+    "tanagra": "rm -rf src/tanagra-api && openapi-generator-cli generate -i ../tanagra/api/src/main/resources/api/service_openapi.yaml -g typescript-fetch -t ./openapi-templates --additional-properties=typescriptThreePlus=true -o src/tanagra-api"
```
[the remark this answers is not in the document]
tjennison-google: I went with codegen.

ui/src/state.ts
tjennison-google: I already know I need to make major changes to what's here as part of the next steps so I'll keep approach in mind. Is the advantage just that the constructors are unnecessary or are there other reasons this is better?

ui/src/state.ts
tjennison-google: It's the state for the entire app. I've renamed it Dataset instead which is more descriptive.

ui/src/conceptList.tsx
```diff
@@ -0,0 +1,103 @@
...
+            res.instances.forEach((instance, i) => {
```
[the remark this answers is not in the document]
tjennison-google: Done

ui/src/conceptList.tsx
tjennison-google: Done

ui/src/pdb.test.tsx
tjennison-google: Done

ui/src/pdb.tsx
tjennison-google: This came from Create React App and I just haven't done anything with it yet.

ui/public/index.html
tjennison-google: I've cleaned up a bit but I'm just leaving most of it until I actually touch those features.

ui/src/criteria/concept.tsx
```diff
@@ -1,9 +1,33 @@
...
 import { DataGrid, GridColDef } from "@mui/x-data-grid";
 import React, { useContext, useEffect, useMemo, useState } from "react";
-import { useParams } from "react-router-dom";
-import { EntityInstancesApiContext } from "./apiContext";
-import { DataSet } from "./dataSet";
+import { EntityInstancesApiContext } from "../apiContext";
+import { Criteria, DataSet, Group } from "../dataSet";
+
+export class ConceptCriteria extends Criteria {
+  constructor(name: string, filter: string) {
```
calbach, line 8: typescript constructor shorthand can reduce some boilerplate here and avoid some issues, e.g. `name` is not declared

https://dev.to/satansdeer/typescript-constructor-shorthand-3ibd

ui/src/criteria/concept.tsx
```diff
@@ -1,9 +1,33 @@
...
+    super();
+    this.name = name;
+    this.filter = filter;
+  }
+
+  edit(dataSet: DataSet, group: Group): JSX.Element {
+    return (
```
calbach, line 15: Hmm, I thought this was a model class but it also contains rendering logic - should it? Minimally I think it would be good to name the method to be more obvious related to React / rendering, e.g. `renderEdit`

ui/src/dataSet.ts
```diff
@@ -1,29 +1,55 @@
...
+  }
+
+  findGroup(id: string): Group | undefined {
+    return this.groups.find((group) => group.id === id);
   }
 
-  included: Array<Group>;
+  listGroups(kind: GroupKind): Group[] {
+    const groups: Group[] = [];
+    this.groups.forEach((group) => {
```
calbach, line 15: nit: `this.groups.filter` probably better suited

calbach (commented):
Per chat - last commit was added accidentally; commenting for follow-up PR

ui/src/dataSet.ts
tjennison-google: Done

ui/src/criteria/concept.tsx
```diff
@@ -1,9 +1,33 @@
 import Typography from "@mui/material/Typography";
 import { DataGrid, GridColDef } from "@mui/x-data-grid";
 import React, { useContext, useEffect, useMemo, useState } from "react";
-import { useParams } from "react-router-dom";
-import { EntityInstancesApiContext } from "./apiContext";
-import { DataSet } from "./dataSet";
+import { EntityInstancesApiContext } from "../apiContext";
```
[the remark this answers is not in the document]
tjennison-google: Done

PR #46, opened 2021-10-07
Fix issues in commit accidentally added to previous PR #45

tjennison-work:


ui/src/dataset.ts
```diff
@@ -1,6 +1,6 @@
 import { v4 as uuid } from "uuid";
 
-export class DataSet {
+export class Dataset {
   constructor(underlayName: string, groups: Group[] = []) {
```
calbach, line 4: nit: this could also use the constructor shorthand

PR #47, opened 2021-10-14
Add ability to add criteria

tjennison-work:


ui/src/criteria/concept.tsx
```diff
@@ -80,22 +87,26 @@ function ConceptEdit(props: ConceptEditProps) {
...
-                  } else if (!!v.int64Val) {
-                    row[k] = v.int64Val;
-                  } else if (!!v.boolVal) {
-                    row[k] = v.boolVal;
-                  } else {
-                    row[k] = v.stringVal;
+              res.instances
+                .map((instance) => {
+                  const row: Row = {
+                    id: instance["concept_id"]?.int64Val || 0,
```
calbach, line 93: Wow, first time seeing this, seems null-safe chaining was added in TS 3.7, which I've just now realized Workbench can use as of a recent upgrade. Thanks!

ui/src/criteria/concept.tsx
```diff
@@ -80,22 +87,26 @@ function ConceptEdit(props: ConceptEditProps) {
...
+                  };
+                  for (const k in instance) {
+                    const v = instance[k];
+                    if (!v) {
+                      row[k] = "";
+                    } else if (!!v.int64Val) {
+                      row[k] = v.int64Val;
+                    } else if (!!v.boolVal) {
```
calbach, line 101: Existing issue, but this is a bug - `false` is a legitimate value for this field. I would probably switch all of these checks to explicitly `!==` against null or undefined to avoid surprises here (true for int64Val as well: `0` will skip that conditional, in this current code)

ui/src/updaterContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export function useUpdater(): Updater {
+  const updater = useContext(UpdaterContext);
+  if (!updater) {
+    throw new Error("invalid updater context");
+  }
+  return updater;
+}
+
+class Updater {
```
calbach, line 15: This could use some comments, or a more specific name

ui/src/criteria/concept.tsx
```diff
@@ -80,22 +87,26 @@ function ConceptEdit(props: ConceptEditProps) {
...
+                    } else {
+                      row[k] = v.stringVal;
+                    }
                   }
-                }
-                return row;
-              })
+                  return row;
+                })
+                .filter((row) => row.id != 0)
```
calbach, line 109: nit: I recommend using === or !== only, the truth table for == is mind-bending: https://dorey.github.io/JavaScript-Equality-Table/

ui/src/datasetUpdaterContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+      const criteria = group?.findCriteria(criteriaId);
+      if (!criteria) {
+        throw new Error("criteria not found");
+      }
+      callback(criteria as Type);
+    });
+  }
+}
+
+type UpdaterProviderProps = {
```
calbach, line 43: nit: DatasetUpdaterProviderProps? DatasetUpdaterProvider?

ui/src/updaterContext.tsx
```diff
@@ -0,0 +1,60 @@
+import * as _ from "lodash";
+import { createContext, ReactNode, useContext, useMemo } from "react";
+import { Criteria, Dataset } from "./dataset";
+
+const UpdaterContext = createContext<Updater | null>(null);
```
calbach, line 5: I assume you meant to delete this file

calbach (approved):

ui/src/updaterContext.tsx
tjennison-google: Yep, deleted.

ui/src/datasetUpdaterContext.tsx
tjennison-google: Done. Not sure how I missed those.

PR #50, opened 2021-10-15
Change concept_code type to STRING

tjennison-google:
Concept codes sometimes include letters.


api/src/main/resources/underlays/synpuf.yaml
```diff
@@ -31,7 +31,7 @@ entities:
       - name: condition_standard
         dataType: STRING
       - name: condition_concept_code
-        dataType: INT64
+        dataType: STRING
```
calbach, line 34: Is this a string in the source BigQuery dataset? I suspect this will cause issues if not.

PR #52, opened 2021-10-15
Redirect to new criteria and away from unknown ones

tjennison-google:


ui/src/overview.tsx
```diff
@@ -60,19 +64,24 @@ function AddCriteriaButton(props: { group: string | GroupKind }) {
...
+  const history = useHistory();
   const updater = useDatasetUpdater();
 
   const onAddCriteria = useCallback(
     (create: () => Criteria) => {
       addCriteriaState.close();
 
+      let groupId = "";
+      const criteria = create();
       updater.update((dataset: Dataset) => {
```
calbach, line 76: Presumably `updater.update()` is guaranteed to execute synchrously?

ui/src/overview.tsx
```diff
@@ -60,19 +64,24 @@ function AddCriteriaButton(props: { group: string | GroupKind }) {
...
         if (typeof props.group === "string") {
-          dataset.addCriteria(props.group, create());
+          dataset.addCriteria(props.group, criteria);
+          groupId = props.group;
         } else {
-          dataset.addGroupAndCriteria(props.group, create());
+          groupId = dataset.addGroupAndCriteria(props.group, criteria);
```
calbach, line 81: Could you add a comment to this component to explain the multiplexing going on here? Is this `string | GroupKind` union to support both editing a criteria as well as a adding a new one? If yes - would it make sense to just split this into two parameters instead, e.g:

```
groupKind: GroupKind, existingGroup?: string
```

tjennison-google: Yes, I've added a comment. If I use two parameters, then both are optional and I have to check for the existence of neither and both. This way the compiler guarantees exactly one is set.

calbach (approved):

PR #54, opened 2021-10-20
Add a dialog to display generated SQL

tjennison-work:


ui/src/sqlDialog.tsx
```diff
@@ -0,0 +1,88 @@
...
+          );
+      } else {
+        setError(new Error("No criteria have been selected"));
+      }
+    }
+  }, [api, props.dataset, open]);
+
+  return (
+    <>
+      <IconButton
```
calbach, line 55: I'm more accustomed to a dialog component just being a dialog (not including a button to open/close it). This would allow the dialog to be opened from different contexts, e.g. from a menu but also from a button.

ui/src/sqlDialog.tsx
```diff
@@ -0,0 +1,88 @@
...
+};
+
+export default function SqlDialog(props: SqlDialogProps) {
+  const [open, setOpen] = useState(false);
+  const [sql, setSql] = React.useState<string | null>(null);
+  const [error, setError] = useState<Error | null>(null);
+
+  const api = useContext(EntityInstancesApiContext);
+
+  useEffect(() => {
```
calbach, line 22: nit: should this use the async / fetch pattern (I believe you added this elsewhere), e.g. for XHR cleanup on navigate?

ui/src/sqlDialog.tsx
```diff
@@ -0,0 +1,88 @@
...
+    if (open) {
+      const params = props.dataset.generateQueryParameters();
+      if (!!params) {
+        api
+          .generateDatasetSqlQuery({
```
calbach, line 27: optional: `async` / `await` are more in vogue and can result in slightly more straight-line code. That said, you do need some trickery because `useEffect` doesn't support an async function param, usually looks something like this:

```
(async () => {
  const resp = await api.generateDatasetSqlQuery(...)
})();
```

ui/src/sqlDialog.tsx
```diff
@@ -0,0 +1,88 @@
...
+import React, { useContext, useEffect, useState } from "react";
+import { Dataset } from "./dataset";
+
+type SqlDialogProps = {
+  dataset: Dataset;
+};
+
+export default function SqlDialog(props: SqlDialogProps) {
+  const [open, setOpen] = useState(false);
+  const [sql, setSql] = React.useState<string | null>(null);
```
calbach, line 17: nit: unneeded `React.`

ui/src/criteria/concept.tsx
```diff
@@ -43,6 +44,42 @@ export class ConceptCriteria extends Criteria {
     return <ConceptDetails criteria={this} />;
   }
 
+  generateFilter(): tanagra.Filter | null {
+    if (this.selected.length === 0) {
+      return null;
+    }
+
+    const operands = new Array<tanagra.Filter>();
+    this.selected.forEach((row) => {
```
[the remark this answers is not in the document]
tjennison-google: Done. I don't know why I wasn't using map here.

ui/src/sqlDialog.tsx
```diff
@@ -0,0 +1,88 @@
...
+        aria-describedby="scroll-dialog-description"
+      >
+        <DialogTitle id="scroll-dialog-title">SQL Query</DialogTitle>
+        <DialogContent dividers={true}>
+          <DialogContentText
+            id="scroll-dialog-description"
+            tabIndex={-1}
+            sx={{ fontFamily: "Monospace" }}
+          >
+            {!!error ? "Failed: " + error.message : !!sql ? sql : "Loading..."}
```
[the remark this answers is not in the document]
tjennison-google: Done, and that was helpful here.

ui/src/sqlDialog.tsx
tjennison-google: It's a little weird because of how state works and it's not trivial to access to other components. What do you think about this update?

ui/src/sqlDialog.tsx
```diff
@@ -0,0 +1,97 @@
...
+                }
+              );
+          } else {
+            setError(new Error("No criteria have been selected"));
+          }
+        }
+      }, [api, props.dataset, open]);
+
+      return (
+        <>
```
calbach, line 67: Do you still need the fragment wrapper?

ui/src/criteria/concept.tsx
```diff
@@ -43,6 +44,41 @@ export class ConceptCriteria extends Criteria {
...
+            int64Val: row.id,
+          },
+        },
+      };
+    });
+
+    return {
+      relationshipFilter: {
+        outerVariable: "p",
+        newVariable: "cc",
```
calbach, line 70: I'm wondering whether having a hardcoded new variable name here is going to work, in particular if you start nesting or have an array of filters; might be fine though, I'd need to re-reviewing the variable declaration logic in the API

ui/src/actionBar.tsx
```diff
@@ -1,19 +1,24 @@
...
+import { useSqlDialog } from "./sqlDialog";
 
 type ActionBarProps = {
   title: string;
   backUrl?: string;
+  dataset: Dataset;
 };
 
 export default function ActionBar(props: ActionBarProps) {
+  const [dialog, show] = useSqlDialog({ dataset: props.dataset });
```
calbach, line 20: nit: maybe `showDialog`? In the scope of the action bar component, it's not clear what a variable named `show` might be showing

calbach (approved):

PR #56, opened 2021-10-21
Replace material-ui-popup-state with custom menu

tjennison-work:
It requires clunky code and interacts poorly with other theming.

### DIFF
--- a/ui/src/menu.tsx
+++ b/ui/src/menu.tsx
@@ -1,0 +1,35 @@
+import {
+  default as BaseMenu,
+  MenuProps as BaseMenuProps,
+} from "@mui/material/Menu";
+import { MouseEvent, ReactElement, useCallback, useState } from "react";
+
+type MenuProps = Omit<
+  BaseMenuProps,
+  "anchorEl" | "open" | "onClick" | "onClose"
+>;
+
+export function useMenu({
+  children,
+  ...props
+}: MenuProps): [
+  ReactElement,
+  (e: MouseEvent<HTMLElement | undefined>) => void
+] {
+  const [anchorEl, setAnchorEl] = useState<HTMLElement | undefined>();
+  const show = useCallback((e) => setAnchorEl(e.currentTarget), []);
+
+  return [
+    // eslint-disable-next-line react/jsx-key
+    <BaseMenu
+      {...props}
+      anchorEl={anchorEl}
+      open={!!anchorEl}
+      onClick={() => setAnchorEl(undefined)}
+      onClose={() => setAnchorEl(undefined)}
+    >
+      {children}
+    </BaseMenu>,
+    show,
+  ];
+}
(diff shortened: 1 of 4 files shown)


ui/src/menu.tsx
```diff
@@ -0,0 +1,45 @@
...
+  useCallback,
+  useState,
+} from "react";
+
+interface MenuProps
+  extends Omit<BaseMenuProps, "anchorEl" | "open" | "onClick" | "onClose"> {
+  items: ReactNode;
+}
+
+export function useMenu(
```
calbach, line 19: Interesting, are you basing this on an existing pattern or style guide from somewhere? I've not seen this "hook returning a component" pattern before, but from searching around I do see some mentions of it online.

calbach (approved):

ui/src/overview.tsx
```diff
@@ -140,36 +130,36 @@ function ParticipantsGroup(props: { group: Group }) {
...
-    variant: "popover",
-    popupId: "criteria",
-  });
-
   const updater = useDatasetUpdater();
 
+  const [menu, show] = useMenu({
+    items: [
+      <MenuItem
+        key="1"
```
cbookg, line 138: do you need these keys?

ui/src/overview.tsx
```diff
@@ -98,25 +86,27 @@ function AddCriteriaButton(props: { group: string | GroupKind }) {
     },
   ];
 
+  const [menu, show] = useMenu({
+    items: items.map((item) => {
+      return (
```
cbookg, line 91: can be reduced to:  (item) => (
  <MenuItem  ... >
)

ui/src/menu.tsx
```diff
@@ -0,0 +1,45 @@
...
+  props: MenuProps
+): [ReactElement, (e: MouseEvent<HTMLElement | undefined>) => void] {
+  const [anchorEl, setAnchorEl] = useState<HTMLElement | undefined>();
+  const show = useCallback((e) => setAnchorEl(e.currentTarget), []);
+
+  const menu = useCallback(
+    (props: MenuProps): ReactElement => {
+      return (
+        <>
```
cbookg, line 28: drop this and use BaseMenu as top level?

ui/src/menu.tsx
```diff
@@ -0,0 +1,45 @@
...
+}
+
+export function useMenu(
+  props: MenuProps
+): [ReactElement, (e: MouseEvent<HTMLElement | undefined>) => void] {
+  const [anchorEl, setAnchorEl] = useState<HTMLElement | undefined>();
+  const show = useCallback((e) => setAnchorEl(e.currentTarget), []);
+
+  const menu = useCallback(
+    (props: MenuProps): ReactElement => {
```
cbookg, line 26: just => ( <element/> )

ui/src/menu.tsx
```diff
@@ -0,0 +1,45 @@
...
+          >
+            {props.items}
+          </BaseMenu>
+        </>
+      );
+    },
+    [props]
+  );
+
+  return [createElement(menu, props, null), show];
```
cbookg, line 44: im trying to understand this, but not quite following based on the docs, which have the first param being a type (but you're padding a callback)?

mainly curious when the callback would be invoked, and how its not run each time.

PR #74, opened 2021-11-05
Use useAsync instead of useEffect

tjennison-google:


ui/src/loading.tsx
```diff
@@ -1,26 +1,65 @@
...
 
-export default function Loading() {
+type State = {
+  error?: Error;
+  isPending: boolean;
+  reload: () => void;
+};
+
+type Props = {
+  state?: State;
```
calbach, line 14: I would avoid having a prop called `state`. I realize you probably don't mean to imply "React state" with this naming, I would guess most readers make that assumption. In general I would not recommend using the term state in the vicinity of a React component unless it's actually referring to React state.

Something like "status" maybe.

ui/src/loading.tsx
```diff
@@ -1,26 +1,65 @@
+import Box from "@mui/material/Box";
+import Button from "@mui/material/Button";
 import CircularProgress from "@mui/material/CircularProgress";
-import Grid from "@mui/material/Grid";
-import React, { useEffect, useRef, useState } from "react";
+import Typography from "@mui/material/Typography";
+import React, { ReactNode, useEffect, useRef, useState } from "react";
 
-export default function Loading() {
+type State = {
```
calbach, line 7: opt: I don't know if it's necessary to split out this definition into its own type. I haven't done too much experimentation in this area, but pretty sure you can just inline the object type into Props below, if you preferred that.

ui/src/loading.tsx
tjennison-google: Done.

ui/src/loading.tsx
tjennison-google: I'm using it below for showStatus.

calbach (approved):

PR #77, opened 2021-11-15
Add GitHub action for UI tests

tjennison-google:


.github/workflows/ui-test.yaml
```diff
@@ -0,0 +1,27 @@
+name: UI Integration Tests
```
calbach, line 1: These seem like unit tests to me

calbach (commented):
Presumably a path misconfiguration somewhere based on the failaure logs

.github/workflows/ui-test.yaml
tjennison-google: Yes, I think you're right.

calbach (approved):

PR #78, opened 2021-11-22
Add test for Loading component

tjennison-google:


ui/src/loading.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+import Loading from "./loading";
+
+test("loading", async () => {
+  const reload = jest.fn();
+  const props = { isPending: true, reload: reload };
+  const loading = () => <Loading status={props}>loaded</Loading>;
+
+  const { rerender } = render(loading());
+  await waitFor(() =>
+    expect(screen.getByRole("progressbar")).toBeInTheDocument()
```
cbookg, line 13: you don't need to use waitFor if you replace getByRole with findByRole.  not sure if you just find this more clear.

also the "toBeInTheDocument" seems maybe recommended, but doesn't matter.  ie the entire thing could be replace by:
await screen.findByRole("progressbar");

ui/src/loading.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+  );
+
+  expect(screen.queryByText("Reload")).not.toBeInTheDocument();
+
+  props.isPending = false;
+  props.error = new Error("test-error");
+  rerender(loading());
```
cbookg, line 20: instead of storing/updating props, how about just remaking the whole thing since its really a new test?  somethng like:
rerender(<Loading status={{isPending: false, error: new Error()}}/>);

is there a need to store the component in a variable?

tjennison-google: Per our discussion this morning, looks like inlining the JSX is just as effective and seems cleaner in the end.

ui/src/loading.test.tsx
tjennison-google: Nice, that's much shorter.

tjennison-google: Nice, that's much shorter.

cbookg (approved):

PR #79, opened 2021-11-29
Add test for menu component

tjennison-google:


ui/src/menu.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+    <>
+      <Button onClick={show}>show</Button>
+      {menu}
+    </>
+  );
+}
+
+test("show/hide menu", async () => {
+  const onClick = jest.fn();
+  render(<MenuButton onClick={onClick} />);
```
cbookg, line 26: naming here is a bit weird, because normally a button's onClick is when you click on the button, but in this case its on the menu item being clicked.  maybe onMenuItem?

ui/src/menu.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+import MenuItem from "@mui/material/MenuItem";
+import { render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import React from "react";
+import { useMenu } from "./menu";
+
+function MenuButton(props: { onClick: () => void }) {
+  const [menu, show] = useMenu({
+    children: [
+      <MenuItem key="1" onClick={props.onClick}>
```
cbookg, line 11: since you only have one child, I think you can drop the [], pass <MenuItem>...</MenuItem directly to children, and remove the key.

ui/src/menu.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+      {menu}
+    </>
+  );
+}
+
+test("show/hide menu", async () => {
+  const onClick = jest.fn();
+  render(<MenuButton onClick={onClick} />);
+
+  userEvent.click(screen.getByText("show"));
```
cbookg, line 28: screen.getByRole("button") or screen.getByRole("button", {name: "show"}) preferred?

ie docs say to prefer getByRole over getByText when possible.

ui/src/menu.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  userEvent.click(screen.getByText("menu-item"));
+  expect(onClick).toHaveBeenCalled();
+
+  await waitFor(() =>
```
cbookg, line 33: use waitForElementToBeRemoved:
https://testing-library.com/docs/dom-testing-library/api-async/#waitforelementtoberemoved

await waitForElementToBeRemoved(() =>  screen.queryByText("menu-item"));

calbach (commented):
Some good comments here, review seems well covered so I will remove myself - let me know if there is something you want me to look at specifically.

**Edit**: I guess I can't remove myself in GitHub since I just commented.. you get the point.

ui/src/menu.test.tsx
tjennison-google: Done

ui/src/menu.test.tsx
tjennison-google: Done

ui/src/menu.test.tsx
tjennison-google: Done

ui/src/menu.test.tsx
tjennison-google: Done

cbookg (approved):

PR #80, opened 2021-11-29
Add test for CohortUpdater

tjennison-google:


ui/src/cohortUpdaterContext.test.tsx
```diff
@@ -0,0 +1,38 @@
+import { Cohort, GroupKind } from "cohort";
+import { ConceptCriteria } from "criteria/concept";
+import { CohortUpdater } from "./cohortUpdaterContext";
+
+test("updates", () => {
+  var cohort = new Cohort("name", "underlay", "entity");
```
calbach, line 6: I don't think there are any cases where one should use `var` in ES6, as far as I'm aware. Would use `let` for any non-final variables

ui/src/cohortUpdaterContext.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+      new ConceptCriteria("test-name", "test-filter")
+    );
+  });
+
+  expect(cohort).not.toBe(prevCohort);
+  expect(cohort.groups.length).toBe(1);
+  expect(cohort.groups[0].criteria.length).toBe(1);
+  expect(cohort.groups[0].criteria[0].name).toBe("test-name");
+
+  const group = cohort.groups[0];
```
calbach, line 26: opt: feels like this could be probably be a separate test case

calbach (approved):

ui/src/cohortUpdaterContext.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+  new CohortUpdater(cohort, setCohort).updateCriteria(
+    group.id,
+    group.criteria[0].id,
+    (criteria: ConceptCriteria) => {
+      criteria.name = "new-test-name";
+    }
+  );
+
+  expect(cohort.groups.length).toBe(1);
+  expect(cohort.groups[0].criteria.length).toBe(1);
```
cbookg, line 36: just a suggestion, but there is probably a more jest-y way of doing this...

expect(cohort.groups[0].criteria.length).toBe([expect.objectContaining[{name: "new-test-name"}])

or even more top level.

PR #81, opened 2021-12-02
Add tests for concept criteria

tjennison-google:


ui/src/criteria/concept.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+  );
+
+  const { rerender } = render(components());
+
+  await screen.findByText("Concept Name");
+  return {
+    getCriteria: getCriteria,
+    rerender: () => {
+      rerender(components());
+    },
```
cbookg, line 152: drop braces:
rerender: () => rerender(components()),

ui/src/criteria/concept.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+  ]);
+
+  const checkboxes = screen.getAllByRole("checkbox");
+  expect(checkboxes.length).toBe(2);
+
+  expect(getCriteria().selected.length).toBe(0);
+
+  const getSelected = () => {
+    return getCriteria().selected.map((row) => row.id);
```
cbookg, line 83: simplify? :
const getSelected = () => getCriteria().selected.map((row) => row.id);

ui/src/criteria/concept.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+  };
+
+  // Use waitFor because DataGrid has a focus update that occurs after the event
+  // (i.e. outside of act) which causes a warning.
+  await waitFor(async () => {
+    userEvent.click(checkboxes[0]);
```
cbookg, line 89: i don't think you want an event inside a waitFor.  i think if you had the linter enabled it would complain about this.

can you put the click first and then waitFor the expect?

ui/src/criteria/concept.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+      },
+      matches: ["0-0 of 0"],
+    },
+  ],
+])("%s", async (name, { instance, matches }) => {
+  await renderCriteria([instance]);
+
+  matches.forEach((match) => {
+    screen.getByText(match);
+  });
```
cbookg, line 60: matches.forEach((match) => screen.getByText(match)) ?

ui/src/criteria/concept.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+        },
+      },
+      matches: ["0-0 of 0"],
+    },
+  ],
+])("%s", async (name, { instance, matches }) => {
+  await renderCriteria([instance]);
+
+  matches.forEach((match) => {
+    screen.getByText(match);
```
calbach, line 59: Some of these questions may be naive, I haven't used the react test renderer library extensively:

- On the surface, this test is not making any assertions, and this line looks like a no-op. Is there some undocumented (AFAICT) side-effect of `getByText()`, e.g. that it throws an error? It may be more readable to actually assert something anyways..
- From the docs, it seems that `getByText` is an async function -is that not the case here, and if so shouldn't there be some awaits in this test?

ui/src/criteria/concept.test.tsx
```diff
@@ -0,0 +1,154 @@
+import { render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { EntityInstancesApiContext } from "apiContext";
+import { Cohort, Group, GroupKind } from "cohort";
+import { CohortUpdaterProvider } from "cohortUpdaterContext";
+import { ConceptCriteria } from "criteria/concept";
+import * as tanagra from "./tanagra-api";
+
+test.each([
```
calbach, line 9: Nice

ui/src/criteria/concept.test.tsx
tjennison-google: Done.

ui/src/criteria/concept.test.tsx
tjennison-google: Done.

ui/src/criteria/concept.test.tsx
tjennison-google: That's exactly what the comment is about. The click ends up calling setCellFocus sometime after the initial event which results in "Warning: An update to ForwardRef inside a test was not wrapped in act(...)." I've changed it to act() instead which also seems to be sufficient and doesn't require await but I don't otherwise know how to get rid of it.

ui/src/criteria/concept.test.tsx
tjennison-google: Done.

ui/src/criteria/concept.test.tsx
tjennison-google: - getByText throws if it can't find a matching element (see the summary table at https://testing-library.com/docs/queries/about). So the full assertion could be written as `expect(getByText(match)).toBeInDocument()` but that doesn't seem to actually add any value since it's functionally equivalent. 
- getByText is the non-async version. findByText (like at line 142) is async.

ui/src/criteria/concept.test.tsx
cbookg: act(() => userEvent.click(checkboxes[0])); ?

I found userEvent to be a bit weird.  fireEvent never needs an act, it does that for you. userEvent supports some more complex interactions, and it seems like older versions needed to wrap in act, but newer versions didn't.  are you using the latest?  I don't think i needed the act in my code anymore (v13.5.0)

cbookg (approved):

ui/src/criteria/concept.test.tsx
tjennison-google: Done.

It doesn't usually need the act but the data grid is doing something special that causes an extra update somehow and it isn't being handled automatically.

ui/src/criteria/concept.tsx
```diff
@@ -212,12 +212,17 @@ function ConceptEdit(props: ConceptEditProps) {
           columns={columns}
           disableSelectionOnClick
           className="criteria-concept"
+          disableVirtualization={!!process.env.JEST_WORKER_ID}
```
calbach, line 215: This is not a great code smell; assuming you've already explored better workarounds, I would put a comment here explaining why this is necessary.

calbach (approved):

ui/src/criteria/concept.tsx
tjennison-google: Added a comment. jsdom, which jest uses for testing, has incomplete layout sizing support so virtualization ends up not displaying most of the columns.

PR #83, opened 2021-12-17
Allow defining an entity on a filtered version of a table.

marikomedlock:
Previously, an entity maps to a full SQL table. This change allows an entity to map to a filtered SQL table.

The purpose of this change is to allow defining separate entities for the different domains in the OMOP `concept` table. (e.g. a "conditions" entity, separate from a "procedures" entity, separate from a "drugs" entity).


api/src/main/java/bio/terra/tanagra/service/underlay/ColumnFilter.java
```diff
@@ -0,0 +1,31 @@
+package bio.terra.tanagra.service.underlay;
+
+import com.google.auto.value.AutoValue;
+
+/** A filter on a SQL column within a table in an underlay. */
+@AutoValue
+public abstract class ColumnFilter {
+  /** The column to filter on. */
+  public abstract Column column();
```
marikomedlock, line 9: Here's where we define a `Column` property in the Java class for `ColumnFilter`.

lib/src/main/proto/bio/terra/tanagra/underlay/tables.proto
```diff
@@ -48,3 +48,18 @@ message ColumnId {
...
+message TableFilter {
+  // A condition on a particular column
+  // TODO: allow multiple ColumnFilters and AND/OR'ing them together
+  ColumnFilter column_filter = 1;
+}
+
+// A building block for a TableFilter that defines a condition on a single column.
+message ColumnFilter {
+  // Pointer to the column.
+  ColumnId column = 1;
```
marikomedlock, line 62: Here's where we define a `ColumnId` property in the proto definition for `ColumnFilter`.

calbach: Defining these as Column IDs and then resolving into a richer Column representation in the Autovalue table seems reasonable to me. Looks consistent with what is being done elsewhere in the underlay conversion, e.g. hierarchical references. Am I missing something?

lib/src/main/proto/bio/terra/tanagra/underlay/tables.proto
```diff
@@ -48,3 +48,18 @@ message ColumnId {
...
+  // Column value to filter on
+  string value = 2;
```
calbach, line 64: nit: I think it may be clearer to model this as a oneof to indicate future extension (with just `string_value` for now). The filter here I think can be parallel to what is defined by the API for entity search: https://github.com/DataBiosphere/tanagra/blob/328febf90e7f8a5c31b7ab8258d5a1bcbdb0bcb2/api/src/main/resources/api/service_openapi.yaml#L316-L329

lib/src/main/proto/bio/terra/tanagra/underlay/tables.proto
```diff
@@ -48,3 +48,18 @@ message ColumnId {
   // The name of the column.
   string column = 3;
 }
+
+// A filter specification to get a subset of the rows in a table.
+message TableFilter {
+  // A condition on a particular column
+  // TODO: allow multiple ColumnFilters and AND/OR'ing them together
```
calbach, line 55: For future reference, mirroring the entity filter API here, but for tables rather than entities, would be a consistent way of doing this: https://github.com/DataBiosphere/tanagra/blob/328febf90e7f8a5c31b7ab8258d5a1bcbdb0bcb2/api/src/main/resources/api/service_openapi.yaml#L439

calbach, line 55: For future reference, mirroring the entity filter API here, but for tables rather than entities, could be a consistent way of doing this: https://github.com/DataBiosphere/tanagra/blob/328febf90e7f8a5c31b7ab8258d5a1bcbdb0bcb2/api/src/main/resources/api/service_openapi.yaml#L439

calbach (commented):
Approach looks reasonable. While thinking this over a bit, I'm noticing that the same goal could also be achieved by creating a filtered view on the table, and then just mapping entities to views rather than the original table, e.g. creating a condition_concept view. However, the approach here I think is a bit more flexible as it won't depend on view capabilities or deciding where to create the views and when during the dataset onboarding process, and Tanagra already needs to be capable of filtering the tables.

lib/src/main/proto/bio/terra/tanagra/underlay/tables.proto
marikomedlock: Thanks for the pointer, I've not used protos before so glad to learn about `oneof`. Changed the `ColumnValue` to accommodate both strings and ints, to be consistent with the two currently supported column data types.

lib/src/main/proto/bio/terra/tanagra/underlay/tables.proto
marikomedlock: Thanks for the pointer here, too. Changed the `TableFilter` to accept `oneof` the available filters. Currently, there is only the `BinaryColumnFilter` implemented, but we can add new ones here.

marikomedlock: duplicate comment, closing.

marikomedlock:
> Approach looks reasonable. While thinking this over a bit, I'm noticing that the same goal could also be achieved by creating a filtered view on the table, and then just mapping entities to views rather than the original table, e.g. creating a condition_concept view. However, the approach here I think is a bit more flexible as it won't depend on view capabilities or deciding where to create the views and when during the dataset onboarding process, and Tanagra already needs to be capable of filtering the tables.
> 
> The advantage of using a view would be simplification of the main Tanagra filtering code (not by much), but also the ability to take advantage of things like materialized views. The complexity of creating and managing those views may be too high, though.

In the future when we start looking at query optimization, we may want to always define entities as mapping to a single unfiltered table or view. We could do that in the indexing process by looking at the underlay and generating a new table or view (materialized or not) for any entity with a `TableFilter` defined. But while that may be more performant, I think the flexibility in defining new entities is more useful right now, so we can iterate on the entity model and queries without having to touch the indexing layer.

lib/src/main/proto/bio/terra/tanagra/underlay/tables.proto
marikomedlock: Yeah I was confused why there are both `Column` and `ColumnId` Autovalue classes. But I think it's just easier if the serialization classes generally line up. e.g. We could have a `Column` class that has an `id` property, which we use wherever the `ColumnId` is in the underlay proto, instead of deserializing into two separate objects. I can see how that could be harder to follow the deserialization flow from underlay -> internal classes though.

tjennison-google (approved):

api/src/test/java/bio/terra/tanagra/service/query/BinaryColumnFilterTest.java
```diff
@@ -0,0 +1,321 @@
...
+                    .build())
+            .build();
+
+    Exception exception =
+        assertThrows(
+            IllegalArgumentException.class,
+            () -> convertUnderlayFromProto(tableFilterProto),
+            "exception thrown when binary column filter value doesn't match the column data type");
+    LOG.info("underlay conversion exception message: {}", exception.getMessage());
+    assertTrue(
```
calbach, line 193: nit: suggest removing the log and using the more semantic substring assertion (throughout), similar to this: https://github.com/DataBiosphere/tanagra/blob/10c982b25f5fd437dd884b90fe57c9600db6031b/api/src/test/java/bio/terra/tanagra/service/query/api/ApiConversionServiceTest.java#L75

Personally I have a slight preference for the Truth library, but this is the current pattern.

calbach (approved):
Nice, thanks

api/src/main/java/bio/terra/tanagra/service/query/QueryService.java
```diff
@@ -66,7 +66,7 @@ public QueryResult retrieveResults(EntityDataset entityDataset) {
         .execute(query, SearchContext.builder().underlay(underlay).build());
   }
 
-  private Query createQuery(EntityDataset entityDataset) {
+  Query createQuery(EntityDataset entityDataset) {
```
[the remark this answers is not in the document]
marikomedlock: Added `@VisibleForTesting`.

`generateSql()` requires that the underlay is loaded in the application. This is fine for testing e.g. the nautical underlay, but for this test, it was more convenient to define the underlay programmatically in the test and then pass it to the `createQuery` method instead.

api/src/test/java/bio/terra/tanagra/service/query/BinaryColumnFilterTest.java
marikomedlock: done.

PR #84, opened 2021-12-23
Add condition entity for aou_synthetic underlay. Tests for entity queries.

marikomedlock:
First pass at defining an entity for each partition of the `concept` table.
- `condition` entity filters on `domain_id='Condition'`
- `procedure` entity filters on `domain_id='Procedure'`
- `measurement` entity filters on `domain_id='Measurement'`
- `visit` entity filters on `vocabulary_id='Visit'`
- `observation` entity filters on `domain_id='Observation'`
- `device` entity filters on `domain_id='Device'`
- `ingredient` entity filters on `concept_class_id='Ingredient'`
- `brand` entity filters on `concept_class_id='Brand Name'`

All of these entities should be filtering on more than just the one field (e.g. `valid_end_date>='2022-01-01'`), but right now the `TableFilter` property on the `EntityMapping` in the underlay only supports filtering on a single property. A subsequent PR will add additional column filter(s) for each of the above entities.

Also added tests for querying entity instances using the `aou_synthetic` underlay. I tested these queries by running them manually in BQ in the `broad-tanagra-dev` project. The results don't exactly match what the AoU cohort builder returns, so another follow-on here is to find the exact queries for each page in the AoU cohort builder and compare them to what these return.


tjennison-google (approved):

api/src/main/resources/underlays/aou_synthetic.yaml
```diff
@@ -249,6 +455,390 @@ attributeMappings:
         dataset: aou_synthetic
         table: concept
         column: concept_name
+  - attribute:
+      entity: condition
```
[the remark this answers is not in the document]
marikomedlock: I agree. It was a lot of repetition to define the new entities. Not sure what the right approach is to making this underlay less verbose/more readable, but seems like a lower priority for the moment.

api/src/test/java/bio/terra/tanagra/aousynthetic/DeviceEntityQueries.java
```diff
@@ -0,0 +1,60 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+
+/**
+ * Tests for device entity queries on the AoU synthetic underlay. There is no need to specify an
+ * active profile for this test, because we want to test the main application definition.
+ */
+public class DeviceEntityQueries extends BaseSpringUnitTest {
```
calbach, line 26: nit: existing pattern in the codebase is to suffix tests with `Test` in the classname

api/src/test/java/bio/terra/tanagra/aousynthetic/DeviceEntityQueries.java
```diff
@@ -0,0 +1,60 @@
...
+            DEVICE_ENTITY,
+            new ApiGenerateDatasetSqlQueryRequest()
+                .entityDataset(
+                    new ApiEntityDataset()
+                        .entityVariable("device_alias")
+                        .selectedAttributes(ALL_DEVICE_ATTRIBUTES)));
+    assertEquals(HttpStatus.OK, response.getStatusCode());
+    String generatedSql = response.getBody().getQuery();
+    LOG.info(generatedSql);
+    assertEquals(
```
calbach, line 45: opt: this PR introduces a bunch of tests that depend on exact SQL string matches. This is configured in such a way that an innocuous change in the engine code could result in a large number of tedious Java string updates across many files.

Some alternatives to consider:
1. Use a golden output files pattern here for the SQL text. Make it easy to regenerate the golden files programmatically.
2. Don't check the SQL string. Execute the queries and make targeted assertions against the responses (or use the API to do this, rather than executing it ourselves). If we're not there yet, maybe just don't bother checking the SQL string at all for now (or just look for a short substring to confirm you got the right thing).

calbach (approved):
Thanks.

> All of these entities should be filtering on more than just the one field (e.g. valid_end_date>='2022-01-01'),

Can you add more detail around this? What are the concrete requirements for additional table-level filters we need at the Underlay level? Was date a real example?

api/src/test/java/bio/terra/tanagra/aousynthetic/DeviceEntityQueries.java
marikomedlock: done.

api/src/test/java/bio/terra/tanagra/aousynthetic/GeneratedSqlUtils.java
```diff
@@ -0,0 +1,84 @@
...
+   * @param fileName the name of the file in the `api/src/test/resources/aousynthetic` directory
+   *     where the generated SQL to compare against during testing lives.
+   */
+  public static void checkMatchesOrOverwriteGoldenFile(String generatedSql, String fileName)
+      throws IOException {
+    LOG.info(generatedSql);
+    if (System.getProperty("GENERATE_SQL_FILES") == null) {
+      LOG.info("reading generated sql from file because generateSqlFiles flag is not set");
+      String expectedSql = readSqlFromFile(fileName);
+      assertEquals(expectedSql, generatedSql);
```
calbach, line 51: opt: while it would slightly break separation of concerns for this code to know about gradle, it is definitely nice to include a copy/pasteable command line for golden file regeneration in the failure message here

api/src/test/java/bio/terra/tanagra/aousynthetic/GeneratedSqlUtils.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Utility class for:
+ *
+ * <p>- comparing the SQL generated by the API code against an expected string saved in a golden
+ * file.
+ *
+ * <p>- overwriting the existing golden file with a new expected string.
+ */
+public final class GeneratedSqlUtils {
```
calbach, line 23: nit: not necessarily specific to aousynthetic, could potentially live in a more general location. Fine to also just move later once we have a second use case.

calbach (approved):
Awesome! Thanks for running with the suggestion, I think this will be a lot easier to maintain.

api/src/test/java/bio/terra/tanagra/aousynthetic/GeneratedSqlUtils.java
marikomedlock: good idea. done.

api/src/test/java/bio/terra/tanagra/aousynthetic/GeneratedSqlUtils.java
marikomedlock: I was planning to just move it once we have a second use case, but then I noticed there's a generic-sounding `testing` package. Moved it there instead.

PR #85, opened 2022-01-12
Add new entity relationship type IntermediateTable.

marikomedlock:
Previously, there is one type of relationship between two entities, defined by a direct foreign key relationship between the two entity primary tables. This change adds a second type of relationship, defined by an intermediate table that has foreign key relationships with the two entity primary tables.

The purpose of this change is to allow defining a relationship between the "brand" and "ingredient" entities. The OMOP `concept_relationship` table holds the mapping of which ingredients are included in a brand.

### DIFF
--- a/lib/src/main/proto/bio/terra/tanagra/underlay/mapping.proto
+++ b/lib/src/main/proto/bio/terra/tanagra/underlay/mapping.proto
@@ -72,8 +72,23 @@ message RelationshipMapping {
     ColumnId foreign_key = 2;
   }
 
+  // A mapping where a relationship is navigated by a separate intermediate table.
+  // Each entity table has a column that foreign keys to a column in the intermediate table.
+  message IntermediateTable {
+    // The column in the entity1 table that FKs to the relationship table.
+    ColumnId entity1_entity_table_key = 1;
+    // The column in the relationship table that FKs to the entity1 table.
+    ColumnId entity1_intermediate_table_key = 2;
+
+    // The column in the entity2 table that FKs to the intermediate table.
+    ColumnId entity2_entity_table_key = 3;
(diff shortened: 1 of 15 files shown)


lib/src/main/proto/bio/terra/tanagra/underlay/mapping.proto
```diff
@@ -72,8 +72,23 @@ message RelationshipMapping {
     ColumnId foreign_key = 2;
   }
 
+  // A mapping where a relationship is navigated by a separate intermediate table.
```
calbach, line 75: How does this interact with foreign key mappings? Should the relationships of entity1 <-> intermediate, and/or entity2 <-> intermediate exist? Not exist? Doesn't matter?

calbach (approved):
Looks reasonable, thanks

lib/src/main/proto/bio/terra/tanagra/underlay/mapping.proto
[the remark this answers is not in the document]
marikomedlock: Yes, it should also be possible to express such a relationship (table 1 FKs to intermediate table FKs to table 2) using two FK mappings. As you say, this would require 3 entity definitions, one for each table.

The goal here was to allow defining this relationship with just the existing 2 entities. The intermediate table doesn't really mean anything to the UI, in the same way that the other two tables do. I wanted to avoid defining extra entities just based on the SQL schema.

marikomedlock: There is no intermediate entity, just an intermediate table, so there should only be one relationship: entity 1 <-> entity 2.

calbach (approved):

tjennison-google (approved):

PR #86, opened 2022-01-13
Add brand-ingredient entity relationship for aou_synthetic underlay.

marikomedlock:
Added a relationship between the "brand" and "ingredient" entities, defined by the intermediate table `concept_relationship`.

Added a test for getting all the ingredients for a particular brand. I tested the query by running it manually in BQ in the `broad-tanagra-dev` project.

In the future, I think we should add an additional filter on the intermediate table (e.g. `concept_relationship.relationship_id = 'Brand name of'`), but this will be in a separate PR once we have that support in the underlay schema.


api/src/test/java/bio/terra/tanagra/testing/GeneratedSqlUtils.java
```diff
@@ -48,10 +48,39 @@ private GeneratedSqlUtils() {}
...
   public static void checkMatchesOrOverwriteGoldenFile(String generatedSql, String fileName)
       throws IOException {
+    checkMatchesOrOverwriteGoldenFile(generatedSql, fileName, null);
+  }
+
+  /**
+   * This method is intended for SQL strings that include generated table aliases. This is the case
+   * for relationships that are defined by intermediate tables. If the relationshipName is not null,
+   * then before doing the comparison, this method replaces the generated table aliases in the
+   * expected SQL with those in the actual SQL. This is necessary because generated table names
```
calbach, line 58: I think it would also be reasonable to inject a seeded Random from the test, if that's not already happening. This might be simpler.

calbach (approved):

tjennison-google (approved):

api/src/test/java/bio/terra/tanagra/testing/GeneratedSqlUtils.java
marikomedlock: Good idea. I've done something similar before using an interface and annotating different implementation with different Spring profiles. Perhaps that approach will work here too. I'll plan to tackle that as a separate PR.

PR #87, opened 2022-01-14
Add ChildrenTable to hierarchy definition. Add CHILD_OF binary filter operator.

marikomedlock:
- Added a new required property for hierarchy definitions in the underlay: `ChildrenTable`. This table has one row for each parent-child relation. Included an optional `TableFilter`, so we can reuse the same table for multiple hierarchy definitions.
- This means we can use the existing OMOP `concept_relationship` table, filtered on the `relationship_id`, for this `ChildrenTable` property of the hierarchy definition. We don’t need to generate a new index table for each entity’s parent-child relations.
- Added a new query operator `CHILD_OF` to get the immediate children of an entity. Now we can generate the query needed by the UI: “list the immediate child ingredients of the opioids category”.

- Follow-on PR will define the hierarchy, including the new `ChildrenTable` property, for all concept-domain entities for the `aou_synthetic` dataset. This PR defines the `ChildrenTable` property for the existing `condition` entity to an existing table, just to satisfy the requirement that the property is defined. The definition as is is incorrect and will be fixed in the follow-on PR.


api/src/main/java/bio/terra/tanagra/service/search/SqlVisitor.java
```diff
@@ -261,6 +300,10 @@ private String resolveTable(Table table) {
     }
 
     private String resolveTable(Table table, TableFilter tableFilter) {
```
calbach, line 302: nit: `@Nullable` ?

api/src/main/java/bio/terra/tanagra/service/underlay/UnderlayConversion.java
```diff
@@ -541,22 +561,30 @@ private static ColumnId convert(bio.terra.tanagra.proto.underlay.ColumnId column
...
 
+  private static TableFilter convert(
+      bio.terra.tanagra.proto.underlay.TableFilter tableFilterProto,
+      Map<ColumnId, Column> columns) {
+    switch (tableFilterProto.getFilterCase()) {
+      case BINARY_COLUMN_FILTER:
+        return TableFilter.builder()
+            .binaryColumnFilter(convert(tableFilterProto.getBinaryColumnFilter(), columns))
+            .build();
+      case FILTER_NOT_SET:
```
calbach, line 572: nit: This is treated equivalently to `null` elsewhere: https://github.com/DataBiosphere/tanagra/blob/a9ad6ff4045b30a6e204c519b733c69961e31228/api/src/main/java/bio/terra/tanagra/service/underlay/UnderlayConversion.java#L189

calbach (approved):
Makes sense. I think an alternative approach could be to refactor the descendants table with additional schema details (and add more information to the index). We may want to consider this consolidation in this direction when we get to more performance indexing.

tjennison-google (approved):

api/src/main/java/bio/terra/tanagra/service/search/SqlVisitor.java
marikomedlock: added.

api/src/main/java/bio/terra/tanagra/service/underlay/UnderlayConversion.java
marikomedlock: Thanks for the catch, changed the validation switch statement to also throw on `FILTER_NOT_SET` so they're consistent.

PR #88, opened 2022-01-14
Add ChildrenTable for all concept-domain entities in aou_synthetic underlay.

marikomedlock:


render-config.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+## This script renders configuration files needed for development.
+## Dependencies: vault
+## Inputs: VAULT_TOKEN (arg, optional) default is $HOME/.vault-token
+## Usage: ./tools/render-config.sh
```
calbach, line 6: `./tools` is not a directory here

render-config.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+## This script renders configuration files needed for development.
```
calbach, line 3: Is the terminology borrowed from some other framework? Just seems like this is grabbing the SA key, and not doing much in the way of config rendering

render-config.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
```
calbach, line 1: Should this be under `/api`? Not sure there will be much of a need for this in the UI

workflow/worklog.md
```diff
@@ -18,3 +18,20 @@ Initial flattening of SNOMED conditions in aou synthetic:
 Need to export postgres password and service account credentials file.
 ```
 ./gradlew workflow:execute -DmainClass=bio.terra.tanagra.workflow.CopyBigQueryDatasetToPostgres -Dexec.args="--underlayYaml=api/src/main/resources/underlays/synpuf.yaml --cloudSqlInstanceName=broad-tanagra-dev:us-central1:tanagra-cloudsql-2cd088adcd745a91 --cloudSqlDatabaseName=indexes --cloudSqlUserName=tanagra --runner=dataflow --project=broad-tanagra-dev --region=us-central1 --serviceAccount=tanagra@broad-tanagra-dev.iam.gserviceaccount.com"```
+```
+
+## Generate hierarchy tables for all concept domain-based entities
+Fetch the tanagra-dev SA key file from Vault. (You may need to be on the Broad VPN to do this.)
```
calbach, line 24: Could you leave a link to the vault setup? I never remember how to initialize this

render-config.sh
marikomedlock: fixed.

render-config.sh
marikomedlock: Yes, this was a copy/paste error. We have a similar script in most of the Enterprise Terra git repos and it's generally called `render-config.sh`. I'm not sure where that name came from, but agreed it's not a great description. Changed the filename to `pull-credentials.sh` and updated the description.

workflow/worklog.md
marikomedlock: Added some links to setting up Vault credentials. Let me know if that's not what you meant about vault setup.

api/src/main/resources/underlays/aou_synthetic.yaml
```diff
@@ -1072,39 +1078,142 @@ hierarchies:
...
+        table: concept_relationship
+        column: concept_id_2
+      tableFilter:
+        binaryColumnFilter:
+          column:
+            dataset: aou_synthetic
+            table: concept_relationship
+            column: relationship_id
+          operator: EQUALS
+          stringVal: Subsumes
```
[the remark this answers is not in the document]
marikomedlock: I chose `Subsumes` because:
- The [previous SQL query](https://github.com/DataBiosphere/tanagra/blob/8bb26c00459fc3aafc75499c79859786ac918bdf/workflow/src/main/resources/queries/aou_omop_snomed_condition_relations.sql#L10) for generating the ancestor-descendant index table for conditions filtered on this relationship name.
- Comparing SQL query results with the AoU-CB behavior seemed to mostly match when I used `relationship_id='Subsumes'`.
- An example of using `Subsumes` in the [OMOP documentation](https://ohdsi.github.io/CommonDataModel/dataModelConventions.html#Concept_Relationships) seems to match what we want for the hierarchy definition.

I'll add this to the PR description also.

calbach (approved):
thanks!

PR #89, opened 2022-01-17
Add new treegrid component

tjennison-google:
This will replace the existing table views and also provide the basis
for the hierarchical view.

### DIFF
--- a/ui/src/treegrid.test.tsx
+++ b/ui/src/treegrid.test.tsx
@@ -1,0 +1,71 @@
+import AccountTreeIcon from "@mui/icons-material/AccountTree";
+import CheckBoxIcon from "@mui/icons-material/CheckBox";
+import IconButton from "@mui/material/IconButton";
+import renderer from "react-test-renderer";
+import { TreeGrid, TreeGridData, TreeGridId } from "treegrid";
+
+test("Table renders correctly", () => {
+  const columns = [
+    {
+      key: "col1",
+      width: "100%",
+      title: "Column 1",
+    },
+    {
+      key: "col2",
+      width: 150,
+      title: "Column 2",
+    },
+    {
+      key: "col3",
+      width: 50,
+      title: <AccountTreeIcon />,
+    },
+  ];
+
+  const data: TreeGridData = {
+    root: {
+      data: {},
+      children: [1, 3],
+    },
+    1: {
+      data: {
+        col1: "1-col1",
+        col2: "1-col2",
+      },
+      children: [2],
+    },
+    2: {
+      data: {
+        col1: "2-col1",
+        col2: "2-col2",
+        col3: "2-col3",
+      },
+    },
+    3: {
+      data: {
+        col1: "3-col1",
+        col2: "3-col2",
+        col3: <AccountTreeIcon />,
+      },
+    },
+  };
+
+  const prefixElements = (id: TreeGridId) => {
+    // Add variety to the rows.
+    if (id !== 2) {
+      return undefined;
+    }
+    return;
+    <IconButton size="small">
+      <CheckBoxIcon fontSize="inherit" />
+    </IconButton>;
+  };
+
+  const tree = renderer
+    .create(
+      <TreeGrid columns={columns} data={data} prefixElements={prefixElements} />
+    )
+    .toJSON();
+  expect(tree).toMatchSnapshot();
+});
--- a/ui/src/treegrid.tsx
+++ b/ui/src/treegrid.tsx
@@ -1,0 +1,257 @@
+import ErrorIcon from "@mui/icons-material/Error";
+import KeyboardArrowDownIcon from "@mui/icons-material/KeyboardArrowDown";
+import KeyboardArrowRightIcon from "@mui/icons-material/KeyboardArrowRight";
+import CircularProgress from "@mui/material/CircularProgress";
+import IconButton from "@mui/material/IconButton";
+import Typography from "@mui/material/Typography";
+import { ReactNode, useCallback } from "react";
+import { useImmer } from "use-immer";
+
+export type TreeGridId = string | number;
+
+export type TreeGridRowData = {
+  [key: string]: undefined | string | number | boolean | JSX.Element;
+};
+
+export type TreeGridItem = {
+  children?: TreeGridId[];
+  data: TreeGridRowData;
+};
+
+export type TreeGridData = {
+  [key: TreeGridId]: TreeGridItem;
+};
+
+export type TreeGridColumn = {
+  key: string;
+  width: string | number;
+  title?: string | JSX.Element;
+};
+
+export type TreeGridProps = {
+  columns: TreeGridColumn[];
+  data: TreeGridData;
+  prefixElements?: (id: TreeGridId, data: TreeGridRowData) => ReactNode;
+  // The returned string is displayed as an error message for failures and
+  // ignored for successes.
+  loadChildren?: (id: TreeGridId) => Promise<string>;
+};
+
+export function TreeGrid(props: TreeGridProps) {
+  const [state, updateState] = useImmer<TreeGridState>(
+    new Map<TreeGridId, TreeGridItemState>()
+  );
+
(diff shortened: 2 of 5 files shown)


cbookg (commented):
Can you post a screenshot so I can have some context?

tjennison-google:
Sure

image

ui/src/treegrid.tsx
```diff
@@ -0,0 +1,253 @@
...
+      const itemState = draft.get(id) || {
+        status: Status.Collapsed,
+      };
+      switch (itemState.status) {
+        case Status.Loading:
+          return;
+        case Status.Expanded:
+          itemState.status = Status.Collapsed;
+          break;
+        default:
```
cbookg, line 54: do you want this to be what is run in the 'error' state?

ui/src/treegrid.tsx
```diff
@@ -0,0 +1,253 @@
...
+        <thead>
+          <tr>
+            {props.columns.map((col, i) => (
+              <td
+                key={i}
+                style={{
+                  ...(col.width && {
+                    maxWidth: col.width,
+                    width: col.width,
+                    minWidth: col.width,
```
cbookg, line 99: seems wrong to be setting all 3 of these?

ui/src/treegrid.tsx
```diff
@@ -0,0 +1,253 @@
...
+                          title={childState?.errorMessage}
+                          onClick={() => {
+                            toggleExpanded(childId);
+                          }}
+                        >
+                          {itemIcon(childState)}
+                        </IconButton>
+                      )}
+                      {props.prefixElements &&
+                        props.prefixElements(childId, child.data)}
```
cbookg, line 202: i think just props.prefixElement?.(childId, child.data)

ui/src/treegrid.tsx
```diff
@@ -0,0 +1,253 @@
...
+export type TreeGridColumnData = {
+  key: string;
+  width: string | number;
+  title?: string | JSX.Element;
+};
+
+export type TreeGridProps = {
+  columns: TreeGridColumnData[];
+  data: TreeGridData;
+  prefixElements?: (id: TreeGridId, data: TreeGridRowData) => JSX.Element[];
```
cbookg, line 34: shouldn't this just return ReactNode and then you can return any valid JSX?  as an example, its weird that your test has to return an array

ui/src/treegrid.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+        col3: <AccountTreeIcon />,
+      },
+    },
+  };
+
+  const prefixElements = (id: TreeGridId) => {
+    if (id !== 2) {
+      return [];
+    }
+    return [
```
cbookg, line 58: i would expect this function to just be able to return any JSX:
return (
      <IconButton key={0} size="small">
        <CheckBoxIcon fontSize="inherit" />
      </IconButton>
)

ui/src/treegrid.tsx
tjennison-google: Done

ui/src/treegrid.tsx
```diff
@@ -0,0 +1,253 @@
...
+          <tbody>
+            {renderChildren(props, state, toggleExpanded, "root", 0, false)}
+          </tbody>
+        </table>
+      </div>
+    </>
+  );
+}
+
+function renderChildren(
```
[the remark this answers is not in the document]
tjennison-google: That's the design I started with but it mostly just shuffled things around. Since the root node isn't rendered, the code to render children needs to be shared between the tree and potential child components anyway, and the actual rendering is pretty contained inside the results.push call.

ui/src/treegrid.tsx
```diff
@@ -0,0 +1,247 @@
...
+import CircularProgress from "@mui/material/CircularProgress";
+import IconButton from "@mui/material/IconButton";
+import Typography from "@mui/material/Typography";
+import { ReactNode, useCallback } from "react";
+import { useImmer } from "use-immer";
+
+export type TreeGridId = string | number;
+
+export type TreeGridRowData = {
+  [key: string]: string | number | JSX.Element;
```
calbach, line 13: Another approach you could consider is to move an optional rendering function into the Column definition. This would require less data prep by the client to pass in the new values for `rows`, as it doesn't need to worry about display concerns.

ui/src/treegrid.tsx
```diff
@@ -0,0 +1,247 @@
...
+};
+
+export type TreeGridItem = {
+  children?: TreeGridId[];
+  data: TreeGridRowData;
+};
+
+export type TreeGridData = {
+  [key: TreeGridId]: TreeGridItem;
```
calbach, line 22: The caller cannot reliably impose an order on the rows in this model (except maybe with certain browsers and a deep understanding of object key iteration order). Is that OK? Something like `Map` would preserve insertion order.

ui/src/treegrid.tsx
```diff
@@ -0,0 +1,247 @@
...
+      <tr
+        key={childId}
+        style={{
+          ...(collapse && { visibility: "collapse" }),
+        }}
+      >
+        {props.columns.map((col, i) => {
+          let value = child.data[col.key] || "";
+          let title = "";
+          if (!(value instanceof Object)) {
```
calbach, line 151: Can you add a comment here? Is this attempting to check for JSX elements?

ui/src/treegrid.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+type TreeGridItemState = {
+  status: Status;
+  loaded?: boolean;
+  errorMessage?: string;
+};
+
+type TreeGridState = Map<TreeGridId, TreeGridItemState>;
+
+function itemIcon(state?: TreeGridItemState): JSX.Element {
```
calbach, line 237: nit: is there a reason not to make this a normal react function component?

ui/src/treegrid.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+export type TreeGridProps = {
+  columns: TreeGridColumnData[];
+  data: TreeGridData;
+  prefixElements?: (id: TreeGridId, data: TreeGridRowData) => ReactNode;
+  loadChildren?: (id: TreeGridId) => Promise<string>;
+};
+
+export function TreeGrid(props: TreeGridProps) {
+  const [state, updateState] = useImmer<TreeGridState>(
```
calbach, line 39: I'm not familiar with immer, though seems reasonable at a glance - how did you land on this? Any alternatives considered?

ui/src/treegrid.tsx
```diff
@@ -0,0 +1,247 @@
...
+                    status: Status.Failed,
+                    errorMessage: error,
+                  });
+                });
+              });
+          }
+      }
+      draft.set(id, itemState);
+    });
+  }, []);
```
calbach, line 81: Should this be invalidated by `props.loadChildren` changing?

calbach (commented):
Nice, very clean / minimal

ui/src/treegrid.tsx
tjennison-google: I was unsatisfied with the lodash deepCopy method I had been using so far. I didn't find any other simple options beyond using basic javascript (e.g. spreading) which I've found unwieldy with nontrivial state. I'm also considering whether it makes sense to buy into a larger ecosystem like Redux or Recoil. Conveniently, I'm leaning towards Redux which uses immer so this works well both independently and also if I go that route. The other options for this used by Recoil and other systems generally aren't separable from their systems.

ui/src/treegrid.tsx
tjennison-google: Done.

ui/src/treegrid.tsx
tjennison-google: Yeah, done.

ui/src/treegrid.tsx
tjennison-google: Done. I had just thought of it as a helper but it was basically a component already.

calbach (approved):

PR #90, opened 2022-01-19
Convert concept view to use the new TreeGrid

tjennison-google:


ui/src/criteria/concept.tsx
```diff
@@ -66,25 +69,21 @@ export class ConceptCriteria extends Criteria {
...
+  { key: "concept_id", width: 120, title: "Concept ID" },
+  { key: "domain_id", width: 100, title: "Domain" },
+  { key: "standard_concept", width: 180, title: "Source/Standard" },
+  { key: "vocabulary_id", width: 120, title: "Vocab" },
+  { key: "concept_code", width: 120, title: "Code" },
+];
+
+const allColumns: TreeGridColumn[] = [
+  ...fetchedColumns,
+  { key: "view_hierarchy", width: 160, title: "View Hierarchy" },
```
calbach, line 86: SHould this key be special, e.g.  an exported constant from TreeGrid? If it's not a special value, how is the TreeGrid recognizing where to put the heirarchy expansion handler?

ui/src/criteria/concept.tsx
```diff
@@ -121,101 +120,89 @@ function ConceptEdit(props: ConceptEditProps) {
...
+                    row[k] = v.boolVal;
+                  } else {
+                    row[k] = v.stringVal;
                   }
-                  return row;
-                })
-                .filter((row) => row.id !== 0);
+                }
+
+                row["standard_concept"] = !row["standard_concept"]
```
calbach, line 153: nit: stared at this for a while before correctly parsing; I think just `row["standard_concept"]` as the predicate is probably a bit clearer. I also think moving it into the above for loop might be slightly clearer to avoid the in-place modification.

ui/src/criteria/concept.tsx
```diff
@@ -121,101 +120,89 @@ function ConceptEdit(props: ConceptEditProps) {
...
+                  ? "Source"
+                  : "Standard";
+
+                children.push(id);
+                data[id] = { data: row };
+              });
             }
-            return [];
+            data.root = { children: children, data: {} };
```
calbach, line 161:
```suggestion
            data.root = { children, data: {} };
```

calbach (approved):

ui/src/criteria/concept.tsx
tjennison-google: The TreeGrid doesn't know anything about it. TreeGridData accepts a JSX.Element instead of string. It's filled in on line 134, though without a handler attached.

Something along those lines came up in one of your comments on the previous commit and I didn't think to mention this. There's overlap with prefixElements but this replaces the whole cell which would require handling the layout and formatting of the text. I think this is a good balance for what we need now but we'll see how it evolves.

ui/src/criteria/concept.tsx
tjennison-google: Done.

ui/src/criteria/concept.tsx
tjennison-google: Done.

PR #92, opened 2022-01-25
Add new ArrayColumnFilter type of TableFilter.

marikomedlock:
- Added a new type of `TableFilter`. Previously, we could only filter on a single column with a `BinaryColumnFilter`. Now we can compose multiple sub-filters into a new filter that filters on >1 column with an `ArrayColumnFilter`.
- Tests use the nautical underlay.

The purpose of this change is to allow us to define entities as filtered versions of a table, where the filter includes >1 condition. For example, standard conditions may be rows of the `concept` table where `domain_id=Condition AND standard_concept IS NOT NULL`.

### DIFF
--- a/api/src/main/java/bio/terra/tanagra/service/search/SqlVisitor.java
+++ b/api/src/main/java/bio/terra/tanagra/service/search/SqlVisitor.java
@@ -296,12 +296,14 @@ public class SqlVisitor {
       return String.format("%s AS %s", resolvedTable, entityVariable.variable().name());
     }
 
+    /** Resolve a {@link Table} into a SQL string. */
     private String resolveTable(Table table) {
       // `projectId.datasetId`.table
       return String.format(
           "`%s.%s`.%s", table.dataset().projectId(), table.dataset().datasetId(), table.name());
     }
 
+    /** Resolve a {@link Table} with a {@link TableFilter} into a SQL string. */
     private String resolveTable(Table table, @Nullable TableFilter tableFilter) {
       if (tableFilter == null) {
         return resolveTable(table);
@@ -315,6 +317,12 @@ public class SqlVisitor {
       }
     }
 
+    /**
+     * Resolve a {@link Table} with an {@link ArrayColumnFilter} into a SQL string.
+     *
+     * @param whereClauseOnly If true, only returns the part of the SQL string after the WHERE. This
+     *     is helpful when building an {@link ArrayColumnFilter} composed of multiple sub-filters.
+     */
     private String resolveArrayColumnFilter(
         Table table, ArrayColumnFilter arrayColumnFilter, boolean whereClauseOnly) {
       String joinOperatorInWhereClause;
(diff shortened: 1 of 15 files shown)


api/src/main/java/bio/terra/tanagra/service/search/SqlVisitor.java
```diff
@@ -294,19 +296,89 @@ public String resolveTable(EntityVariable entityVariable) {
...
+          break;
+        default:
+          throw new IllegalArgumentException(
+              "Unknown array column filter operator type: " + arrayColumnFilter.operator());
+      }
+
+      // %s AND %s AND (%s OR %s) ...
+      StringBuilder joinedWhereClauses = new StringBuilder();
+      int subFiltersCtr = 0;
+      for (BinaryColumnFilter subFilter : arrayColumnFilter.binaryColumnFilters()) {
```
calbach, line 344: opt nit: streams + joiner could avoid the need for the `ctr` variable

```
Streams.concat(
  arrayColumnFilter.binaryColumnFilters().stream().map(f -> resolveBinaryColumnFilter(table, f, true)),
  arrayColumnFilter.arrayColumnFilters().stream().map(f -> resolveArrayColumnFilters(table, f, true))
).collect(Collectors.joining(" " + joinOperatorInWhereClause + " "))
```

api/src/main/java/bio/terra/tanagra/service/search/SqlVisitor.java
```diff
@@ -294,19 +296,89 @@ public String resolveTable(EntityVariable entityVariable) {
...
+      if (whereClauseOnly) {
+        return "(" + joinedWhereClauses + ")";
+      } else {
+        // (SELECT * FROM `projectId.datasetId`.table WHERE joinedWhereClauses)
+        return String.format(
+            "(SELECT * FROM `%s.%s`.%s WHERE %s)",
+            table.dataset().projectId(),
+            table.dataset().datasetId(),
+            table.name(),
+            joinedWhereClauses);
```
calbach, line 368: nit: maybe avoid the `whereClauseOnly` boolean, make this method only worry about the WHERE clause, and extract the full query logic out to the top level. This would also avoid this duplication across these two methods (and future filter methods)

calbach (approved):
Looks good, a few optional style suggestions

api/src/main/java/bio/terra/tanagra/service/search/SqlVisitor.java
marikomedlock: Good tip, thanks! Modified slightly to avoid appending the join operator at the beginning.

api/src/main/java/bio/terra/tanagra/service/search/SqlVisitor.java
marikomedlock: Good idea. Done.

api/src/main/java/bio/terra/tanagra/service/search/SqlVisitor.java
calbach: Hmm perhaps I'm missing something.. a joiner only places the separator between elements so the first element shouldn't need any special handling.

marikomedlock: Oh you're totally right. I just realized that when I tried this I wasn't updating the original `StringBuilder` variable, so I need to just skip that the return the result of this `collect()`. Thanks for the correction! I'll add this change to one of the existing open PRs.

PR #95, opened 2022-01-28
Migrate state management to Redux

tjennison-google:
Management of cohort state has been migrated out of UserData and context
and into the Redux store and reducers. The current set of actions
mirrors what was there before but it would be good to break
updateCriteria into several actions for more precise edits, especially
for undo/redo support.

One side effect is that Redux supports only basic objects and arrays so
the existing classes have had their functionality moved into separate
functions. In general, this isn't a big deal but it it prevents us from
continuing to use inheritance to support multiple criteria types. I have
instead implemented a basic plugin system which allows multiple types of
criteria without pulling the spcifics into the data model, and also
provides a prototype for what a more robust plugin system might look
like long term. The plugin system will likely evolve over time but for
now it demonstrates how criteria plugins can be defined separately and
referenced only in a single, easily merged location.

While the TypeScript decorators are still experimental, they have
existed for a couple years and are unlikely to go away. At some point
they may be deprecated in favor of the latest JavaScript decorator
proposal, https://github.com/tc39/proposal-decorators, at which point
would could easily migrate if necessary.


ui/src/criteria/concept.tsx
```diff
@@ -163,39 +177,40 @@ function ConceptEdit(props: ConceptEditProps) {
...
-                          name: !!name ? String(name) : "",
-                        });
-                      }
-                    }
+                  dispatch(
+                    updateCriteriaData({
+                      cohortId: props.cohort.id,
+                      groupId: props.group.id,
+                      criteriaId: props.criteriaId,
+                      data: produce(props.data, (data) => {
```
calbach, line 202: I'm not quite able to follow what's happening here, mainly due to a lack of redux experience. Can you explain why immer's produce is necessary here?

calbach (commented):
Sorry, my review is quite slow as I haven't used redux extensively before. Am slowly ramping up but will likely have more questions/comments

ui/src/criteria/concept.tsx
```diff
@@ -23,32 +25,46 @@ type Selection = {
...
 
-export class ConceptCriteria extends Criteria {
-  constructor(name: string, public filter: string) {
-    super(name);
+type Data = {
+  filter: string;
+  selected: Selection[];
+};
+
+@registerCriteriaPlugin("condition", "Condition", "Contains Conditions Codes")
```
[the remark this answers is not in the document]
tjennison-google: This will be necessary long term. Once the entity model and UI hints are fully realized, this will become more generic and some or all of the concept domains will be handled by a single plugin but there will be at least one more for demographics in the core repo. Further down the road I expect additional plugins for things like documents, genomes, and FitBit data. Beyond even those, this is a key point of extensibility where users will need to integrate custom criteria, particularly if they integrate with external data sources.

I also think there's potential here to have partial plugins or extensions where users want to customize the UI in some fashion over standard data but I haven't thought about that too much yet.

ui/src/cohort.ts
```diff
@@ -1,136 +1,149 @@
...
+  return {
+    arrayFilter: {
+      operands: operands,
+      operator: tanagra.ArrayFilterOperator.Or,
+    },
+  };
+}
 
+// Since Redux doesn't allow classes in its data model, criteria specific data
+// is stored in an object that the criteria knows the contents of but is
```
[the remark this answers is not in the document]
tjennison-google: The data field. I've updated the comment to hopefully make that clearer. It addresses the Redux constraint by storing only the data in the model as plain objects and keeping the functionality in the plugins which are created as needed.

ui/src/criteria/concept.tsx
tjennison-google: I can't change the existing data, just like with React state. So I would otherwise have to deep copy data or use spreads to get a new object. immer handles that automatically and allows more natural code using methods like splice.

tjennison-google: The end game here is to stop using the generic updateData method and instead allow the plugins to implement their own actions and reducer. Then this would change to two actions: select and unselect, and the parameters would just be the ids so there would be no need to copy the data here. That will also be much nicer for undo/redo messages. This is already a large change though so I tried to keep things the same where possible.

calbach: I see, thanks. Sorry - I missed that this was modifying `props.data`, and this was not a normal redux update (which from reading the docs - it seems already uses immer behind the scenes).

calbach (approved):
Thanks, looks good. From what I learned about Redux, this seems to be following the usual conventions - but I'm not able to give very sophisticated feedback on the approach at this point (if we have any local experts who can do a guest review, that would be neat - I don't know of any).

tjennison-google:
I unfortunately don't know any Redux experts either so we'll be learning as we go.

PR #96, opened 2022-01-28
Add workflow for generating a node-path index table.

marikomedlock:
- Added a new Apache Beam workflow for generating a (node,path) index table for hierarchies. The path is a list of ancestors in order for the given node. e.g. `(123,"456.789")` where 456 is the parent of 123 and 789 is the grandparent of 123. `(111,null)` where 111 is a root node.
- Generated 2 new node-path index tables for all rows in the `concept` table in the `aou_synthetic` and `synpuf` datasets.
- I originally wrote a simpler workflow to recursively call BQ to get the next node in the path for each row (see initial commit). This worked for small numbers of rows, but hit BQ quota limits when running against all rows in the `concept` table for a given `domain_id`. The current version of the workflow reads in all parent-child relationships and does this recursion in Beam without calling out to BQ again.
- Currently the `path` is a `nullable string` field in the output BQ table because we don't support the array data type in Tanagra, yet. Once we do, we can change this field to be a `repeated integer`.


tjennison-google (approved):

calbach:
Sorry for the slow turnaround, I will try to get to this later today.

workflow/src/main/java/bio/terra/tanagra/workflow/BuildPathsForHierarchy.java
```diff
@@ -0,0 +1,325 @@
...
+                  new DoFn<KV<Long, CoGbkResult>, KV<Long, String>>() {
+                    @ProcessElement
+                    public void processElement(ProcessContext context) {
+                      KV<Long, CoGbkResult> element = context.element();
+                      Long node = element.getKey();
+                      Iterator<String> pathTagIter = element.getValue().getAll(pathTag).iterator();
+                      Iterator<String> parentTagIter =
+                          element.getValue().getAll(parentTag).iterator();
+
+                      // there may be multiple possible next steps (i.e. the current node may have
```
calbach, line 188: Some of the nodes in the standard hierarchy are polyhierarchical, e.g. conditions. Have you covered this with Brian/ Robert already? I think the currently designed model doesn't really handle this scenario - which maybe is reasonable..

workflow/src/main/java/bio/terra/tanagra/workflow/BuildPathsForHierarchy.java
```diff
@@ -0,0 +1,325 @@
...
+    //    [iteration 1] nodeParentPathJoin (next node,paths,next node parents)
+    //    (a1,[],["a2"])
+    //    (a2,["a2.a1"],["a3"])
+    //    (a3,["a3.a2","a3"],[])
+    //    (b1,[],["b2]")
+    //    (b2,["b2.b1","b2"],[])
+    //    (c1,["c1"],[])
+
+    // swap the key of all pairs from the next node in the path (which should all be root nodes at
+    // this point), to the first node in the path
```
calbach, line 258: nit: and trim the target node from the path?

workflow/src/main/java/bio/terra/tanagra/workflow/BuildPathsForHierarchy.java
```diff
@@ -0,0 +1,325 @@
...
+
+                    String path = kvPair.getValue();
+
+                    // strip out the first node in the path, and make it the key
+                    // e.g. (a3,"a1.a2.a3") => (a1,"a2.a3")
+                    //      (c1,"c1") => (c1,"")
+                    int indexOfFirstPeriod = path.indexOf(".");
+                    String firstNodeInPath;
+                    String pathWithoutFirstNode;
+                    if (indexOfFirstPeriod > 0) {
```
calbach, line 275: opt nit: I'd probably be tempted to split by ".", pop the first element, then join them back together here. A little less dependent on the encoding details

workflow/src/main/java/bio/terra/tanagra/workflow/BuildPathsForHierarchy.java
```diff
@@ -0,0 +1,325 @@
...
+                    } else {
+                      firstNodeInPath = path;
+                      pathWithoutFirstNode = "";
+                    }
+
+                    Long firstNode;
+                    try {
+                      firstNode = Long.valueOf(firstNodeInPath);
+                    } catch (NumberFormatException nfEx) {
+                      firstNode = 0L;
```
calbach, line 287: Is this OK? Expected? Did you observe this happening?

calbach (approved):
This looks reasonable, thanks. I have limited Beam / Flume experience, but I think the pipeline here looks reasonable and is probably similar to what I would have done.

I assumed that you've been referencing Will's previous work on Beam. **However** I forgot that some of the work he added was in this library: https://github.com/DataBiosphere/tanagra/blob/main/workflow/src/main/java/bio/terra/tanagra/workflow/GraphUtils.java  - hopefully you found this (see also, the collocated tests). Using a similar algorithm the one used for the transitive closure helper  may perform better than this solution - but I don't think it can be reused directly  and I don't think it's significant enough to justify reworking it. The pattern of pulling out the main pieces into a helper library is definitely a nice pattern for testability.

workflow/src/main/java/bio/terra/tanagra/workflow/BuildPathsForHierarchy.java
```diff
@@ -0,0 +1,325 @@
...
+  /** Defines the BigQuery schema for the output hierarchy table. */
+  private static TableSchema pathsForHierarchySchema(BuildPathsForHierarchyOptions options) {
+    return new TableSchema()
+        .setFields(
+            List.of(
+                new TableFieldSchema()
+                    .setName(options.getOutputNodeColumn())
+                    .setType("INTEGER")
+                    .setMode("REQUIRED"),
+                // TODO: make the path column an INTEGER/REPEATED field instead of a string so we
```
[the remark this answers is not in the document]
marikomedlock: Another possibility is to serialize a list into a string. Expanded on the comment to include your note about BQ repeated fields.

workflow/src/main/java/bio/terra/tanagra/workflow/BuildPathsForHierarchy.java
marikomedlock: We haven't discussed it much so far, but I do know they exist. I think we do currently support this, in the sense that:
- All parent-child relationships are included in one of the hierarchy index tables, so if a node has >1 parent, it should show up in a `CHILD_OF` query for both parents.
- The node-path table does only specify one path for each node, but I think that's what we need for Tanagra because even if a node can show up in >1 place in a hierarchy tree, we need to highlight exactly one of those places when someone clicks the hierarchy icon.

I am not sure whether the ancestor-descendant hierarchy index table that is generated accounts for nodes with >1 parent.

workflow/src/main/java/bio/terra/tanagra/workflow/BuildPathsForHierarchy.java
marikomedlock: Right, added.

workflow/src/main/java/bio/terra/tanagra/workflow/BuildPathsForHierarchy.java
marikomedlock: This should never happen because the underlying BQ columns (`concept.concept_id`, `concept_relationship.concept_id_1`, `concept_relationship.concept_id_2`) are all of type `INTEGER REQUIRED`. I added this for debugging only. Removed now.

marikomedlock:
> I assumed that you've been referencing Will's previous work on Beam. **However** I forgot that some of the work he added was in this library: https://github.com/DataBiosphere/tanagra/blob/main/workflow/src/main/java/bio/terra/tanagra/workflow/GraphUtils.java - hopefully you found this (see also, the collocated tests). Using a similar algorithm to the one used for the transitive closure helper may perform better than this solution - but I don't think it can be reused directly and I don't think it's significant enough to justify reworking it. The pattern of pulling out the main pieces into a helper library is definitely a nice pattern for testability.

Yes, I spent a while going through Will's existing Beam pipeline before writing this one. I've never used Beam or Flume before, so I started with his and modified as I studied Beam tutorials. I also saw the `GraphUtils` and tests, but couldn't immediately see how to modify it to generate this path instead. I think it's possible, and I'm sure there's a better algorithm for generating these paths, but I wanted to prioritize getting something working over performance for now.

Agreed that pulling some pieces of this out into a helper library would help test this. I'll try to tackle that in a follow-on PR.

PR #97, opened 2022-01-31
Add a PathsTable to the underlay hierarchy definition.

marikomedlock:
- Added a `PathsTable` to the definition of a hierarchy in the underlay. This specifies a table with two columns `(node,path)` that contains the hierarchy path for each of the nodes in the hierarchy.
- When there is a hierarchy attached to an entity attribute, we now add a new (generated) path attribute to the same entity. This path attribute's name is `t_path_[hierarchical attribute name]` and can be included in the list of fields to return in a query like other normal attributes.
- Added tests that use the nautical underlay and `aou_synthetic` underlay.

From spot-testing with the `aou_synthetic` underlay, the condition and procedure entity paths look mostly correct, but the ingredient and measurement paths are mostly empty. This is likely due to a bad definition of those entities or relationships. In particular, I think we need to generate the ingredient hierarchy paths using the `'ATC - RxNorm'`,`'ATC - RxNorm pr up'`,`RxNorm - ATC sec up'` relationships, instead of `'Subsumes'` ([ref](https://www.ohdsi.org/web/wiki/doku.php?id=documentation:vocabulary:atc)). And for measurements, the hierarchy appears to cross domains (measurements and observations), so I'm not sure whether we need to expand the definition of a measurement entity or define a new entity that crosses domains and holds the hierarchy. These are questions I've got for upcoming discussions with Brian Freeman, so follow-on PRs should address this.


tjennison-google (approved):

api/src/main/java/bio/terra/tanagra/service/underlay/UnderlayConversion.java
```diff
@@ -406,16 +408,47 @@ public static Underlay convert(bio.terra.tanagra.proto.underlay.Underlay underla
...
       Hierarchy hierarchy =
           Hierarchy.builder()
               .descendantsTable(descendantsTable.build())
               .childrenTable(childrenTable.build())
+              .pathsTable(pathsTable.build())
               .build();
+      // add a new attribute+mapping to the entity that contains the hierarchy path
+      Attribute hierarchyPathAttribute =
+          Attribute.builder()
+              .name("t_path_" + attribute.name())
```
calbach, line 427: What does "t" in "t_path" stand for?

lib/src/main/proto/bio/terra/tanagra/underlay/hierarchy.proto
```diff
@@ -17,6 +17,8 @@ message Hierarchy {
   DescendantsTable descendants_table = 2;
 
   ChildrenTable children_table = 3;
+
+  PathsTable paths_table = 4;
```
calbach, line 21: Should probably comment here that inclusion of this will result in additional generated entity attributes, and the naming convention

api/src/main/java/bio/terra/tanagra/service/underlay/UnderlayConversion.java
calbach, line 427: nit: should we validate user-specified attribute names in general against reserved Tanagra generated prefixes? e.g. I imagine things crash and burn if our condition entity already had an attribute with this name, however unlikely.

calbach (approved):
I think the general pattern of generating an attribute on the entity seems OK. This may slightly expand the scope later when we get to export of the dataset - specifically because we can now potentially select a generated attribute sourced by an index table (may have alternate perimissioning from the primary dataset), rather than just selecting a cohort based on an index. This is just a small extension of a problem we already needed to solve though, so it may not have much of a meaningful impact.

api/src/main/java/bio/terra/tanagra/service/underlay/UnderlayConversion.java
marikomedlock: Tanagra. I just wanted to namespace generated attributes and picked a prefix that's relatively short since there's a 31 character limit on the attribute name. I don't know of a hard reason for that limit, so we could also increase it and use a longer prefix like "tanagra_".

lib/src/main/proto/bio/terra/tanagra/underlay/hierarchy.proto
marikomedlock: Added.

api/src/main/java/bio/terra/tanagra/service/underlay/UnderlayConversion.java
marikomedlock: Yes, good point. I added an `isGenerated` property to the internal `Attribute` class, so we can validate that the name doesn't include the prefix for user-defined, not-generated properties. Per your other comment, I'm thinking we can also use this property in the future to determine whether an attribute should be selectable by a user in dataset builder.

PR #99, opened 2022-02-02
Support NULL value and add a NOT_EQUALS operator for binary filters in the API.

marikomedlock:
- Added support for a `NULL` value in an API binary filter.
- Added a `NOT_EQUALS` operator for API binary filters.
- Wrote tests for these changes against the demo nautical underlay and the `aou_synthetic` underlay.

The immediate purpose of this change is to allow querying for root nodes, which have the generated `t_path_...` attribute set to `NULL`. A future PR may alter the query for fetching root nodes (i.e. to `t_path_...` = empty string), but both operators will still be useful for other queries against nullable attributes.

Note that the get all root nodes queries for the hierarchical OMOP entities (condition, procedure, measurement, ingredient) currently return a lot of rows. This will hopefully be reduced in a follow-up PR that implements some of the suggestions from Brian Freeman and Robert Carroll (i.e. restricting vocabularies, removing some lone root nodes with no children from the hierarchy, using different relationships for drugs).


tjennison-google (approved):

api/src/test/java/bio/terra/tanagra/aousynthetic/MeasurementEntityQueriesTest.java
```diff
@@ -143,4 +143,37 @@ void generateSqlForHierarchyPathOfAMeasurementEntity() throws IOException {
...
+  @DisplayName(
+      "correct SQL string for listing all measurement entity instances that are root nodes in the hierarchy")
+  void generateSqlForAllRootNodeMeasurementEntities() throws IOException {
+    // filter for "measurement" entity instances that have t_path_concept_id IS NULL
+    // i.e. measurement root nodes
+    ApiFilter isRootNode =
+        new ApiFilter()
+            .binaryFilter(
+                new ApiBinaryFilter()
+                    // not setting AttributeValue means to use a null value
```
calbach, line 157: Existing issue, but looking at this, I'm wondering if a more explicit NULL signal in AttributeValue would be cleaner. The current encoding feels a bit hard to distinguish from a malformed request.

api/src/main/java/bio/terra/tanagra/service/query/api/ExpressionConverter.java
```diff
@@ -43,6 +43,10 @@
   }
 
   public Expression.Literal convert(ApiAttributeValue apiAttributeValue) {
+    if (apiAttributeValue == null) {
+      return Expression.Literal.create(DataType.STRING, null);
```
calbach, line 47: Why STRING? Is this the default? maybe add a comment here

calbach (approved):

api/src/test/java/bio/terra/tanagra/aousynthetic/MeasurementEntityQueriesTest.java
marikomedlock: Agreed. I'll plan to tackle that in a future PR.

api/src/main/java/bio/terra/tanagra/service/query/api/ExpressionConverter.java
marikomedlock: The `DataType` here doesn't matter because when doing the special-case `null` handling, we just check whether the value is `null` and ignore the `DataType`. Added a comment.

PR #102, opened 2022-02-08
Add SearchTextFilter to the search entity instances API endpoints.

marikomedlock:
- Added a new section to the underlay `text_search_information`, which maps an entity to a table with information needed to do a text search for entity instances.
  - Wrote tests against the nautical underlay.
- Added a new `ApiTextSearchFilter` that clients can specify to filter entity instances on a text string. This is different than using an `ApiBinaryFilter` on a string attribute, because the text search may filter on multiple attribute values (e.g. first name and last name) and non-attribute values (e.g. synonyms) all in the same search. This will also give us the flexibility to implement more sophisticated text search (e.g. fuzzy matching) in the future.
  - Added a corresponding new internal class `Filter.TextSearchFilter` that converts from the `ApiTextSearchFilter`.
  - Added a `visitTextSearch` method to `SqlVisitor` to turn the `Filter.TextSearchFilter` into a SQL string.
  - Wrote a test in `FilterConverterTest` to check the api -> internal object conversion.
  - Wrote a test in `SqlVisitorTest` to check the generated SQL string when called from the internal object level.
  - Wrote a test in `EntitiesFilterApiControllerTest` to check the generated SQL string when called from the API level. This test also provides an example of how to do this kind of search from the client.
  
There were two existing fields in the API definition: `SearchTermFilter` as a type of `ApiFilter` and `searchTermSupported` as a property of an `Entity`. Neither appeared to be implemented on the backend or used by any tests. I removed the `searchTermSupported` property, renamed `SearchTermFilter` -> `TextSearchFilter` and implemented it on the backend.


lib/src/main/proto/bio/terra/tanagra/underlay/underlay.proto
```diff
@@ -35,6 +36,9 @@ message Underlay {
   // The hierarchies for attributes with hierarchies in the underlay.
   repeated Hierarchy hierarchies = 9;
 
+  // The text search information for entities in the underlay.
+  repeated Text texts = 10;
```
calbach, line 40: Is it correct to have this at the top level? Could it belong under `EntityFiltersSchema` instead?

If this level is correct, possibly it is worth renaming; from a quick scan it's not clear whether this pertains to:
1. the source dataset
2. the entity model
3. or, the mapping

I don't have great suggestions yet as I'm not sure I fully grok the scope

lib/src/main/proto/bio/terra/tanagra/underlay/text.proto
```diff
@@ -0,0 +1,27 @@
...
+
+  TextTable text_table = 2;
+}
+
+// The table of information required to perform the text search.
+// There should be a row (lookupTableKey, fullText) for each possible
+// text string to match against per entity.
+message TextTable {
+  // A column that foreign keys to the entity primary key.
+  ColumnId lookup_table_key = 1;
```
calbach, line 24: I don't quite follow why this layer of misdirection is necessary. It seems like this is a design decision around responsibility of the client to construct a query vs how much semantic knowledge is being encoded into the underlay.

Do you have a concrete example to help illustrate this? i.e. showing the tradeoff between
1. the client constructing a filter on their target entity which includes a relationship filter + text match filter, vs
2. the client constructing a filter on their target entity with a text match (underlay encodes the relationship)

Is this because we don't want this relationship  to exist in the entity model?

lib/src/main/proto/bio/terra/tanagra/underlay/underlay.proto
marikomedlock: I would say this pertains to the entity model, similar to how hierarchies pertain to the entity model. i.e. "this is the information you need to do text search/hierarchical queries".

I could have put it under `EntityFiltersSchema`, since that also holds a reference to the entity. But my understanding of the intent behind `EntityFiltersSchema` is that it would hold information for the UI about how to construct filters when making API requests. This is based on initial handoff conversations with Will and the example in the test `nautical` [underlay](https://github.com/DataBiosphere/tanagra/blob/08a8740352c8e142cd8b58a5293e8e595895cc01/api/src/test/resources/underlays/nautical.yaml#L408). So this information about where to find information required to construct text search queries doesn't really seem to fit there.

I'm certainly open to renaming to make the purpose of this clearer -- maybe `TextSearchInformation`?

lib/src/main/proto/bio/terra/tanagra/underlay/text.proto
marikomedlock: Right now (i.e. with all 3 of these related PRs together, including the workflow one), doing a text search for an entity means doing an exact substring match on a single column in a table that has one row per entity. We could accomplish the same thing by defining a new entity attribute that points to this column. However, I think that approach will not be flexible enough for future changes. e.g.:
- Allowing multiple rows per entity instance in this text search info table (e.g. one row for each synonym or language).
- Allowing multiple columns per entity instance in this text search info table (e.g. text string to search, a relevance ranking).

I think you're right about this being a design choice between how much information the client needs to know vs how much information we encode into the underlay. With this `Text` section of the underlay, the client doesn't need to know how text search is implemented -- they just say "do a text search on entity A". A more client-heavy approach could have been to let the client specify a list of entity attributes to perform the text search on (e.g. text search on name, id and concept_code). That would be a larger burden on the client and also force us to expose all text searchable strings as entity attributes, which may not make much sense (e.g. when we have a list of synonyms, it's not really 1-to-1). I think that could also become a problem once we get to performance tuning because it will be harder to build a good index when the client can specify a variable number of fields to text search on.

RE relationship filter: Relationships require two entities. For text search, it's not clear to me what the second entity would be. To do a text search for conditions, we have the condition entity already, but would need to define a text-search-condition entity? I generally want to think of entities as "objects the UI (user) cares about". So the user can explain what a condition is, but probably not a text-search-condition.

I'm not sure I've answered your questions here, particularly the part about relationships I may be off base to what you were thinking. Please rephrase/ask again if not, and maybe this will be good to talk about in person at stand-up today or in a separate time block if needed.

tjennison-google (approved):

lib/src/main/proto/bio/terra/tanagra/underlay/text.proto
calbach: Yes, you answered my question - thanks. I think this makes sense to me as a guiding principle. Follow-up question based on your examples:

Is the cardinality correct here? Is it reasonable to have an "at most one" relationship between an entity, and text search configuration? For example, is it possible that we'd want to support text search on two different logical attributes (or multiple search configurations on the same attribute)? Or if tying the search to a specific attribute is too limiting (as you mention above, maybe you want to search multiple attributes with one operation), should it still be possible to have multiple distinct text search types on the same entity?

This could definitely be a future extension, if there's no foreseeable need for it

lib/src/main/proto/bio/terra/tanagra/underlay/underlay.proto
calbach: I can't come up with names that I love that much better than what you have here, especially pluralized. 

Some that don't quite fit (I don't think):
TextIndex
TextEngine

Fine with your suggestion of TextSearchInformation (I suppose that's already plural?) or keeping what you had.

Maybe just some restructuring of the underlay proto into related sections / groupings would address my concern. Down the road, also we could consider having a top-level SearchSupport section; but I'm not going to push for that as it would cause a mass indentation as currently structured (maybe more viable later if we support splitting underlay sections into multiple files)

calbach (approved):

lib/src/main/proto/bio/terra/tanagra/underlay/underlay.proto
marikomedlock: Agreed it would be good to do some restructuring of the underlay. In particular, breaking it into at least 2 separate groups of files would make sense to me: entity model, mapping of underlying dataset(s) to entity model. This might be good to do in preparation for driving the indexing workflows directly from the underlay definition. Regardless, I'll take your suggestion to do this as follow-on work instead of in this PR.

lib/src/main/proto/bio/terra/tanagra/underlay/text.proto
marikomedlock: RE cardinality: You're right we may want multiple configurations of text search in the future. I think the exact use case will influence how we decide to implement it.

e.g. fuzzy text search & exact match text search
For this, I'd expect we could reuse the same index table and just use a different algorithm (e.g. bigquery function) to do the query. So we could continue to restrict it to 1 `Text` block in the underlay per entity.

e.g. english text search & japanese text search
For this, I'd expect we'd have different index tables and would need two pointers to dataset tables per entity. So we'd need to allow >1 `Text` block in the underlay per entity.

As you say, I expect we'll need to revisit this in the future, but I don't see an immediate use case for >1 text search index table per entity.

lib/src/main/proto/bio/terra/tanagra/underlay/underlay.proto
marikomedlock: Renamed Text -> TextSearchInformation.

PR #104, opened 2022-02-08
Add workflow for generating a text search index table.

marikomedlock:
- Added a new Beam workflow to generate the text search index table.
  - Input is two SQL queries: all entity instances (e.g. all conditions), all (entity instance, text search string) pairs.
  - The workflow generates a table with one row per entity instance: `(node, text)`, where the `node` column foreign keys to the entity primary key (e.g. `concept_id`) and the `text` column is a concatenation of all text search strings for that instance.
- Generated text search index tables for all `aou_synthetic` entities and updated the underlay to point to these index tables instead of the raw `concept_synonym` table.
  - The text search string for all entities includes the `concept_id`, `concept_name` and `concept_code` columns in the `concept` table, and any rows in the `concept_synonym` table that map to the concept.
  - This means we can search on the concept name, id, code, or synonyms.
- Using these index tables instead of the raw `concept_synonym` table for the `aou_synthetic` entities made the text search queries run much faster (~5-15sec => ~1sec when run in the BQ cloud console). That's probably because the index table is much smaller when split up for each entity. There are fewer concepts per entity and the workflow also consolidated them so there's only one row per entity instance.

This PR also fixes a bad merge in the `aou_synthetic` underlay for brand and visit entity mappings.


tjennison-google (approved):

workflow/src/main/java/bio/terra/tanagra/workflow/BuildTextSearchInformation.java
```diff
@@ -0,0 +1,166 @@
...
+import org.apache.beam.sdk.transforms.ParDo;
+import org.apache.beam.sdk.values.KV;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.TypeDescriptors;
+
+/**
+ * A batch Apache Beam pipeline for building a table that contains all text search information for
+ * an entity.
+ */
+public final class BuildTextSearchInformation {
```
calbach, line 27: I'm anticipating that eventually we may want to stitch all of these together to create a combined indexing pipeline (i.e. where most of the parameters are common and/or the pulled dynamically from the underlay). I'm wondering whether there are any patterns we can establish now to make this easier later

workflow/src/main/java/bio/terra/tanagra/workflow/TextSearchUtils.java
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Concatenate all search strings for a node into a single string. e.g. (123, "apple pie"), (123,
+   * "fruit dessert"), (123, "apple crumble") => (123, "apple pie , fruit dessert , apple crumble")
+   *
+   * @param allNodes a collection of all (node) entity instances
+   * @param searchStrings a collection of all (node, text) mappings, where there may be zero or
+   *     multiple elements per node
+   * @return a collection of (node, text) mappings, where there is exactly one element per node
+   */
+  public static PCollection<KV<Long, String>> buildSingleSearchString(
```
calbach, line 32: nit:  would `concatenateSearchStrings` or  `concatenateSearchStringsByKey` be more accurate?

workflow/src/main/java/bio/terra/tanagra/workflow/TextSearchUtils.java
```diff
@@ -0,0 +1,90 @@
...
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.TupleTag;
+import org.apache.beam.sdk.values.TypeDescriptors;
+
+public final class TextSearchUtils {
+  private TextSearchUtils() {}
+
+  /**
+   * Concatenate all search strings for a node into a single string. e.g. (123, "apple pie"), (123,
+   * "fruit dessert"), (123, "apple crumble") => (123, "apple pie , fruit dessert , apple crumble")
```
calbach, line 25: Are there drawbacks to using "," as the delimiter here? Should we be escaping actual ","s in the input value and search string?

calbach (approved):

workflow/src/main/java/bio/terra/tanagra/workflow/BuildTextSearchInformation.java
marikomedlock: I agree about driving the indexing pipeline from the underlay. For indexing, the inputs we need right now are queries that return the:
- set of all entity instances (e.g. all conditions)
- set of all parent-child relationships between entity instances if there's a hierarchy
- set of all entity instance-text search strings if we want to support text search on the entity

For each query, we can either:
- generate it from the entity mapping in the underlay
- take a SQL query text file

I expect we'll need to support both methods. Specifying an entity mapping in the underlay seems cleaner and we can already do this in backend (e.g. generate a query for all conditions). But some things are too convoluted to specify in the underlay (e.g. parent-child relationships in the ingredient hierarchy exist across two separate tables), and I don't think we should force the underlay to express anything more than basic SQL syntax (i.e. no unions). Maybe one way to handle this is to allow users to specify a SQL query text file in the underlay?

RE patterns to establish now before we've coupled the indexing to the underlay: I think the main thing is to make sure the indexing input queries are general enough to apply to other non-OMOP datasets. So we don't want e.g. a workflow that takes a `domain_id` as input and slices the `concept` table by that domain. But a workflow that takes a parent-child query and flattens the hierarchy into ancestor-descendant or calculates the paths is okay.

workflow/src/main/java/bio/terra/tanagra/workflow/TextSearchUtils.java
marikomedlock: Yes, that's clearer thanks. Changed to `concatenateSearchStringsByKey`.

workflow/src/main/java/bio/terra/tanagra/workflow/TextSearchUtils.java
marikomedlock: Good point. I think we probably should be either escaping whatever the delimiter is in the search strings, or picking a delimiter that's not allowed in the UI search box. For AoU-CB, it looks like the following characters are not allowed: `~ @ [ ] | < > ( )`. I guess we should enforce that not-allowed list on the backend too if we're going to depend on it for indexing.

I'll plan to tackle this as follow-on work and perhaps we can discuss in stand-up whether the not-allowed search characters is something we'd be open to supporting generally.

PR #105, opened 2022-02-09
Auxiliary table entities and relationships for aou_synthetic underlay.

marikomedlock:
- Goal 1: Allow querying the people with a particular condition.
  - Needed for the summary/breakdown view in the cohort builder. "Show me the people in my cohort."
- Goal 2: Allow querying the condition occurrences for people with a particular condition.
  - Needed for the dataset builder preview and export. "Show me all the condition_occurrence rows for people in my cohort."

- Change 1: Define a relationship between the `person` and `condition_occurrence` entities.
  - This relationship is defined by a foreign key. A `person` entity instance is related to a `condition_occurrence` entity instance if the `person_id` columns match.
  - The client can use a `RelationshipFilter` to query for `person` entity instances that are related to a particular `condition_occurrence` entity instance.
- Change 2: Define a relationship between the `condition` and `condition_occurrence` entities.
  - This relationship is defined by a foreign key. A `condition` entity instance is related to a `condition_occurrence` entity instance if the `concept_id` and `condition_concept_id` columns, respectively, match.
  - The client can use a `RelationshipFilter` to query for `condition_occurrence` entity instances that are related to a `condition` entity instance.

All of the above applies to other entity/auxiliary table pairs also (e.g. `visit`/`visit_occurrence`); I just used `condition` as an example.

- Cohort builder example query: All condition occurrences for people who had covid. Note that this will include rows for things other than covid.
  - [occurrences of any condition for people who had covid] `condition_occurrence` entity instances that are related to
  - [people who had covid] `person` entity instances that are related to
  - [occurrences of covid] `condition_occurrence` entity instances that are related to a
  - [covid] `condition` entity instance
- Dataset builder example query: All people who had covid.
  - [people who had covid] `person` entity instances that are related to
  - [occurrences of covid] `condition_occurrence` entity instances that are related to a
  - [covid] `condition` entity instance

- Per an earlier suggestion from CH, I also updated the test harness to use a seeded random number during tests.


api/src/main/resources/underlays/aou_synthetic.yaml
```diff
@@ -1338,6 +1331,34 @@ relationshipMappings:
         dataset: aou_synthetic
         table: concept_relationship
         column: concept_id_2
+  - name: condition_person
```
calbach, line 1334: My expectation was that for the general use case of the cohort builder, we would instead want to create a relationship between condition and condition_occurrence, and then apply two relationship filters ( **or** materialize the results of the condition search, and pass this back in as a literal to a condition_occurence filter ). This is because eventually modifiers will need to be applied at the condition_occurence row level. There are some more complex modifiers as well, including things like "at least X instances" or "occurring within X date".

Creating a direct relationship as you have here would simplify basic presence checks (i.e. does this participant have at least one row matching this condition), but I don't know if it's worth supporting this if we need to implement the general case anyways.

marikomedlock: I'm not sure a direct relationship between `condition` and `condition_occurrence` will produce the right cohort if there are search criteria defined in different domains. e.g.
- Define a cohort = people with covid who've had a mammogram (i.e. condition=covid and procedure=mammogram).
- Define a dataset on the cohort that includes all `condition_occurence` rows.

In this case, I think the `condition_occurrence` rows can't just be filtered on condition=covid, because then we'll include people with covid who haven't had a mammogram. We'd need to first get the set of people with `condition_occurrence` rows for covid and `procedure_occurrence` rows for mammogram. Then filter the `condition_occurrence` rows on that set of people. Does this sound right to you?

RE modifiers: I think we can still handle these. It means that when fetching rows of `condition_occurrence` for dataset builder, we'd use an `ArrayFilter` that contains the nested `RelationshipFilter` on `Person` & `Condition` and an e.g. `BinaryFilter` for checking the date.

calbach: RE condition=covid example: right - here is logically how I imagine constructing your example dataset (just pseudo SQL - not meant to be representative of the best way of writing this):

```
SELECT * FROM condition_occurence WHERE person_id IN (my_cohort);
```

Defining my_cohort as:
```
SELECT person_id FROM person
INNER JOIN (condition_occurrence WHERE concept_id IN (covid concepts))
INNER JOIN (procedure_occurence WHERE concept_id IN (mammogram concepts))
```

(X concepts) could either be another nested filter / JOIN with the concepts table, or a materialized list of concept IDs, generated from a previous search.

RE modifiers: the X occurrences and other modifiers need to be filters applied to the `condition_occurence` entity, not the `condition` entity. Not sure if this was a typo or misunderstanding

This may be easier to talk through. Will setup some time tomorrow (feel free to move or cancel)

marikomedlock: Thanks for setting up the meeting. It will be very helpful to talk through this also.

After thinking about it some more, I think I'm following you better now. Here's a restatement, could you see if it makes sense to you?
- We need to be able to get a cohort of people with condition=covid.
  - One way to do this is to have a `people_condition` relationship and use a single `RelationshipFilter`. This is what I originally implemented.
  - Another way to do this is to have 2 relationships `people-condition_occurrence` and `condition_occurrence-condition` and use a nested `RelationshipFilter`. I pushed up a change to implement this with my latest commit.
  - Both of these will produce very similar queries when getting the list of people for a cohort.
- We also need to be able to get a dataset that includes condition occurrences where condition=diabetes for a cohort. 
  - To do this, we need two relationships `people-condition_occurrence` and `condition_occurrence-condition`. The former will be used for the filter that defines the cohort, and the later will be used for the filter that defines the concept set.
  - Since we're going to need these two separate relationships anyway, then for building a cohort, we can just navigate these same two relationships rather than defining a new one that effectively consolidates the two.
  
I pushed up a change now that I think implements this.

api/src/main/java/bio/terra/tanagra/service/search/utils/RandomNumberGenerator.java
```diff
@@ -0,0 +1,60 @@
...
+   * each test method.
+   */
+  private static boolean useSeededRandom() {
+    String useRandomSeedSysProp = System.getProperty(USE_RANDOM_SEED_SYSTEM_PROPERTY);
+    return useRandomSeedSysProp != null && !useRandomSeedSysProp.isEmpty();
+  }
+
+  /** Return a seeded random generator. */
+  private static Random getSeededRandom() {
+    synchronized (RandomNumberGenerator.class) {
```
calbach, line 48: I don't use much of the Java synchronzied keyword, so I may be missing something basic. What is the benefit of using a synchronized statement here vs making `getSeededRandom` a synchronized method?

calbach (approved):
Changes look reasonable, though I didn't review all of the yaml in close detail. Some questions about Random testing.

api/src/main/java/bio/terra/tanagra/service/search/utils/RandomNumberGenerator.java
marikomedlock: I originally defined the whole method as `synchronized`, but static analysis [prompted](https://pmd.github.io/pmd-6.31.0/pmd_rules_java_multithreading.html#avoidsynchronizedatmethodlevel) me to change it to a `synchronized` block instead so that we're not over-synchronizing in case of future additions to the method.

api/src/main/java/bio/terra/tanagra/service/search/utils/RandomNumberGenerator.java
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * Check if the system property is set. This method is called to set a final static boolean, so
+   * that we only need to check the system property once in the non-testing case, not each time we
+   * need a random number.
+   *
+   * <p>See See bio.terra.tanagra.testing.BaseSpringUnitTest for where this is property set before
+   * each test method.
+   */
+  private static boolean useSeededRandom() {
+    String useRandomSeedSysProp = System.getProperty(USE_RANDOM_SEED_SYSTEM_PROPERTY);
```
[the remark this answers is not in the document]
marikomedlock: I tried this, but I couldn't figure out how to get it to reset the seed before every test method was run, only before every test class was run. So when I ran methods in a different order, I'd get different SQL.

After reading up a bit, I tried a different tack here and used Mockito to mock the bean from the tests instead. I think this looks cleaner than what I had before. PTAL and see what you think. (I'm going to merge this PR now, so we get the main underlay changes in, but happy to adjust the random approach in a future PR.)

PR #107, opened 2022-02-11
Add search bar to concept criteria

tjennison-google:
Currently only in the list view because search in the hierarchy view has
different behavior.


ui/src/search.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export type SearchProps = {
+  placeholder?: string;
+  onSearch: (query: string) => void;
+};
+
+export function Search(props: SearchProps) {
+  return (
+    <Box m={2}>
+      <Form
```
calbach, line 16: Does the enter key work automatically with this pattern?

calbach (approved):
Sorry for the slow review, lgtm

ui/src/search.tsx
tjennison-google: Yeah, that's why I used the form.

PR #108, opened 2022-02-11
Define UI hints for demographic person attributes.

marikomedlock:
- Added an `EntityFiltersSchema` section to the `aou_synthetic` underlay. There was already support for this on the backend, and an [example](https://github.com/DataBiosphere/tanagra/blob/1ae6a1db544dec3d8b6dbb4c011516406b993b50/api/src/test/resources/underlays/nautical.yaml#L419) in the test `nautical` underlay. This section holds hints for the UI about entity attributes.
- Defined UI hints for the following `person` entity attributes, to support the Demographics search criteria sub-menu in the AoU-CB.
  - `year_of_birth` : `IntegerBoundsHint`. AoU-CB has an age demographic criteria, but we don't have a way to specify a dynamically calculated attribute, yet. Using the birth year as a placeholder here, because it's an integer value and we can use it to build the numerical slider in the UI.
  - `ethnicity_concept_id` : `EnumHint`
  - `gender_concept_id` : `EnumHint`
  - `race_concept_id` : `EnumHint`
  - `sex_at_birth_concept_id` : `EnumHint`
- I specified these hints manually. i.e. I manually ran queries in the BQ cloud console to find the list of possible values for each field. Then I hard-coded `concept_id=0` to "Unknown" and trimmed the strings, which is what the AoU-CB indexing [bash scripts](https://github.com/all-of-us/workbench/blob/5577e8d3c1cd9241fd66c3ca1a5a62dd27369755/api/db-cdr/generate-cdr/cb-criteria-parallel/make-cb-criteria-demographics.sh#L210) are doing.
  - e.g. for `year_of_birth`:
```
SELECT min(p.year_of_birth), max(p.year_of_birth) 
FROM broad-tanagra-dev.aou_synthetic_SR2019q4r4.person p
```
  - e.g. for `ethnicity_concept_id`:
```
SELECT c.concept_id, c.concept_name
FROM broad-tanagra-dev.aou_synthetic_SR2019q4r4.person p
JOIN broad-tanagra-dev.aou_synthetic_SR2019q4r4.concept c
on c.concept_id=p.ethnicity_concept_id
GROUP BY c.concept_id, c.concept_name
```
- Eventually, we want these integer bounds and enum values to be generated from the underlay. However, since we're not yet at the stage of driving indexing or other pipelines from the underlay definition, I've defined them manually for now. This will allow the UI to consume the hints, and we can work on populating them automatically as a separate piece of work.
- The UI can see these hints as part of the entity definition. I included a test that fetches the `person` entity and checks for these UI hints.


calbach (approved):
Makes sense to me, thanks

PR #109, opened 2022-02-13
Add optional order by attribute to API entity instances endpoint.

marikomedlock:
- Added an optional `orderByAttribute` property to the `EntityDataset` API object definition. This property specifies an attribute of the primary entity to order the results by.
- The order by attribute does not need to be one of the selected attributes. e.g. You can select just `concept_id` and order by `concept_name`.
- Updated the test to fetch all `condition` entity instances for the `aou_synthetic` underlay to order by the `concept_name` attribute.
- Eventually for the `aou_synthetic` underlay, we want to order by the rollup counts. Related work to make that happen includes: generating an index table with the rollup counts, adding a pointer to this index table in the underlay, adding a new generated attribute to the entity (similar to what we do for the path). At that point, we can use this `orderByAttribute` in the API requests.
- In the meantime, this can be used to order e.g. conditions alphabetically.


api/src/main/resources/api/service_openapi.yaml
```diff
@@ -364,9 +364,21 @@ components:
           type: array
           items:
             type: string
+        orderByAttribute:
```
tjennison-google, line 367: This should be an array of attribute/direction pairs for flexibility since it's easy to support in SQL.

marikomedlock: Good suggestion, thanks. Do we have a use case for multiple order-by criteria currently? If not, then l'd like to address it in a follow-on PR.

tjennison-google: We don't right now but it's pretty common. Just suggesting it while you're defining the API. Putting it off for now is fine though since we don't immediately need it.

tjennison-google (approved):

calbach (approved):

api/src/main/resources/api/service_openapi.yaml
marikomedlock: Sounds good, I'll add it to the list of future enhancements then. Thanks

api/src/main/java/bio/terra/tanagra/service/query/QueryService.java
```diff
@@ -102,11 +102,23 @@ Query createQuery(EntityDataset entityDataset) {
...
+    Query.Builder queryBuilder =
+        Query.builder()
+            .selections(selections)
+            .primaryEntity(entityDataset.primaryEntity())
+            .filter(entityDataset.filter());
+
+    if (entityDataset.orderByAttribute() != null) {
+      Selection orderBy =
+          selections.stream()
+              .filter(selection -> selection.name().equals(entityDataset.orderByAttribute().name()))
```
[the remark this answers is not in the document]
marikomedlock: Agreed, my thought was to lift this restriction in the future -- this was just as far as I got last week. After a second look though, this turned out to be not too much more code, so I went ahead and lifted the restriction now.

PR #110, opened 2022-02-15
Generalize concept criteria and add support for procedures

tjennison-google:


ui/src/cohort.ts
```diff
@@ -68,12 +68,20 @@ function generateFilter(group: Group): tanagra.Filter | null {
...
+  config: CriteriaConfig;
   name: string;
   count: number;
   data: unknown;
 }
 
+// CriteriaConfigs are used to initialize CriteriaPlugins and provide a list of
+// possible criteria.
+export interface CriteriaConfig {
+  type: string;
```
calbach, line 80: Could you add some documentation specifically for this property? I'm not sure if this is a UI concept, a Tanagra API concept, or something else

ui/src/overview.tsx
```diff
@@ -91,17 +90,31 @@ function AddCriteriaButton(props: { group: string | GroupKind }) {
...
   );
 
-  const items = getCriteriaTitles();
+  // TODO(tjennison): Fetch configs from the backend.
+  const configs = [
+    {
+      type: "concept",
+      title: "Conditions",
+      defaultName: "Contains Conditions Codes",
+      entity: "condition",
```
calbach, line 99: Is this supposed to be the same interface as CriteriaConfig? Why is `entity` not defined there?

Also - may be worth typing this array

ui/src/overview.tsx
```diff
@@ -91,17 +90,31 @@ function AddCriteriaButton(props: { group: string | GroupKind }) {
     [props.group]
   );
 
-  const items = getCriteriaTitles();
+  // TODO(tjennison): Fetch configs from the backend.
+  const configs = [
+    {
+      type: "concept",
```
calbach, line 96: Does this need some kind of linkage to the `condition_occurrence` table as well? Or not yet?

calbach (approved):

ui/src/cohort.ts
tjennison-google: Done.

ui/src/overview.tsx
tjennison-google: It will at some point to handle the modifiers. We're still going to need to planning to figure out how to handle those connections generically.

ui/src/overview.tsx
tjennison-google: That comes from the concept plugin config. I've broken that out into a separate object to better indicate that.

PR #114, opened 2022-02-23
Add custom checkbox

tjennison-google:
This makes the checkboxes and expansion icons in the criteria be the
same size. Because the Material UI Checkbox and IconButton are built out
of different components, making them the same size is much more
complicated than it should be.

### DIFF
--- a/ui/src/checkbox.tsx
+++ b/ui/src/checkbox.tsx
@@ -1,0 +1,35 @@
+import CheckBoxIcon from "@mui/icons-material/CheckBox";
+import CheckBoxOutlineBlankIcon from "@mui/icons-material/CheckBoxOutlineBlank";
+import IconButton from "@mui/material/IconButton";
+
+export type CheckboxProps = {
+  checked?: boolean;
+  onChange?: () => void;
+  size?: "small" | "medium" | "large";
+  fontSize?: "small" | "medium" | "large" | "inherit";
+};
+
+export default function Checkbox({
+  checked,
+  onChange,
+  size,
+  fontSize,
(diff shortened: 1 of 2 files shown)


ui/src/checkbox.tsx
```diff
@@ -0,0 +1,30 @@
...
+import IconButton from "@mui/material/IconButton";
+
+export type CheckboxProps = {
+  checked?: boolean;
+  onChange?: () => void;
+  size?: "small" | "medium" | "large";
+  fontSize?: "small" | "medium" | "large" | "inherit";
+};
+
+export default function Checkbox(props: CheckboxProps) {
```
calbach, line 12: opt: personal style preference, but destructuring the props here can be nice, i.e. `Checkbox({checked, onChange, size, fontSize}: CheckboxProps)`

ui/src/checkbox.tsx
```diff
@@ -0,0 +1,30 @@
+import CheckBoxIcon from "@mui/icons-material/CheckBox";
```
calbach, line 1: Are there enough files here to start adding some directory structure? e.g. maybe a `components/` subdir?

calbach (approved):

ui/src/checkbox.tsx
tjennison-google: Done.

ui/src/checkbox.tsx
tjennison-google: Yeah, I have been thinking it's getting to that point too. I'll look at a `components/` directory at minimum and possibly others.

PR #116, opened 2022-02-25
Fetch underlay configuration from the backend

tjennison-google:
It is stored for later access to hints and other information.


ui/src/app.tsx
```diff
@@ -17,25 +22,65 @@ import Overview from "./overview";
...
+            }
+            return entitiesApi.listEntities({ underlayName: u.name });
+          })
+        ).then((entitiesResList) => {
+          dispatch(
+            setUnderlays(
+              entitiesResList.map((entitiesRes, i) => {
+                const name = res.underlays?.[i]?.name;
+                if (!name) {
+                  throw "Unnamed underlay.";
```
calbach, line 53: nit: this should be unreachable due to above check at L42

ui/src/app.tsx
```diff
@@ -17,25 +22,65 @@ import Overview from "./overview";
...
+                }
+                if (!entitiesRes.entities) {
```
calbach, line 55: Should this check length as well?

ui/src/app.tsx
```diff
@@ -17,25 +22,65 @@ import Overview from "./overview";
...
+                  throw `No entities in underlay ${name}`;
+                }
+
+                return {
+                  name: name,
```
calbach, line 60: nit: `name,`

ui/.env.test
```diff
@@ -0,0 +1 @@
+REACT_APP_USE_FAKE_API=y
```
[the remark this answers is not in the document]
tjennison-google: The flag sets up the fake API and `.env.test` is pulled in whenever tests are being run. Previously none of the tests resulted in network requests so it wasn't necessary before.

ui/src/app.tsx
```diff
@@ -17,25 +22,65 @@ import Overview from "./overview";
...
 export default function App(props: AppProps) {
+  const dispatch = useAppDispatch();
+  const underlaysApi = useContext(UnderlaysApiContext);
+  const entitiesApi = useContext(EntitiesApiContext);
+
+  const underlaysState = useAsyncWithApi(
+    useCallback(() => {
+      return underlaysApi.listUnderlays({}).then((res) => {
+        if (!res?.underlays || res.underlays.length == 0) {
+          throw "No underlays are configured.";
```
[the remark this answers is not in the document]
tjennison-google: No, if throwing Error is standard that's what I'll do as well.

ui/src/app.tsx
tjennison-google: Yes, but unfortunately the compiler can't confirm it so it complains without the extra check.

ui/src/app.tsx
tjennison-google: I don't think it needs to. For our current schemas we'd always expect there to be entities but it doesn't seem like an error in general.

ui/src/app.tsx
tjennison-google: Done.

ui/src/app.tsx
```diff
@@ -17,25 +22,65 @@ import Overview from "./overview";
...
   entityName: string;
 };
 
 export default function App(props: AppProps) {
+  const dispatch = useAppDispatch();
+  const underlaysApi = useContext(UnderlaysApiContext);
+  const entitiesApi = useContext(EntitiesApiContext);
+
+  const underlaysState = useAsyncWithApi(
+    useCallback(() => {
```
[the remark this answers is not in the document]
tjennison-google: Done. I agree it's more readable.

calbach (approved):

PR #119, opened 2022-03-03
Standardize current underlay and cohort access

tjennison-google:
* Adds common functions for accessing the current underlay and cohort
  based on the current URL. The higher components redirect when
  presented with invalid values so the lower components don't need to
  worry about them.
* Handling the multi-level redirects and id checking with fixed URL
  pieces (i.e.  `/underlays/`) ends up being surprisingly complicated,
  at least with this version of react-router. We could look into options
  if desired but for now I've just removed them.
* Adds an inital underlay selection page. The previous behavior of
  choosing per cohort is weird once concept sets are introduced because
  they can't be mixed and matched.  The selection page could instead be
  a dropdown in the action bar or a menu but that's awkward too because
  there's nothing to do when switching except jump back to the initial
  datasets page.  We'll need to give more thought to what behavior we
  actually want when there are multiple underlays.
* Moves criteria config to the App where it can eventually be fetched
  along with the other underlay config and changes accesses to go
  through the common functions.
* Removes underlay selection from the New Cohort dialog since it's
  determined by the selection page.

### DIFF
--- a/ui/src/hooks.ts
+++ b/ui/src/hooks.ts
@@ -1,5 +1,40 @@
 import { TypedUseSelectorHook, useDispatch, useSelector } from "react-redux";
+import { useParams } from "react-router-dom";
 import { AppDispatch, RootState } from "store";
 
 export const useAppDispatch = () => useDispatch<AppDispatch>();
 export const useAppSelector: TypedUseSelectorHook<RootState> = useSelector;
+
+export const useUnderlay = () => {
+  const { underlayName } = useParams<{ underlayName: string }>();
+  return useAppSelector((state) =>
+    state.underlays.find((underlay) => underlay.name === underlayName)
+  );
+};
+
+// The useXOrFail versions are generally the correct choice since they can be
+// used where the existence of X is enforced by higher level parts of the UI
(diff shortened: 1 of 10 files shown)


ui/src/hooks.ts
```diff
@@ -1,5 +1,36 @@
...
 export const useAppSelector: TypedUseSelectorHook<RootState> = useSelector;
+
+export const useUnderlay = () => {
+  const { underlayName } = useParams<{ underlayName: string }>();
+  return useAppSelector((state) =>
+    state.underlays.find((underlay) => underlay.name === underlayName)
+  );
+};
+
+export const useUnderlayOrFail = () => {
```
calbach, line 15: Could you add some comments here? What is the guidance on when to use this vs `useUnderlay`? Are there any considerations around the time during which the underlay is still loading?

calbach (approved):

ui/src/hooks.ts
tjennison-google: Done. The idea is that the UI enforces the existence of a valid underlay before allowing access to lower level pages so the useXOrFail versions allow the code to ignore the possibility of them being undefined.

PR #121, opened 2022-03-11
Add initial concept set support

tjennison-google:
This change adds initial support for actual concept sets along with two
sample prepackaged concept sets.
* The primary entity (i.e. person) was moved from a hardcoded value to a
  property of the underlay configuration.
* The existing fake concept st was replaced with real concept sets tied
  to actual entities.
* The SQLDialog was replaced by a panel with tabs for each relevant
  entity to display the SQL and eventually the preview.
* Concept criteria query generation was made more flexible. I had
  originally planned to tie the prepackaged concept sets into criteria
  since that would better align with the behavior of custom concept
  sets. That ended up requiring more boilerplate to define them than I
  anticipated so I opted to directly use filters instead. The
  flexibility will still be required for custom concept sets though so
  I've left it in.


calbach:
Sorry in advance, due to launch activities earliest I will probably be able to review this is Tuesday. Feel free to reassign if anyone else has capacity.

ui/cypress/integration/tests.js
```diff
@@ -4,5 +4,17 @@ describe("Basic tests", () => {
...
+    cy.get("button:Contains(Create)").click();
+    cy.get("a[id^=edit-cohort]").click();
+    cy.get("button:Contains(Add Criteria)").click();
+    cy.get("li:Contains(Conditions)").click();
+    cy.get("button[role=checkbox]").click();
+    cy.get("a[aria-label=back]").click();
+    cy.get("a[aria-label=back]").click();
+    cy.get("input[name='New Cohort']").click();
+    cy.get("input[name=Demographics]").click();
+    cy.contains("SELECT *");
```
calbach, line 18: I haven't used cypress before. Am I correct in understanding this just asserts that the text "SELECT *" appears anywhere on the page?

Should the selector be more targeted?

ui/src/criteria/concept.tsx
```diff
@@ -92,7 +92,7 @@ class _ implements CriteriaPlugin<Data> {
     return <ConceptDetails data={this.data} />;
   }
 
-  generateFilter(entityVar: string) {
+  generateFilter(entityVar: string, fromOccurrence: boolean) {
```
calbach, line 95: nit: After reading the code I'm still a bit unclear on the semantics of `fromOccurence`, consider adding a comment here

calbach (approved):

PR #122, opened 2022-03-16
Consolidate routing logic

tjennison-google:
* Define the page and URL layout into a single location so that future
  changes aren't spread across many locations.
* Automate error handling of non-existing pages and pages that reference
  invalid objects.
* Automate action bar back button handling.
* Provide a way for components to link to other pages without hardcoding
  the layout.


ui/src/actionBar.tsx
```diff
@@ -7,14 +7,20 @@ import Typography from "@mui/material/Typography";
...
   title: string;
-  backUrl?: string;
 };
 
 export default function ActionBar(props: ActionBarProps) {
-  const underlay = useUnderlay();
+  let underlay: Underlay | undefined;
+  try {
+    underlay = useUnderlay();
+  } catch {}
```
calbach, line 21: Did you mean to leave this in? TODO or logging to handle this? If it's truly needed I would add a comment - blanket exception supression like this can obfuscate programming errors

ui/src/router.tsx
```diff
@@ -0,0 +1,94 @@
...
+  "/:underlayName/cohorts/:cohortId": Overview,
+  "/:underlayName/cohorts/:cohortId/edit/:groupId/:criteriaId": Edit,
+};
+
+export function AppRouter() {
+  return (
+    <ErrorBoundary FallbackComponent={NotFound}>
+      <Switch>
+        {Object.entries(pages).map(([page, component]) => (
+          <Route exact key={page} path={page} component={component} />
```
calbach, line 28: I'm surprised typescript likes this. You didn't need to apply stronger explicit typing to `pages`?

ui/src/router.tsx
```diff
@@ -0,0 +1,94 @@
...
+        ))}
+        <Route path="/">
+          <NotFound error={new PathError("Invalid URL.")} />
+        </Route>
+      </Switch>
+    </ErrorBoundary>
+  );
+}
+
+// useBackUrl returns the second longest matching page, assuming that the
```
calbach, line 38: I can imagine cases where this logic won't work as expected, for example if you're in a flow between siblings. Are there issues using the history more explicitly for the current usages?

Alternatively, name this something like `useParentUrl()`

ui/src/router.tsx
```diff
@@ -0,0 +1,94 @@
...
+// longest is the current page since longer ones wouldn't match.
+export function useBackUrl(): string | undefined {
+  const page = Object.keys(pages)
+    .filter((page) => useRouteMatch({ path: page }))
```
calbach, line 42: This breaks the first rule of hooks: https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

ui/src/actionBar.tsx
tjennison-google: You're right, this is confusing and asking for trouble down the road. This is the one place an undefined underlay is potentially expected. Instead I've included the code to fetch the underlay directly since that seems better than a separate function that doesn't throw plus a comment explaining why it shouldn't generally be used.

ui/src/router.tsx
tjennison-google: So I had originally built a more complicated version that explicitly specified the back URLs for each page but it was more complicated for no benefit. I'm not planning any behaviour that wouldn't fit this model but this can be revisited if something does come up.

Using history explicitly is more of a UX decision. In my opinion, the back button on the page is more of an "up" button allowing the user to get to the parent page. There's no reason to duplicate the browser's back button and having it work this way is actually helpful in situations where there is more complicated navigation since it guarantees a way to get to the parent even if that's not where the user came from.

ui/src/router.tsx
tjennison-google: All of the components take no props so they have the same type. It make sense since they're all top level pages that fetch their data from the the URL.

ui/src/router.tsx
tjennison-google: This usage is okay because it's iterating over a constant array. The order and number of hooks never changes.

ui/src/router.tsx
calbach: How about something like `useParentUrl()`, then? `useBackUrl()` seems overly prescriptive on where you're expecting this to be used the app, and inaccurate based on what it's doing.

ui/src/router.tsx
calbach: I would say it still violates the rule, even if it doesn't break the underlying premise. In terms of readability, I think this also requires too much derivation to conclude that it's safe.

If there's no other way of implementing this without hooks, then I think at least a comment here is warranted.

calbach (approved):

ui/src/router.tsx
tjennison-google: Done.

ui/src/router.tsx
tjennison-google: Done. I found a way to implement it without without using a hook in the loop.

PR #124, opened 2022-03-18
Add support for use generated concept sets

tjennison-google:
* Add concept sets to the store.
* Add concept editing page.
* Add user generated concept sets to datasets page and menu to create
  them.
* Return concept sets from criteria plugins. Filter generation needs to
  be refactored to clean up occurrence code and support multiple concept
  sets.

### DIFF
--- a/ui/src/cohort.ts
+++ b/ui/src/cohort.ts
@@ -98,6 +98,7 @@ export interface CriteriaPlugin<DataType> {
     entityVar: string,
     fromOccurrence: boolean
   ) => tanagra.Filter | null;
+  occurrenceEntities: () => string[];
 }
 
 // registerCriteriaPlugin is a decorator that allows criteria to automatically
--- a/ui/src/datasets.tsx
+++ b/ui/src/datasets.tsx
@@ -2,13 +2,14 @@ import AddIcon from "@mui/icons-material/Add";
 import EditIcon from "@mui/icons-material/Edit";
 import Box from "@mui/material/Box";
 import Button from "@mui/material/Button";
-import Checkbox from "@mui/material/Checkbox";
 import Dialog from "@mui/material/Dialog";
 import DialogActions from "@mui/material/DialogActions";
 import DialogContent from "@mui/material/DialogContent";
 import DialogTitle from "@mui/material/DialogTitle";
 import Grid from "@mui/material/Grid";
 import IconButton from "@mui/material/IconButton";
+import Link from "@mui/material/Link";
+import MenuItem from "@mui/material/MenuItem";
 import Paper from "@mui/material/Paper";
 import Stack from "@mui/material/Stack";
 import Tab from "@mui/material/Tab";
@@ -17,9 +18,18 @@ import TextField from "@mui/material/TextField";
 import Typography from "@mui/material/Typography";
 import ActionBar from "actionBar";
 import { EntityInstancesApiContext } from "apiContext";
+import {
+  createCriteria,
+  Criteria,
+  generateQueryFilter,
+  getCriteriaPlugin,
+} from "cohort";
 import { insertCohort } from "cohortsSlice";
+import Checkbox from "components/checkbox";
 import Loading from "components/loading";
+import { useMenu } from "components/menu";
 import { TreeGridData } from "components/treegrid";
+import { insertConceptSet } from "conceptSetsSlice";
 import { useAsyncWithApi } from "errors";
 import { useAppDispatch, useAppSelector, useUnderlay } from "hooks";
 import {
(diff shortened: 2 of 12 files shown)


ui/src/cohort.ts
```diff
@@ -98,6 +98,7 @@ export interface CriteriaPlugin<DataType> {
     entityVar: string,
     fromOccurrence: boolean
   ) => tanagra.Filter | null;
+  conceptSetEntities: () => string[];
```
calbach, line 101: I believe this is diverging a bit from how the AoU dataset builder defines "Criteria" and "Cohorts". In AoU cohort builder:

dataset contains cohorts
dataset contains concept sets
cohort contains criteria

Cohort criteria in AoU would have no bearing on what concept sets can be selected.

If "criteria" here means something different, a README or doc that explains the terminology probably makes sense, to avoid confusion with the existing AoU terminology.

ui/src/datasets.tsx
```diff
@@ -81,6 +91,64 @@ export function Datasets() {
...
+    conceptSets: { id: string; name: string }[]
+  ) => {
+    return conceptSets.map((conceptSet) => (
+      <Stack key={conceptSet.name} direction="row" alignItems="center">
+        <Checkbox
+          size="small"
+          fontSize="inherit"
+          name={conceptSet.name}
+          checked={selectedConceptSets.has(conceptSet.id)}
+          onChange={() => onToggle(updateSelectedConceptSets, conceptSet.id)}
```
calbach, line 105: This would also be an existing issue with cohorts, if it is an issue: what happens when you switch underlays? Is the redux data being reset somewhere?

calbach (approved):

ui/src/datasets.tsx
tjennison-google: Fixed. The cohorts and concept sets are stored for all underlays and filtered depending on which is active. This was done correctly for cohorts but I forgot to do it for concept sets.

ui/src/cohort.ts
tjennison-google: I've defined concept sets as a single criteria. Except for some minor differences, that's how they appears in the UI. So when a criteria is used as part of a concept set it needs to know which entities it applies to. Does AoU implement them differently in some way and just mimic the UI?

calbach: > I've defined concept sets as a single criteria

I'm not following this, can you rephrase? Are you saying you're defining a concept set the same way as a cohort, except you only allow it to contain a single criteria?

I think the commonality between criteria and concept sets in AoU, is that both may use concept search. Once you reach that point, they do share the same search UI in AoU. The biggest distinction I would say are modifiers. If we were to allow modifiers in concept sets, we would want to come up with a different term for them. Right now, it's usage in the product depends on its semantics matching the name: it's a set of concept ids.

tjennison-google: > > I've defined concept sets as a single criteria
> 
> I'm not following this, can you rephrase? Are you saying you're defining a concept set the same way as a cohort, except you only allow it to contain a single criteria?

Not exactly. I mean that the UI and data (i.e. one or more concept ids for conditions) are the same for both a concept set and an individual criteria within a cohort so I'm reusing both for concept sets. There will need to be some UI adjustments for each case (e.g. not showing modifiers) but it makes sense to reuse the existing code.

> I think the commonality between criteria and concept sets in AoU, is that both may use concept search. Once you reach that point, they do share the same search UI in AoU. The biggest distinction I would say are modifiers. If we were to allow modifiers in concept sets, we would want to come up with a different term for them. Right now, its usage in the product depends on its semantics matching the name: it's a set of concept ids.

Cohorts and concept sets seem like the same thing to me, except that one applies to people and the other applies to occurrences. Conceptually, things like exclusion criteria and modifiers could be equally applicable to concept sets. The may be UX or performance reasons for some differences but I think it makes sense to start from a unified position and customize individual pieces as needed.

Is there a reason concept search and modifiers don't exist for cohorts? It seems like they would be just as useful there, and my understanding from the AoU sprint planning is that search is actually in the works. It looks to me like most of the UI divergences between cohorts and concept sets are deficiencies rather than intentional designs, except for things like the demographics breakdown that obviously doesn't make sense.

calbach: Thanks, that helps to clarify. Yes: I think you're correct on the cohorts vs concept sets point. I'm not sure whether I would characterize the differences as deficiencies or intentional simplifications. If we took things in that direction though, we should name it something different, and it would have slightly different product semantics.

Getting back to the line I commented on, I think what I find confusing at this point is that you have a property on criteria named `conceptSetEntities`. Should this property name be generalized in some way, e.g. occurenceEntities? If concept sets are really just a particular limited presentation of a criteria?

tjennison-google: I agree, `occurrenceEntities` is a better name. Done.

PR #127, opened 2022-03-25
Add preview support for datasets

tjennison-google:
* Adds ellipses for TreeGrid titles, wrapping for TreeGrid body
  cells, and support for TreeGrids wider than their parent.
* Refactors the list of entity filters and attributes into a shared
  function so it can be used for both the preview and the attribute
  selection.

### DIFF
--- a/ui/src/datasets.tsx
+++ b/ui/src/datasets.tsx
@@ -1,6 +1,5 @@
 import AddIcon from "@mui/icons-material/Add";
 import EditIcon from "@mui/icons-material/Edit";
-import Box from "@mui/material/Box";
 import Button from "@mui/material/Button";
 import Dialog from "@mui/material/Dialog";
 import DialogActions from "@mui/material/DialogActions";
@@ -12,6 +11,7 @@ import Link from "@mui/material/Link";
 import MenuItem from "@mui/material/MenuItem";
 import Paper from "@mui/material/Paper";
 import Stack from "@mui/material/Stack";
+import Switch from "@mui/material/Switch";
 import Tab from "@mui/material/Tab";
 import Tabs from "@mui/material/Tabs";
 import TextField from "@mui/material/TextField";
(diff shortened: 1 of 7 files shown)


ui/src/datasets.tsx
```diff
@@ -223,14 +235,47 @@ export function Datasets() {
             <Typography variant="h4">3. Values</Typography>
             <Typography variant="h5">(Columns)</Typography>
           </Stack>
-          <Paper>{/* TODO(tjennison): Implement attribute selection. */}</Paper>
+          <Paper
+            sx={{ overflowY: "auto", display: "block" }}
+            className="datasets-select-panel"
+          >
+            {conceptSetEntities.map((entity) => (
+              <>
```
calbach, line 243: nit: should there be a React key here?

ui/src/datasets.tsx
```diff
@@ -425,13 +487,40 @@ function Preview(props: PreviewProps) {
...
+                underlayName: underlay.name,
+                searchEntityInstancesRequest: {
+                  entityDataset,
+                },
+              });
+
+              const data: TreeGridData = {
+                root: { data: {}, children: [] },
+              };
+              res?.instances?.slice(0, 100)?.forEach((instance, i) => {
```
calbach, line 501: Presumably this slice is only here because we're currently returning way too much data. Should there be some kind of TODO here around setting a maxResults  / pageSizeabove instead?

ui/src/util/valid.tsx
```diff
@@ -0,0 +1,3 @@
+export function isValid<Type>(arg: Type) {
+  return arg !== null && typeof arg !== "undefined";
```
calbach, line 2: I'm curious about the `typeof arg !== "undefined"` check. What cases does this cover beyond the usual `arg !== undefined` ?

calbach (approved):

PR #129, opened 2022-03-29
Gy/submenu

ginaygoogle:
- submenu creat
- subpage for enum type 
- deal with conflicts

### DIFF
--- a/ui/src/actionBar.tsx
+++ b/ui/src/actionBar.tsx
@@ -11,6 +11,7 @@ import { useParentUrl } from "router";
 
 type ActionBarProps = {
   title: string;
+  backUrl?: string;
 };
 
 export default function ActionBar(props: ActionBarProps) {
--- a/ui/src/app.tsx
+++ b/ui/src/app.tsx
@@ -22,77 +22,156 @@ const columns = [
   { key: "concept_code", width: 120, title: "Code" },
 ];
 
-const criteriaConfigs = [
+const criteriaMenu = [
   {
-    type: "concept",
-    title: "Conditions",
-    defaultName: "Contains Conditions Codes",
-    plugin: {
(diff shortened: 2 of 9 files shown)


ginaygoogle:
Question: The UI unit tests could not pass because we changed the data structure for the concept. Should I modify the concet.test.tsx with the submenu commit? or separate a new commit only for the test file modification? @tjennison-google

ui/src/actionBar.tsx
```diff
@@ -11,6 +11,7 @@ import { useParentUrl } from "router";
 
 type ActionBarProps = {
   title: string;
+  backUrl?: string;
```
ginaygoogle, line 14: @tjennison-google I supposed modification in actionBar.tsx and menu.tsx and several changes in overview.tsx are belongs to your commit. They shows in my commit maybe because I manually deal with the conflict when I commit it. I know I should make my commit clean and only contain the features code. But I am not sure if I still can commit/push after I checkout these changes later?

tjennison-google: It looks like you're changes have already been rebased onto main, so there aren't any particularly special steps left. I would think that you had to fix some merge conflicts when you did that and this is just an instance where you didn't get it quite right so you commit has a change you didn't intend for it to have. In this case, that's re-adding a like I deleted.

You should go through your code and remove any changes like this that aren't intentionally part of the commit.

tjennison-google:
Before worrying about the failing tests or other details, the next step now that you've removed the unnecessary file changes is to split the two commits onto separate branches so they are independent. The attribute page change is more important so focus on getting that one working first. Since it will be separate from the submenus at that point, just add the criteria directly into the menu along with `Condition`, `Procedures`, etc.

PR #130, opened 2022-03-31
TNG-14: add attribute subpage for enum

ginaygoogle:


ui/src/criteria/attribute.tsx
```diff
@@ -0,0 +1,163 @@
...
+        props.data.selected.map(({ id, name }) => (
+          <Stack direction="row" alignItems="baseline" key={id}>
+            <Typography variant="body1">{id}</Typography>&nbsp;
+            <Typography variant="body2">{name}</Typography>
+          </Stack>
+        ))
+      )}
+    </>
+  );
+}
```
tjennison-google, line 163: The usual standard to end files with newlines. See if you can configure your editor to do it automatically.

tjennison-google (changes requested):
Some comments to get you started. It may take a few times back and forth to get it ready to submit.

ui/src/criteria/attribute.tsx
```diff
@@ -0,0 +1,163 @@
...
+  registerCriteriaPlugin,
+} from "cohort";
+import produce from "immer";
+import React from "react";
+import * as tanagra from "tanagra-api";
+import { updateCriteriaData } from "../cohortsSlice";
+import { useAppDispatch, useUnderlay } from "../hooks";
+
+type Selection = {
+  entity: string;
```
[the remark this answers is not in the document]
ginaygoogle: @tjennison-google  Tim. I suppose in the attribute page. we should support to select multiple entities. For example, Like the all of us, the race works multi selections.

ui/src/criteria/attribute.tsx
ginaygoogle: Got it. Done from setting-general-every save with one line

ui/src/criteria/attribute.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+  function hintDisplayName(hint: tanagra.EnumHintValue) {
+    return hint.displayName || "Unknown Value";
+  }
+
+  const enumHintValues = underlay?.entities
+    .filter((g) => g.name === underlay.primaryEntity)[0]
+    .attributes?.filter((g) => g.name === attributeName)[0].attributeFilterHint
+    ?.enumHint?.enumHintValues;
+  const index =
```
tjennison-google, line 77: This approach works but is difficult to understand. Instead, create a function that finds the index for a single id when needed:

```
const selectionIndex = (hint: tanagra.EnumHintValue) =>
  props.data.selected.findIndex((row) => row.id === hint.attributeValue?.int64Val))
```

ui/src/criteria/attribute.tsx
```diff
@@ -0,0 +1,141 @@
...
+  const dispatch = useAppDispatch();
+  const underlay = useUnderlay();
+  const attribute = props.data.entity;
+  const attributeName = verifyName(attribute);
+
+  function verifyName(demographic: string) {
+    return demographic.concat("_concept_id");
+  }
+
+  function hintDisplayName(hint: tanagra.EnumHintValue) {
```
tjennison-google, line 69: For inline functions we prefer using arrow functions:
```
const hintDisplayName = (hint) => ...
```

tjennison-google (changes requested):

ui/src/criteria/attribute.tsx
```diff
@@ -45,6 +46,11 @@ class _ implements CriteriaPlugin<Data> {
   }
 
   generateFilter(entityVar: string, fromOccurrence: boolean) {
+    if (this.data.selected.length === 0) {
+      return null;
+    }
+    console.log(entityVar);
```
ginaygoogle, line 52: before I finish the generateFilter. It will fail the test due to warning (do not use the parameters, entityVar, formOccurence).So just put 2 console there.

ui/cypress/integration/tests.js
```diff
@@ -11,6 +11,11 @@ describe("Basic tests", () => {
     cy.get("li:Contains(Conditions)").click();
     cy.get("button[role=checkbox]").click();
     cy.get("a[aria-label=back]").click();
+
+    cy.get("button:Contains(Add Criteria)").first().click();
+    cy.get("li:Contains(Race)").click();
+    // required attribute.test.tsx cy.get("button[role=checkbox]").click();
```
tjennison-google, line 17: What does this mean? Why is it a comment?

ui/src/criteria/attribute.tsx
tjennison-google: For cases like this you should add a comment of the form:
```
// TODO(ginay): Implement generateFilter.
```

The important part is the TODO since that's a standard marker so we can find similar instances across the code.

You can also combine these into a single statement:
```
console.log(entityVar, fromOccurrence);
```

ui/src/criteria/attribute.tsx
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  occurrenceEntities() {
+    return [];
+  }
+}
+
+function AttributeEdit(props: AttributeEditProps) {
+  const underlay = useUnderlay();
+  const attributeName = props.data.entity;
```
tjennison-google, line 64: This can't be correct. It's either an entity or an attribute. From the usage it looks like it's an attribute so change it to `attribute` in `Data`.

ui/src/criteria/attribute.tsx
```diff
@@ -0,0 +1,131 @@
...
+    .find((g) => g.name === underlay.primaryEntity)
+    ?.attributes?.find((attribute) => attribute.name === attributeName)
+    ?.attributeFilterHint?.enumHint?.enumHintValues;
+
+  const selectionIndex = (hint: tanagra.EnumHintValue) =>
+    props.data.selected.findIndex(
+      (row) => row.id === hint.attributeValue?.int64Val
+    );
+
+  return (
```
tjennison-google, line 79: Add a check for `enumHintValues?.length === 0` and display a message so that it's not a completely blank page in that case.

tjennison-google (changes requested):

ui/cypress/integration/tests.js
ginaygoogle: because I do not have the fake content with checkboxes in the attribute page to click. (So, require attribute.test.tsx file)

tjennison-google: You should add whatever fake data you need to apiContext.ts so that you can write a meaningful test.

ginaygoogle: after communication -- 
add fake data in apiContext.ts within FakeEntitiesApi-listentities-entities

ui/src/criteria/attribute.tsx
ginaygoogle: got it. change the data structure in App.tsx file from "entity" to "attribute". and also change the following parameter in attribute.tsx

ui/src/criteria/attribute.tsx
ginaygoogle: correction --
add one more if sentence from line 78. if ===0, return <div>No attribute information</div> else return ...
Not sure whether needs warning or pop up windows.

ui/src/criteria/attribute.tsx
```diff
@@ -0,0 +1,153 @@
...
+  renderDetails() {
+    return <ConceptDetails data={this.data} />;
+  }
+
+  generateFilter() {
+    if (this.data.selected.length === 0) {
+      return null;
+    }
+
+    const operands = () =>
```
tjennison-google, line 53: There's no need for this to be a function here. Either inline it below or turn it into a variable:
```
const operands = this.data.selected.map(...);
```

ui/src/criteria/attribute.tsx
```diff
@@ -0,0 +1,153 @@
...
+    ?.attributes?.find((attribute) => attribute.name === attributeName)
+    ?.attributeFilterHint?.enumHint?.enumHintValues;
+
+  const selectionIndex = (hint: tanagra.EnumHintValue) =>
+    props.data.selected.findIndex(
+      (row) => row.id === hint.attributeValue?.int64Val
+    );
+
+  if (enumHintValues?.length === 0) {
+    return <Typography>No attribute information!</Typography>;
```
tjennison-google, line 98: It's always good to include some context in error message to make it easier for debugging. How about ```{`No information for attribute "${attributeName}".`}```?

ui/src/criteria/attribute.tsx
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  occurrenceEntities() {
+    return [];
+  }
+}
+
+function AttributeEdit(props: AttributeEditProps) {
+  const underlay = useUnderlay();
+  const attributeName = props.data.attribute;
```
tjennison-google, line 82: You don't have to change this if you like it this way but this is only a few characters shorter. Introducing a new variable makes the code more complicated since you have to go back and check what was assigned to it. It makes more sense to do that when it removes duplicated logic or makes the code notably smaller or easier to read.

ui/src/criteria/attribute.tsx
```diff
@@ -0,0 +1,153 @@
...
+    ?.attributeFilterHint?.enumHint?.enumHintValues;
+
+  const selectionIndex = (hint: tanagra.EnumHintValue) =>
+    props.data.selected.findIndex(
+      (row) => row.id === hint.attributeValue?.int64Val
+    );
+
+  if (enumHintValues?.length === 0) {
+    return <Typography>No attribute information!</Typography>;
+  } else {
```
tjennison-google, line 99: Since the `if` case always returns, there's no need to have this block in an `else` case and it adds an extra layer of nesting.

ui/src/criteria/attribute.tsx
```diff
@@ -0,0 +1,153 @@
...
+                  size="small"
+                  checked={selectionIndex(hint) > -1}
+                  onChange={() => {
+                    props.dispatchFn(
+                      produce(props.data, (data) => {
+                        if (selectionIndex(hint) > -1) {
+                          data.selected.splice(selectionIndex(hint), 1);
+                        } else {
+                          data.selected.push({
+                            id: hint.attributeValue?.int64Val as number,
```
tjennison-google, line 117: `as number` was required in concept.tsx because the table data could be multiple types. Here you don't need it because `int64Val` is always a number.

tjennison-google (changes requested):

ui/src/criteria/attribute.tsx
ginaygoogle: Thanks, taken notes. won't do it next time.

ui/src/criteria/attribute.tsx
ginaygoogle: Note!

ui/src/criteria/attribute.tsx
ginaygoogle: Good to learn us `|| -1` replace `as number`. 

Note: the `|| -1` can cover the case which is undefined 
but `as number` will hiding the undefined case will cause bug in the future and hard to figure out.

ui/src/criteria/attribute.tsx
```diff
@@ -81,52 +85,160 @@ function AttributeEdit(props: AttributeEditProps) {
...
+    attributeFilterHint?.integerBoundsHint !== "undefined"
+  ) {
+    const integerBoundsHint = attributeFilterHint?.integerBoundsHint;
+
+    if (
+      integerBoundsHint?.max !== null &&
+      typeof integerBoundsHint?.max !== "undefined" &&
+      integerBoundsHint?.min !== null &&
+      typeof integerBoundsHint?.min !== "undefined"
+    ) {
```
ginaygoogle, line 107: @tjennison-google Tim. Could you please help me think about if there is another smarter or efficient way to rewrite the 2 if sentences?

ui/src/criteria/attribute.tsx
```diff
@@ -81,52 +85,160 @@ function AttributeEdit(props: AttributeEditProps) {
...
+        const newValue =
+          event.target.value === "" ? max : Number(event.target.value);
+        updateValue([
+          Math.min(value[0], newValue),
+          Math.max(value[0], newValue),
+        ]);
+      };
+
+      return (
+        <Box sx={{ width: 300 }}>
```
ginaygoogle, line 149: @tjennison-google  I know we should not put the css there, but without the css, it will be as long as the screen.  I suppose we could not have another css file only for one box. Please advise. Thank you. Tim

ui/src/criteria/attribute.tsx
tjennison-google: There's an `isValid` function in `util/valid.tsx` that checks for undefined and null if that's what you're looking for.

ui/src/criteria/attribute.tsx
tjennison-google: Can you specify it as a percentage instead (e.g. `width: "30%", minWidth: 100`)? That way it's still prevented from resizing too small and breaking but isn't being forced to a particular size.

tjennison-google:
The age work should go into a separate PR. This PR shouldn't just keep growing while in review.

ui/src/criteria/attribute.tsx
ginaygoogle: Oh! Good catch!

ginaygoogle:
@tjennison-google  Hi Tim. I clean out the age stuffs and pr separately only for the attribute page now.

tjennison-google (approved):

PR #131, opened 2022-04-01
Fix formatting errors that previously slipped in

tjennison-google:


ui/.eslintignore
```diff
@@ -0,0 +1 @@
+src/tanagra-api/
```
ginaygoogle, line 1: Question: Is this automatic generated or manually add? What kind of things igore in this file?

ui/.prettierignore
```diff
@@ -1 +1 @@
-.gitignore
\ No newline at end of file
+.eslintignore
```
ginaygoogle, line 1: is there any reason for changing the name?

ui/package.json
```diff
@@ -53,10 +53,17 @@
     "codegen": "rm -rf src/tanagra-api && openapi-generator-cli generate -i ../api/src/main/resources/api/service_openapi.yaml -g typescript-fetch -t ./openapi-templates --additional-properties=typescriptThreePlus=true -o src/tanagra-api && sed -i 's/AttributeValue;/AttributeValue | null;/g' src/tanagra-api/models/SearchEntityInstancesResponse.ts"
   },
   "eslintConfig": {
-    "extends": ["react-app", "react-app/jest"]
+    "extends": [
```
ginaygoogle, line 56: Is there any standard for me to know when should I separate the items in different lines?

ginaygoogle (approved):

ui/.prettierignore
tjennison-google: This is a symlink and I've changed it from pointing at .gitignore to .eslintignore. It would be nice to reuse .gitignore but the paths there are relative to the root of the repo rather than the ui directory so that don't match properly. This will reuse .eslintignore for .prettierignore which will have the same paths.

ui/package.json
tjennison-google: The standard is defined by prettier. You can run `npx prettier --write` to automatically reformat everything in the repo. That should have no effect on anything you haven't changed because the repo should always be properly formatted. It may not be enforced correctly on submit yet but it will be eventually.

PR #133, opened 2022-04-07
gy/age

ginaygoogle:
This is the commit for attribute page for Age. it includes partial generateFilter (lessThan, todo: greaterThan)


ui/src/criteria/attribute.tsx
```diff
@@ -81,17 +110,115 @@ function AttributeEdit(props: AttributeEditProps) {
...
+      (attribute) => attribute.name === props.data.attribute
+    )?.attributeFilterHint;
+
+  const year = new Date().getFullYear();
+
+  if (isValid(attributeFilterHint?.integerBoundsHint)) {
+    const integerBoundsHint = attributeFilterHint?.integerBoundsHint;
+
+    if (
+      integerBoundsHint?.max !== null &&
```
ginaygoogle, line 125: This is the place does not work for isValid()

ginaygoogle:
feel confused why this branch has all previews commit history... Is it because I create this branch based on gy/subpage? But the gy/subpage is already merged...

tjennison-google:
You need to `git pull` on main and rebase these changes onto it.

ginaygoogle:
Hi Tim. I checkout to main and pull. it shows already up to date.

ginaygoogle:
I used 
- git checkout main
- git pull
- git checkout gy/age
- git rebase --onto main gy/age
- git rebase main (shows 11 commits behind)
- meet conflicts many times and git rebase --continue 
- git push --force-with-lease

ui/src/criteria/attribute.tsx
```diff
@@ -81,17 +129,115 @@ function AttributeEdit(props: AttributeEditProps) {
...
+        setMaxValue(newMax);
+        props.dispatchFn(
+          produce(props.data, (data) => {
+            data.min = newMin;
+            data.max = newMax;
+          })
+        );
+      };
+
+      React.useEffect(() => updateValue(min, max), []);
```
tjennison-google, line 166: Updating the data store as part of rendering is generally a bad idea because it results in extra, unexpected updates. This can cause issues with debugging, performance, or features like undo/redo. There are two options:

1) Allow min/max to remain undefined and treat that as meaning unbounded. Once a change is made in the UI, that value would be stored and never return to the unbounded state. `generateFilter` would need to only add operands for defined limits.
2) Set correct initial values when the criteria is created.

In my opinion, option 2 is better because it's less surprising. Somebody working on the code in the future is much less likely to break it because they don't fully understand how it works.

With that in mind, you'll need to to:
1) Move the `CriteriaConfig` definition to `underlaysSlice.ts`. That makes sense anyways since that data would someday come from the backend.
2) Add the `Underlay` as a parameter to `createCriteria` in `cohort.ts` as well as the `initializeData` parameter of `registerCriteriaPlugin`. Update the cohort and attribute criteria to accept the new parameter.
3) In the attribute plugin, lookup the attribute in `initializeData` and set the min and max accordingly if it has an integer hint.

ui/src/criteria/attribute.tsx
```diff
@@ -81,17 +129,115 @@ function AttributeEdit(props: AttributeEditProps) {
...
+
+      const handleChange = (event: Event, newValue: number | number[]) => {
```
tjennison-google, line 168: Why does `newValue` have `number | number[]` type when you immediate force it to be `number[]` on the next line?

ui/src/criteria/attribute.tsx
```diff
@@ -81,17 +129,115 @@ function AttributeEdit(props: AttributeEditProps) {
...
+        newValue = newValue as number[];
+        const [newMin, newMax] = newValue;
+        updateValue(newMin, newMax);
+      };
+
+      const handleInputChange = (
+        event: React.ChangeEvent<HTMLInputElement>
+      ) => {
+        const newMin =
```
tjennison-google, line 177: I think you need some bounds checking here to ensure that `max > min`, especially since that can happen temporarily while entering a multi-digit number. You should maybe also ensure `min >= hintMin` and `max <= hintMax`, but look at AoU and see how it handles these cases.

ui/src/criteria/attribute.tsx
```diff
@@ -81,17 +129,115 @@ function AttributeEdit(props: AttributeEditProps) {
...
+          event.target.value === "" ? min : Number(event.target.value);
+        const newMax =
+          event.target.value === "" ? max : Number(event.target.value);
+        updateValue(Math.min(newMin, minValue), Math.max(newMax, maxValue));
```
tjennison-google, line 181: What is the purpose on Math.min/.max here? Doesn't that it prevent min from ever increasing and max from ever decreasing?

ginaygoogle: I reason why I do that is there is 2 pointers. And mostly, left represents min, and right represent max. But if left goes over to the right, and then the right will be left(min), and left will be right(max). it will reverse. 
in order to avoid this corner case, I just write this

ui/src/criteria/attribute.tsx
```diff
@@ -81,17 +129,115 @@ function AttributeEdit(props: AttributeEditProps) {
...
+                  min: min,
+                  max: max,
+                  type: "number",
+                  "aria-labelledby": "input-slider",
+                }}
+              />
+            </Grid>
+            <Grid item xs>
+              <Slider
+                getAriaLabel={() => "Temperature range"}
```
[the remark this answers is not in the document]
ginaygoogle: OMG! why it is still here. I changed it to "Age Range" ...

ui/src/criteria/attribute.tsx
tjennison-google: Okay, I see what the goal is now. So what happens if min is 20, max is 30, and you start typing 40 into the max input overwriting the 30? When you enter the 4, doesn't that change the minimum to 4 and the maximum to 10? Then adding the 0 for 40 means the minimum is 4 and the maximum is 40? And the maximum input change to 10 while you're typing in it?

Beyond the issues when editing the inputs, having the inputs cross over each other feels weird to me. It's functional but unexpected and the current AoU slider doesn't allow the limits to cross either. We should emulate that behavior here.

tjennison-google (changes requested):

ui/src/criteria/attribute.tsx
ginaygoogle: because the newValue type is defined by this slider package.

ui/src/criteria/attribute.tsx
ginaygoogle: I know how to solve the problem for over sliding. There is a parameter disableSwap. 
I checked it will not change in the middle of the typing.

ui/src/criteria/attribute.tsx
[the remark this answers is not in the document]
ginaygoogle: Because we change it as "year at birth". should we call it as "year at birth range"?

ui/src/app.tsx
```diff
@@ -126,6 +126,14 @@ const criteriaConfigs = [
       attribute: "sex_at_birth_concept_id",
     },
   },
+  {
+    type: "attribute",
+    title: "Age",
```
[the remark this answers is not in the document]
ginaygoogle: agree with your idea

ui/src/criteria/attribute.tsx
```diff
@@ -138,8 +286,17 @@ type ConceptDetailsProps = {
...
     <>
-      {props.data.selected.length === 0 ? (
+      {props.data.selected.length === 0 &&
+      !isValid(props.data.min) &&
+      !isValid(props.data.max) ? (
         <Typography variant="body1">None selected</Typography>
+      ) : isValid(props.data.min) && isValid(props.data.max) ? (
+        <Stack direction="row" alignItems="baseline">
+          <Typography variant="body1">
+            Current Age in Range &nbsp; {props.data.min}&nbsp;
```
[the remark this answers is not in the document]
ginaygoogle: at beginning, i also do not have the "&nbsp"there, but it connects together like "Current Age in Range40"

ui/src/criteria/attribute.tsx
ginaygoogle: Because the slider onChange props requires a function signature `function(event: Event, value: number | Array<number>, activeThumb: number) => void` according to https://mui.com/material-ui/api/slider/#props. However, I am using a range slider which always has two value, so it can be forced to be number[] (according to the source code of range slider example in https://mui.com/material-ui/react-slider/#range-slider).

ui/src/criteria/attribute.tsx
ginaygoogle: I meditated on the input behavior for a while. I update the handleInputChange behavior and now it match the AoU behavior exactly.

ui/src/criteria/attribute.tsx
```diff
@@ -81,17 +129,115 @@ function AttributeEdit(props: AttributeEditProps) {
...
+            Age
+          </Typography>
+          <Grid container spacing={2} alignItems="center">
+            <Grid item>
+              <Input
+                value={minValue}
+                size="small"
+                onChange={handleInputChange}
+                inputProps={{
+                  step: 5,
```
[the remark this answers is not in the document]
ginaygoogle: Good idea for the dynamic interval. It will have decimal. I used (max - min) / 20 | 0 for getting the integer.

ui/src/criteria/attribute.tsx
```diff
@@ -138,8 +286,17 @@ type ConceptDetailsProps = {
 function ConceptDetails(props: ConceptDetailsProps) {
   return (
     <>
-      {props.data.selected.length === 0 ? (
+      {props.data.selected.length === 0 &&
```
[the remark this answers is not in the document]
ginaygoogle: Hi Tim, It could not correct show the content while I use the if in the typescript HTML. Maybe I should keep it here.

ginaygoogle:
Screen Shot 2022-04-15 at 1 58 39 PM
Screen Shot 2022-04-15 at 1 58 58 PM
Screen Shot 2022-04-07 at 5 25 14 PM

ui/src/criteria/attribute.tsx
```diff
@@ -1,11 +1,17 @@
...
 import Stack from "@mui/material/Stack";
 import Typography from "@mui/material/Typography";
-import { CriteriaConfig, CriteriaPlugin, registerCriteriaPlugin } from "cohort";
+import { CriteriaPlugin, registerCriteriaPlugin } from "cohort";
 import produce from "immer";
 import React from "react";
 import * as tanagra from "tanagra-api";
+import { CriteriaConfig } from "underlaysSlice";
 import { useUnderlay } from "../hooks";
+import { Underlay } from "../underlaysSlice";
```
calbach, line 13: Please combine this with L11

ui/src/criteria/attribute.tsx
```diff
@@ -1,11 +1,17 @@
...
+import { isValid } from "../util/valid";
```
calbach, line 14: Existing issue on L12 too, but please avoid parent-relative imports like this. Relative imports should only be used on imports from the same directory (./foo), otherwise they should just be relative to the application root per L11

ui/src/criteria/attribute.tsx
```diff
@@ -18,17 +24,36 @@ interface Config extends CriteriaConfig {
...
+      .find((g) => g.name === underlay.primaryEntity)
+      ?.attributes?.find(
+        (attribute) => attribute.name === data.attribute
+      )?.attributeFilterHint;
+    const integerBoundsHint = attributeFilterHint?.integerBoundsHint;
+    const minBound = integerBoundsHint?.min;
+    const maxBound = integerBoundsHint?.max;
+
+    return {
+      ...(config.plugin as Config),
```
calbach, line 50: `...data,`

ui/src/criteria/attribute.tsx
```diff
@@ -18,17 +24,36 @@ interface Config extends CriteriaConfig {
...
+  "attribute",
+  (underlay: Underlay, config: CriteriaConfig) => {
+    const data = config.plugin as Config;
+    const attributeFilterHint = underlay.entities
+      .find((g) => g.name === underlay.primaryEntity)
+      ?.attributes?.find(
+        (attribute) => attribute.name === data.attribute
+      )?.attributeFilterHint;
+    const integerBoundsHint = attributeFilterHint?.integerBoundsHint;
+    const minBound = integerBoundsHint?.min;
```
calbach, line 46: nit: I don't know that these variables are necessary, might just inline this below instead.

ui/src/criteria/attribute.tsx
```diff
@@ -18,17 +24,36 @@ interface Config extends CriteriaConfig {
 
 interface Data extends Config {
   selected: Selection[];
+  min: number | undefined;
```
calbach, line 27: These properties will not make sense for every attribute. What's the plan for supporting other attributes? Will this plugin be split out, or is the intent to multiplex different attribute modifiers into this plugin?

If it's the latter, should have a comment here around expectations for these being set, e.g. "min/max may be populated for integer value attributes"

ui/src/criteria/attribute.tsx
```diff
@@ -46,29 +71,64 @@ class _ implements CriteriaPlugin<Data> {
...
-    if (this.data.selected.length === 0) {
-      return null;
-    }
+    let operands: tanagra.Filter[] | undefined;
+    if (isValid(this.data.min) && isValid(this.data.max)) {
+      operands = [
+        {
+          binaryFilter: {
+            attributeVariable: {
+              variable: "person",
```
calbach, line 80: Is it intentional that this plugin would **only** work on the person entity and not others? It's fine if this is just not part of the current work, but probably should have a TODO here to track the current status

ui/src/criteria/attribute.tsx
```diff
@@ -46,29 +71,64 @@ class _ implements CriteriaPlugin<Data> {
...
+            },
+          },
+        },
+        {
+          binaryFilter: {
+            attributeVariable: {
+              variable: "person",
+              name: this.data.attribute,
+            },
+            operator: tanagra.BinaryFilterOperator.GreaterThan,
```
calbach, line 95: THe PR description says greatThan is not supported yet. Can you update the description?

ui/src/criteria/attribute.tsx
```diff
@@ -46,29 +71,64 @@ class _ implements CriteriaPlugin<Data> {
   }
 
   generateFilter() {
-    if (this.data.selected.length === 0) {
-      return null;
-    }
+    let operands: tanagra.Filter[] | undefined;
```
calbach, line 74: Please remove this variable, it is only used in more tightly scoped context, so there is no need to define it here. My preference would be that you just inline the operands directly in the return statements below:


```
return {
  operands: [...],
  operator: ...
};
```

If you really prefer a variable though, it should just be defined within the scope it is referenced, i.e. `const operands = [` at L75 below

ui/src/criteria/attribute.tsx
```diff
@@ -46,29 +71,64 @@ class _ implements CriteriaPlugin<Data> {
...
+      return {
+        arrayFilter: {
+          operands: operands,
+          operator: tanagra.ArrayFilterOperator.And,
         },
-        operator: tanagra.BinaryFilterOperator.Equals,
-        attributeValue: {
-          int64Val: id,
+      };
+    } else if (this.data.selected.length >= 0) {
```
calbach, line 109: This code indicates that `.selected` and `.min` +  `.max` are mutually exclusive. This should be made clear in the type documentation at L26.

ui/src/criteria/attribute.tsx
```diff
@@ -81,23 +141,149 @@ function AttributeEdit(props: AttributeEditProps) {
...
+      (attribute) => attribute.name === props.data.attribute
+    )?.attributeFilterHint;
+  const enumHintValues = attributeFilterHint?.enumHint?.enumHintValues;
+  const integerBoundsHint = attributeFilterHint?.integerBoundsHint;
 
-  const selectionIndex = (hint: tanagra.EnumHintValue) =>
-    props.data.selected.findIndex(
-      (row) => row.id === hint.attributeValue?.int64Val
-    );
+  console.log(underlay);
```
calbach, line 152: Please remove debug loggers

ui/src/criteria/attribute.tsx
```diff
@@ -81,23 +141,149 @@ function AttributeEdit(props: AttributeEditProps) {
...
+  if (!isValid(integerBoundsHint) && enumHintValues?.length === 0) {
     return (
       <Typography>
         No information for attribute {props.data.attribute}.
       </Typography>
     );
   }
+
+  if (
+    integerBoundsHint?.max !== null &&
```
calbach, line 164: `isValid(integerBoundsHint?.min) && isValid(integerBoundsHint?.max)` ?

ui/src/criteria/attribute.tsx
```diff
@@ -81,23 +141,149 @@ function AttributeEdit(props: AttributeEditProps) {
...
+    integerBoundsHint?.max !== undefined &&
+    integerBoundsHint?.min !== null &&
+    integerBoundsHint?.min !== undefined
+  ) {
+    const minBound = integerBoundsHint?.min;
+    const maxBound = integerBoundsHint?.max;
+
+    const [minInputValue, setMinInputValue] = React.useState<number>(minBound);
```
calbach, line 172: The convention in the rest of the codebase is to import `useState`. i.e. don't qualify with `React.` everywhere.

ui/src/criteria/attribute.tsx
```diff
@@ -81,23 +141,149 @@ function AttributeEdit(props: AttributeEditProps) {
...
 
-  const selectionIndex = (hint: tanagra.EnumHintValue) =>
-    props.data.selected.findIndex(
-      (row) => row.id === hint.attributeValue?.int64Val
-    );
+  console.log(underlay);
+  console.log(attributeFilterHint);
 
-  if (enumHintValues?.length === 0) {
+  if (!isValid(integerBoundsHint) && enumHintValues?.length === 0) {
```
calbach, line 155: This check should be made consistent with the check at L163. i.e. just make one check above:

```
const hasValidIntegerBounds = isValid(integerBoundsHint?.min) && isValid(integerBoundsHint?.max);
```

ui/src/criteria/attribute.tsx
```diff
@@ -81,23 +141,149 @@ function AttributeEdit(props: AttributeEditProps) {
...
     );
   }
+
+  if (
+    integerBoundsHint?.max !== null &&
+    integerBoundsHint?.max !== undefined &&
+    integerBoundsHint?.min !== null &&
+    integerBoundsHint?.min !== undefined
+  ) {
+    const minBound = integerBoundsHint?.min;
```
calbach, line 169: This is complicated enough that it's definitely worth splitting into a subcomponent. If it's not going to be used anywhere else, defining it privately in this file is fine - alternatively it could go into a components directory. 


```
function IntegerRangeSlider({...}: IntegerRangeSliderProps) {
  ...
}

...
if (hasValidIntegerBounds) {
  return <IntegerRangeSlider ...>;
}
```

ui/src/criteria/attribute.tsx
```diff
@@ -81,23 +141,149 @@ function AttributeEdit(props: AttributeEditProps) {
...
+    const maxBound = integerBoundsHint?.max;
+
+    const [minInputValue, setMinInputValue] = React.useState<number>(minBound);
+    const [maxInputValue, setMaxInputValue] = React.useState<number>(maxBound);
+    const [minValue, setMinValue] = React.useState<number>(minBound);
```
calbach, line 174: I don't understand why two sets of values are needed here (minInput, min, maxInput, max). I would check whether it's actually necessary, or if just a single minValue, maxValue would suffice.

If it is needed, please add a comment in the code explaining this.

ui/src/criteria/attribute.tsx
```diff
@@ -81,23 +141,149 @@ function AttributeEdit(props: AttributeEditProps) {
...
+    };
+
+    const handleRightInputChange = (
+      event: React.ChangeEvent<HTMLInputElement>
+    ) => {
+      let newMax =
+        event.target.value === "" ? maxBound : Number(event.target.value);
+      setMaxInputValue(newMax);
+
+      newMax = newMax < minValue ? minValue : newMax;
```
calbach, line 221: nit: I find it is usually more readable to use Math.max, Math.min over ternaries for these kinds of things.

Math.max(newMax, minValue)
Math.min(newMax, maxBound)

ui/src/criteria/attribute.tsx
```diff
@@ -81,23 +141,149 @@ function AttributeEdit(props: AttributeEditProps) {
...
+      <Box sx={{ width: "30%", minWidth: 100 }}>
+        <Grid container spacing={2} alignItems="center">
+          <Grid item>
+            <Input
+              value={minInputValue}
+              size="small"
+              onChange={handleLeftInputChange}
+              onBlur={handleLeftInputBlur}
+              inputProps={{
+                step: ((maxBound - minBound) / 20) | 0,
```
calbach, line 242: Did you mean to put `| 0` here, i.e. bitwise OR with 0? Is this to force an integer?  Math.floor() is probably more readable for this purpose

ui/src/criteria/attribute.tsx
ginaygoogle: Sure.

ui/src/criteria/attribute.tsx
ginaygoogle: Sorry for missing these

ui/src/criteria/attribute.tsx
ginaygoogle: Yes. I searched online to find a way transfer it into integer. They give me this. Nice to learn the math.floor

ui/src/criteria/attribute.tsx
ginaygoogle: This is a really strange place could not use isValid function there. or it will cause error as "TS2345: Argument of type 'number | null | undefined' is not assignable to parameter of type 'number | (() => number)'.   Type 'undefined' is not assignable to type 'number | (() => number)'. in L167-L170

I talked with Tim and he proposed to keep like that and take a look at it later and I will write a comment there to explain it.

ui/src/criteria/attribute.tsx
ginaygoogle: Thank you CH! That's an obsolete definition from version before. I improve this as you suggested.

ui/src/criteria/attribute.tsx
ginaygoogle: Thank you CH! The compiler always complain that `integerBoundsHint.min` and `integerBoundsHint.max` might be undefined. This explicit check was here to tell the compiler that min and max are valid. I improve the code by using your method and also put comments here to explain this.

ui/src/criteria/attribute.tsx
ginaygoogle: same as the issue for L163

ui/src/criteria/attribute.tsx
ginaygoogle: Yes. this is intentionally work on "person". because the other types include in the concept.tsx

ui/src/criteria/attribute.tsx
ginaygoogle: The situation is that, the values of input box and slider should be isolated. Because, the slider values will be checked within range at all time, but during the time that user enter the number into the input box, no explicit check is supposed to be performed on the input. This is the behavior in All of Us website.

For instance, our year of birth is from 1930 to 2001. A user would like to type "1940", and mistakenly type "19400". If there is checking, the "19400" will be corrected to "2001". The user will need to reenter "1940" instead of simply delete a "0" at the end. So I used two sets of values to implement this behavior as All of Us.

ui/src/criteria/attribute.tsx
ginaygoogle: Hi CH. This place was modified based on the last review comment from Tim
"I don't think these are working the way you intended. Basic data types like number and string must be defined and the compiler will complain if they're not. That's why isValid is "working" for these variables and not the hints. Instead, you need to define them as number | undefined."

ui/src/criteria/attribute.tsx
```diff
@@ -81,23 +133,145 @@ function AttributeEdit(props: AttributeEditProps) {
...
+      setMaxValue(newMax);
+      props.dispatchFn(
+        produce(props.data, (data) => {
+          data.min = newMin;
+          data.max = newMax;
+        })
+      );
+    };
+
+    const handleChange = (event: Event, newValue: number | number[]) => {
```
tjennison-google, line 176: This function has a third parameter `activeThumb` which I think will make this code easier to understand and allow better handling of the edge cases. Look at the code for this example: https://mui.com/material-ui/react-slider/#minimum-distance

ui/src/criteria/attribute.tsx
```diff
@@ -81,23 +133,145 @@ function AttributeEdit(props: AttributeEditProps) {
...
+                min: minBound,
+                max: maxBound,
+                type: "number",
+                "aria-labelledby": "input-slider",
+              }}
+            />
+          </Grid>
+          <Grid item xs>
+            <Slider
+              getAriaLabel={() => "Year at birth range"}
```
tjennison-google, line 240: This can't be hardcoded. It needs to come from the from the `CriteriaConfig`, probably `` `${config.title} Range` ``.

ui/src/criteria/attribute.tsx
tjennison-google: You should use `Math.floor` here instead of `| 0` like CH suggested.

ui/src/criteria/attribute.tsx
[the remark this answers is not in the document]
tjennison-google: I believe this is because you were initializing both `selected` and `min/max` at the same time combined with however you were writing your if statements.

In future, when you encounter something that doesn't work the way you expect you should try to understand why. There's no obvious reason why an if statement wouldn't work the same as the ternary operator so you should look at how the code is different. I would guess the difference is that you split the `selected.length` and `isValid(min/max)` parts into separate if statements and it always displayed the enum data, which makes sense given the initialization issue. You could try using the debugger or `console.log` to inspect the state.

ui/src/criteria/concept.test.tsx
```diff
@@ -138,23 +138,32 @@ beforeAll(() => {
     insertGroup(
       action.payload.id,
       GroupKind.Included,
-      createCriteria({
```
tjennison-google, line 141: As above, fixing formatting issues is good though it shouldn't be in the same PR as other changes. However, in this case I don't understand why it changed, since `npx prettier --write .` doesn't reformat it for me. Do you have something else reformatting files with different options?

ui/src/underlaysSlice.ts
```diff
@@ -17,6 +16,13 @@ export type Underlay = {
   prepackagedConceptSets: PrepackagedConceptSet[];
 };
 
+export interface CriteriaConfig {
```
tjennison-google, line 19: Please also copy the comments and formatting.

ui/src/criteria/attribute.tsx
```diff
@@ -81,23 +133,145 @@ function AttributeEdit(props: AttributeEditProps) {
...
         No information for attribute {props.data.attribute}.
       </Typography>
     );
   }
+
+  if (hasValidIntegerBounds) {
+    const minBound = integerBoundsHint?.min || 0; // This is to ensure the compiler won't complain the object might be undefined
+    const maxBound = integerBoundsHint?.max || 0; // although we already know that min and max is valid
+
+    //two sets of values are needed due to the input box and slider is isolated.
```
tjennison-google, line 159: Comments should have a space after the `//` and should be complete sentences (i.e. capitalized first letter and period at the end).

ui/src/criteria/attribute.tsx
```diff
@@ -81,23 +133,145 @@ function AttributeEdit(props: AttributeEditProps) {
...
+    const [minInputValue, setMinInputValue] = useState<number>(minBound);
+    const [maxInputValue, setMaxInputValue] = useState<number>(maxBound);
+    const [minValue, setMinValue] = useState<number>(minBound);
```
tjennison-google, line 162: These should always be exactly the same as what's stored in the `Data` (what updateValue does) so there's no need for them to have separate state. Remove these `minValue/maxValue` and read directly from the data. The input fields can have different values while editing so you do need those separate state values.

api/src/test/java/bio/terra/tanagra/service/search/SqlVisitorTest.java
```diff
@@ -133,7 +133,8 @@ void filterBinaryFunction() {
             Filter.BinaryFunction.Operator.GREATER_THAN,
             Expression.Literal.create(DataType.INT64, "62"));
     assertEquals(
-        "s.rating > 62", greaterThanFilter.accept(new SqlVisitor.FilterVisitor(getSimpleContext())));
+        "s.rating > 62",
```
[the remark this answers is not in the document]
ginaygoogle: sure. I will delete these non-related changes. 
I only run "npm run build" and "npx prettier --write ." Not sure which one caused it.... But anyway, I will delete them.

ui/src/app.tsx
```diff
@@ -126,6 +126,14 @@ const criteriaConfigs = [
       attribute: "sex_at_birth_concept_id",
     },
   },
+  {
+    type: "attribute",
+    title: "Year at Birth",
+    defaultName: "Contains Year at Birth Codes",
```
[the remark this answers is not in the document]
ginaygoogle: done

ui/src/criteria/attribute.tsx
ginaygoogle: done

ui/src/criteria/attribute.tsx
```diff
@@ -17,18 +22,35 @@ interface Config extends CriteriaConfig {
 }
 
 interface Data extends Config {
-  selected: Selection[];
+  selected: Selection[]; // selected and min/max are mutually exclusive. Only one of them will be used in an attribute.
```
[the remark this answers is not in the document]
ginaygoogle: done

ui/src/criteria/attribute.tsx
```diff
@@ -17,18 +22,35 @@ interface Config extends CriteriaConfig {
...
+    const attributeFilterHint = underlay.entities
+      .find((g) => g.name === underlay.primaryEntity)
+      ?.attributes?.find(
+        (attribute) => attribute.name === data.attribute
+      )?.attributeFilterHint;
+    const integerBoundsHint = attributeFilterHint?.integerBoundsHint;
+
+    return {
+      ...data,
+      selected: [],
```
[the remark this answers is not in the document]
ginaygoogle: nice expression in one line. done

ui/src/criteria/attribute.tsx
```diff
@@ -81,23 +133,145 @@ function AttributeEdit(props: AttributeEditProps) {
...
+  if (!hasValidIntegerBounds && enumHintValues?.length === 0) {
     return (
       <Typography>
         No information for attribute {props.data.attribute}.
       </Typography>
     );
   }
+
+  if (hasValidIntegerBounds) {
+    const minBound = integerBoundsHint?.min || 0; // This is to ensure the compiler won't complain the object might be undefined
```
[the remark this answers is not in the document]
ginaygoogle: done

ui/src/criteria/attribute.tsx
ginaygoogle: good to learn it

ui/src/criteria/attribute.tsx
```diff
@@ -81,23 +133,145 @@ function AttributeEdit(props: AttributeEditProps) {
...
+
+    const handleChange = (event: Event, newValue: number | number[]) => {
+      newValue = newValue as number[];
+      const [newMin, newMax] = newValue;
+      setMinInputValue(newMin);
+      setMaxInputValue(newMax);
+      updateValue(newMin, newMax);
+    };
+
+    const handleLeftInputChange = (
```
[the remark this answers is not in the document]
ginaygoogle: ok. I will change handleLeftInputChange to handleMinInputChange

ui/src/criteria/attribute.tsx
```diff
@@ -81,23 +133,145 @@ function AttributeEdit(props: AttributeEditProps) {
...
+    ) => {
+      let newMin =
+        event.target.value === "" ? minBound : Number(event.target.value);
+      setMinInputValue(newMin);
+
+      newMin = Math.max(minBound, newMin);
+      newMin = Math.min(maxValue, newMin);
+      updateValue(newMin, maxValue);
+    };
+    const handleLeftInputBlur = () => {
```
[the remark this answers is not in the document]
ginaygoogle: https://mui.com/material-ui/api/input-base/#props this is the definition for onBlur. it like an opposite to onFocus. 
it will be called when it unfocus. 
(I know it is tricky, but in order to solve the issue comparing while type "4"  to "40". it will need a kind of stage like onblur and ensure the number input is finished)

ui/src/criteria/attribute.tsx
ginaygoogle: sound good. it also fix the problem which the slider is always initialized to the MinBound and MaxBound.
But it will become longer due to the `props.data.min` also will be undefine.

ui/src/criteria/attribute.tsx
ginaygoogle: Hi Tim, I look at the `activeThumb` parameter. It actually tells the `handleChange` callback which slider knob the user is dragging. It will be useful if we are going to shift the range when the slider knob reach the minimum distance. However, for our case, we don't need to shift the range, but we need to make sure the slider knob won't cross each other, i.e., there is a `minimumDistance  = 0`. I already set `disableSwap` to achieve this. So I think `activeThumb` won't be useful in our case. The edge cases we are handling are from the input box actually, which is not related to the slider and `activeThumb` either. We can have a quick chat on this also.

ui/src/criteria/attribute.tsx
```diff
@@ -81,23 +133,145 @@ function AttributeEdit(props: AttributeEditProps) {
...
+      setMinInputValue(newMin);
+      setMaxInputValue(newMax);
+      updateValue(newMin, newMax);
+    };
+
+    const handleLeftInputChange = (
+      event: React.ChangeEvent<HTMLInputElement>
+    ) => {
+      let newMin =
+        event.target.value === "" ? minBound : Number(event.target.value);
```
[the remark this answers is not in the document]
ginaygoogle: Yes. That is a corner case I missed!

ui/src/criteria/attribute.tsx
tjennison-google: I thought knowing which thumb was being dragged might be helpful but if it's not that's okay.

ui/src/criteria/attribute.tsx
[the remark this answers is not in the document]
tjennison-google: I see. The `InputBase` docs weren't linked from the slider docs so I didn't see them. It makes sense to handle that case though.

ui/src/criteria/attribute.tsx
```diff
@@ -46,29 +68,59 @@ class _ implements CriteriaPlugin<Data> {
...
-          variable: "person",
-          name: this.data.attribute,
+    if (isValid(this.data.min) && isValid(this.data.max)) {
+      return {
+        arrayFilter: {
+          operands: [
+            {
+              binaryFilter: {
+                attributeVariable: {
+                  variable: "person",
```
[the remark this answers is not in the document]
ginaygoogle: got it

ui/src/criteria/attribute.tsx
```diff
@@ -81,23 +133,145 @@ function AttributeEdit(props: AttributeEditProps) {
...
+      newMax = Math.max(minValue, newMax);
+      newMax = Math.min(newMax, maxBound);
+      setMaxInputValue(newMax);
+    };
+
+    return (
+      <Box sx={{ width: "30%", minWidth: 100 }}>
+        <Grid container spacing={2} alignItems="center">
+          <Grid item>
+            <Input
```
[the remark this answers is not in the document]
ginaygoogle: Do you mean I should set a maxlength="4" ?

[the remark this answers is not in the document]
tjennison-google: I'm not familiar with the parameters. Does that make the input large enough for 4 digits?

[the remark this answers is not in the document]
tjennison-google: Actually, that sounds like it would restrict it to only 4 digits which isn't really what we want. It just needs to be wider.

ui/src/criteria/attribute.tsx
ginaygoogle: I find I made a mistake for this property "getAriaLabel" is not used to show the title/label of the slider. It required an index. Seems like we do not really need it. I just delete it

tjennison-google: getAriaLabel and ARIA properties in general are for interacting with screen readers, keyboard input, and other accessibility related things. I'm not sure exactly what issue you're running into but for now it's fine to just delete it.

ui/src/criteria/concept.test.tsx
ginaygoogle: I think this is not from `npx prettier --write .`  it caused by changing the data structure of createCriteria from L141-L147

tjennison-google: Oh, you're right. That makes sense.

ui/src/criteria/attribute.tsx
```diff
@@ -81,23 +133,145 @@ function AttributeEdit(props: AttributeEditProps) {
...
+    const handleRightInputBlur = () => {
+      let newMax = maxInputValue;
+      newMax = Math.max(minValue, newMax);
+      newMax = Math.min(newMax, maxBound);
+      setMaxInputValue(newMax);
+    };
+
+    return (
+      <Box sx={{ width: "30%", minWidth: 100 }}>
+        <Grid container spacing={2} alignItems="center">
```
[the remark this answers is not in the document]
ginaygoogle: Yes. I find It should be L221 add a margin. then it will be good. I set the margin as 2. and looks better
Screen Shot 2022-04-20 at 5 09 38 PM

ui/src/criteria/attribute.tsx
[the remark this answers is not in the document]
ginaygoogle: I think right now(the current commit after this review) will support to see 4 digits. for example, when you go to the left input (minBound), backspace until empty. you can type  9 9 9 9. it is visible to see "9999", but when your mouse move and does not focus on the input. It will automatically changed to the minBound number due to 9999 is not supported.

ui/src/underlaysSlice.ts
ginaygoogle: Done. I deleted before because I found the adjacent code all had no comments...

ui/src/criteria/attribute.tsx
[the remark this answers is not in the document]
ginaygoogle: Ok, I over complicated the problem before... I thought I need to put the if statement into the HTML part, something like:
```
return (
     <> 
          {if (isValid(XXX)) {} else {}}
     </>
)
```
which of course doesn't work. Now I put the return statement into the if statement and make it looks simpler. Thanks for your advice!

ui/src/criteria/attribute.tsx
```diff
@@ -81,73 +136,216 @@ function AttributeEdit(props: AttributeEditProps) {
...
+      const [newMin, newMax] = newValue;
+      setMinInputValue(newMin);
+      setMaxInputValue(newMax);
+      updateValue(newMin, newMax);
+    };
+
+    const handleMinInputChange = (
+      event: React.ChangeEvent<HTMLInputElement>
+    ) => {
+      let newMin = event.target.value === "" ? "" : Number(event.target.value);
```
tjennison-google, line 182: The logic here can be cleaned up:
1. Call `setMinInputValue` as the first step.
2. Make `newMin` always a number: `const newMin = minInputValue === "" ? 0 || Number(minInputValue);`
3. You're using `maxValue` in multiple places, including above in `useState`. Pull it out into a constant.
4. Splitting the min and max across multiple lines is several of unnecessary `newMin`s. Merge it into one line: `updateValue(Math.min(maxValue, Math.max(minBound, newMin)), maxValue;`

ui/src/criteria/attribute.tsx
```diff
@@ -81,73 +136,216 @@ function AttributeEdit(props: AttributeEditProps) {
...
 
-  const selectionIndex = (hint: tanagra.EnumHintValue) =>
-    props.data.selected.findIndex(
-      (row) => row.id === hint.attributeValue?.int64Val
+      newMin = Math.max(minBound, newMin);
+      newMin = Math.min(maxValue, newMin);
+      updateValue(newMin, maxValue);
+    };
+    const handleMinInputBlur = () => {
+      let newMin = typeof minInputValue === "number" ? minInputValue : 0;
```
tjennison-google, line 192: There's no need to recalculate this value since it's not changing here. `setMinInputValue(String(minValue));` is enough.

ui/src/criteria/attribute.tsx
```diff
@@ -81,73 +136,216 @@ function AttributeEdit(props: AttributeEditProps) {
...
+            <Input
+              value={minInputValue}
+              size="medium"
+              onChange={handleMinInputChange}
+              onBlur={handleMinInputBlur}
+              inputProps={{
+                step: ((maxBound - minBound) / 20) | 0,
+                min: minBound,
+                max: maxBound,
+                type: "number",
```
tjennison-google, line 232: According to the MUI docs you shouldn't use `type: "number"`. https://mui.com/material-ui/react-text-field/#type-quot-number-quot has an alternative.

ui/src/criteria/attribute.tsx
```diff
@@ -81,73 +136,216 @@ function AttributeEdit(props: AttributeEditProps) {
...
+        <Grid container spacing={2}>
+          <Grid item>
+            <Input
+              value={minInputValue}
+              size="medium"
+              onChange={handleMinInputChange}
+              onBlur={handleMinInputBlur}
+              inputProps={{
+                step: ((maxBound - minBound) / 20) | 0,
+                min: minBound,
```
tjennison-google, line 230: Do these actually work for an input? What happens if you backspace and the value goes below min? Seems like it either interferes with editing or it can't actually work so it should be removed unless I'm misunderstanding the behavior.

ui/src/criteria/attribute.tsx
```diff
@@ -81,73 +136,216 @@ function AttributeEdit(props: AttributeEditProps) {
...
+              getAriaValueText={(value) => value.toString()}
+              min={minBound}
+              max={maxBound}
+              disableSwap
+            />
+          </Grid>
+          <Grid item>
+            <Input
+              value={maxInputValue}
+              size="small"
```
tjennison-google, line 251: The other `Input` is `size="medium"` so it's either doing nothing and should be removed or they should be the same.

ui/src/criteria/attribute.tsx
```diff
@@ -81,73 +136,216 @@ function AttributeEdit(props: AttributeEditProps) {
...
+              onChange={handleMaxInputChange}
+              onBlur={handleMaxInputBlur}
+              inputProps={{
+                step: Math.floor((maxBound - minBound) / 20),
```
tjennison-google, line 255: I think this won't do anything once the input type isn't "number" anymore but you should check.

ui/src/criteria/attribute.tsx
```diff
@@ -81,73 +136,216 @@ function AttributeEdit(props: AttributeEditProps) {
...
-  return (
-    <>
-      {props.data.selected.length === 0 ? (
-        <Typography variant="body1">None selected</Typography>
-      ) : (
-        props.data.selected.map(({ id, name }) => (
+function AttributeDetails(props: AttributeDetailsProps) {
+  if (
+    typeof props.data.selected !== "undefined" &&
+    props.data.selected.length !== 0
```
tjennison-google, line 321: `if (props.data.selected?.length)` will cover both cases here and I think satisfy the compiler.

ui/src/criteria/attribute.tsx
```diff
@@ -81,73 +136,216 @@ function AttributeEdit(props: AttributeEditProps) {
...
-      )}
+        ))}
+      </>
+    );
+  } else if (isValid(props.data.min) && isValid(props.data.max)) {
+    return (
+      <>
+        <Stack direction="row" alignItems="baseline">
+          <Typography variant="body1">
+            Current {props.data.attribute} in Range {props.data.min} to{" "}
```
tjennison-google, line 338: This should be the name instead of the attribute. For now, in the `registerCriteriaPlugin`  function store the name from the config into a new variable in data. Longer term it might make sense to pull in the name more generically.

tjennison-google (changes requested):

ui/src/criteria/attribute.tsx
```diff
@@ -81,73 +136,216 @@ function AttributeEdit(props: AttributeEditProps) {
...
+
+  const hasValidIntegerBounds =
+    isValid(integerBoundsHint?.min) && isValid(integerBoundsHint?.max);
+
+  if (hasValidIntegerBounds) {
+    const minBound = integerBoundsHint?.min || 0; // This is to ensure the compiler won't complain the object might be undefined.
+    const maxBound = integerBoundsHint?.max || 0; // although we already know that min and max is valid.
+
+    // two sets of values are needed due to the input box and slider is isolated.
+    const [minInputValue, setMinInputValue] = useState<number | string>(
```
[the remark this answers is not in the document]
ginaygoogle: Hi @tjennison-google I found it does not work. Due to the requirement on https://mui.com/material-ui/react-slider/#slider-with-input-field input Slider could not allowed only string. 

I followed with your suggestion. and the button in the input blank (i mean up, and down) are disappear... like this:
Screen Shot 2022-04-26 at 3 43 22 PM

ui/src/criteria/attribute.tsx
```diff
@@ -81,73 +136,216 @@ function AttributeEdit(props: AttributeEditProps) {
...
     .find((g) => g.name === underlay.primaryEntity)
-    ?.attributes?.find((attribute) => attribute.name === props.data.attribute)
-    ?.attributeFilterHint?.enumHint?.enumHintValues;
+    ?.attributes?.find(
+      (attribute) => attribute.name === props.data.attribute
+    )?.attributeFilterHint;
+  const enumHintValues = attributeFilterHint?.enumHint?.enumHintValues;
+  const integerBoundsHint = attributeFilterHint?.integerBoundsHint;
+
+  const hasValidIntegerBounds =
```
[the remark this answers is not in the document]
ginaygoogle: I understand it. I would like to clarify what's the standard we should consider to separate a variable.
like use more than 2 times (include 2 times?),
expression long?

ui/src/criteria/attribute.tsx
```diff
@@ -81,73 +136,216 @@ function AttributeEdit(props: AttributeEditProps) {
...
+    isValid(integerBoundsHint?.min) && isValid(integerBoundsHint?.max);
+
+  if (hasValidIntegerBounds) {
+    const minBound = integerBoundsHint?.min || 0; // This is to ensure the compiler won't complain the object might be undefined.
```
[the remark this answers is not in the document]
ginaygoogle: done

ui/src/criteria/attribute.tsx
```diff
@@ -81,73 +136,216 @@ function AttributeEdit(props: AttributeEditProps) {
...
+    const maxBound = integerBoundsHint?.max || 0; // although we already know that min and max is valid.
+
+    // two sets of values are needed due to the input box and slider is isolated.
```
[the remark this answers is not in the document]
ginaygoogle: done

ui/src/criteria/attribute.tsx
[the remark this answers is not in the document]
ginaygoogle: done

ui/src/criteria/attribute.tsx
```diff
@@ -81,73 +136,216 @@ function AttributeEdit(props: AttributeEditProps) {
...
+        produce(props.data, (data) => {
+          data.min = newMin;
+          data.max = newMax;
+        })
+      );
+    };
+
+    const handleChange = (event: Event, newValue: number | number[]) => {
+      newValue = newValue as number[];
+      const [newMin, newMax] = newValue;
```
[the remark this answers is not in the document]
ginaygoogle: done

ui/src/criteria/attribute.tsx
ginaygoogle: done

ui/src/criteria/attribute.tsx
ginaygoogle: Hi @tjennison-google this does not work really well as our expectation. like the picture shows the left side did not have button on the input blank. 
Screen Shot 2022-04-26 at 4 45 37 PM

ui/src/criteria/attribute.tsx
ginaygoogle: I delete this line

ui/src/criteria/attribute.tsx
ginaygoogle: done

ui/src/criteria/attribute.tsx
```diff
@@ -17,18 +22,38 @@ interface Config extends CriteriaConfig {
...
+@registerCriteriaPlugin(
+  "attribute",
+  (underlay: Underlay, config: CriteriaConfig) => {
+    const data = config.plugin as Config;
+    const attributeFilterHint = underlay.entities
+      .find((g) => g.name === underlay.primaryEntity)
+      ?.attributes?.find(
+        (attribute) => attribute.name === data.attribute
+      )?.attributeFilterHint;
+    const integerBoundsHint = attributeFilterHint?.integerBoundsHint;
```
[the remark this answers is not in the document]
ginaygoogle: done

ui/src/criteria/attribute.tsx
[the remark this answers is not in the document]
ginaygoogle: (my misunderstanding) done now

ui/src/criteria/attribute.tsx
ginaygoogle: done

ui/src/criteria/attribute.tsx
ginaygoogle: Screen Shot 2022-04-26 at 4 53 52 PM
this is I backspace, and the mouse focus on the input number.
Screen Shot 2022-04-26 at 4 52 27 PM
after the mouse click other place and it does not focus on the input blank. it will automatically go to the number which we support(in this case, it is the min number 1930)

ui/src/criteria/attribute.tsx
ginaygoogle: I think it works. there is no decimal.

ui/src/criteria/attribute.tsx
ginaygoogle: done

ui/src/criteria/attribute.tsx
ginaygoogle: Hi Tim, I just double checked. The `min` and `max` props here actually can bound the range of input box during pressing the up or down button at the side (or simply up or down key on keyboard). Like the following screenshot. Without these two props, the input box can either be increased or decreased without bound until the user click somewhere else, which trigger the `onBlur` callback. I think keeping these two bounds here will be better.
Screen Shot 2022-04-28 at 12 18 46 AM

tjennison-google: Okay, good to leave them in then.

ui/src/criteria/attribute.tsx
```diff
@@ -81,73 +136,200 @@ function AttributeEdit(props: AttributeEditProps) {
...
-      )}
+        ))}
+      </>
+    );
+  } else if (isValid(props.data.min) && isValid(props.data.max)) {
+    return (
+      <>
+        <Stack direction="row" alignItems="baseline">
+          <Typography variant="body1">
+            Current {props.data.name || ""} in Range {props.data.min} to{" "}
```
tjennison-google, line 322: Can you change name to a non-optional parameter so you don't need the `|| ""`?

tjennison-google (approved):

ui/src/criteria/attribute.tsx
ginaygoogle: done

PR #134, opened 2022-04-08
TNG-17: rename the criteria

ginaygoogle:


ui/src/overview.tsx
```diff
@@ -129,11 +139,89 @@ function ParticipantsGroup(props: { group: Group }) {
...
+    setName(event.target.value);
+  };
+
+  const onCreate = () => {
+    setOpen(false);
+    props.callback(name);
+  };
+
+  return [
+    // eslint-disable-next-line react/jsx-key
```
ginaygoogle, line 166: This is copy from the <Dialog> from dataset.tsx file

ginaygoogle: Hi Tim. While I am doing it. I have a question. Should I modify the dataset.tsx file's <dialog> also use the new created dialog module in the same branch? Because I remember, last time you said we should make sure the commit be clean and only contain the stuffs related to the same feature. But I think if I do not do this here, we will need create another branch/pr individually for the dataset.tsx file. It might be too light changes?

tjennison-google: That is a good thought. There are no hard rules here though. It's a question of balancing the overhead of multiple PRs with the ability to examine/revert/merge previous changes and the difficulty of reviewing them. If this change was really large I would suggest separating the `dataset.tsx` changes to make it easier to review. On the other hand, fixing several unrelated typos in the same PR wouldn't be unreasonable. Sharing the code would be a significant part of this change so I would leave them together. You could certainly make an argument that it's better to change the existing `dataset.tsx` usage in a separate PR though. Having small PRs isn't a problem as long as they're a complete chunk of work and it's better than combining unrelated changes.

ui/src/components/dialog.tsx
```diff
@@ -0,0 +1,70 @@
...
+    // eslint-disable-next-line react/jsx-key
+    <Dialog
+      open={open}
+      onClose={() => {
+        setOpen(false);
+      }}
+      aria-labelledby={props.titleId}
+      maxWidth="sm"
+      fullWidth
+      className={props.className}
```
tjennison-google, line 43: This also doesn't need to be a property.

ui/src/components/dialog.tsx
```diff
@@ -0,0 +1,70 @@
+import Button from "@mui/material/Button";
```
tjennison-google, line 1: Change this file to `renameDialog.tsx` so it's clear what it does.

ui/src/overview.tsx
```diff
@@ -134,6 +140,23 @@ function ParticipantCriteria(props: { group: Group; criteria: Criteria }) {
   const cohort = useCohort();
   const dispatch = useAppDispatch();
 
+  const [dialog, showRenameCriteria] = useDialog({
```
tjennison-google, line 143: Rename these to `renameDialog, showRenameDialog` so it's more obvious they're related.

tjennison-google (changes requested):

ui/src/components/dialog.tsx
ginaygoogle: I understand you. I have a concern that if we change it to renameDialog.tsx How does the other dialog use it? it will be not generic for other dialog like the dialog in dataset.tsx. Because that is not doing rename work.

tjennison-google: That's fair. It could be used for more than just names. How about textInputDialog?

ginaygoogle: OK. good idea.

ui/src/components/dialog.tsx
ginaygoogle: I change is to className = "textInput-dialog-name"

Due to we do not call it rename dialog.

ui/src/components/dialog.tsx
```diff
@@ -0,0 +1,70 @@
...
+  };
+
+  return [
+    // eslint-disable-next-line react/jsx-key
+    <Dialog
+      open={open}
+      onClose={() => {
+        setOpen(false);
+      }}
+      aria-labelledby={props.titleId}
```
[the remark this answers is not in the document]
ginaygoogle: I change is to className = "textInput-dialog-title"

Due to we do not call it rename dialog.

ui/src/components/dialog.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+type NewDialogProps = {
+  title: string;
+  titleId: string;
+  textLabel: string;
+  className: string;
+  callback: (name: string) => void;
+};
+
+export function useDialog(props: NewDialogProps): [ReactNode, () => void] {
```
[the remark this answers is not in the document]
ginaygoogle: useTextInputDialog

ui/src/components/dialog.tsx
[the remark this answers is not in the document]
tjennison-google: This name is good but is using inconsistent capitalization. It should be "text-input-dialog-title" instead.

ui/src/overview.tsx
ginaygoogle: I correct it in this commit while correct the comments

ui/src/components/dialog.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const show = () => {
+    setOpen(true);
+  };
+
+  const [name, setName] = useState(props.title);
+  const onNameChange = (event: ChangeEvent<HTMLInputElement>) => {
+    setName(event.target.value);
+  };
+
+  const onCreate = () => {
```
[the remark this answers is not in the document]
ginaygoogle: I also proposed it should be property for the hint at beginning. But then, I think the text input type button hint will mostly be "create/confirm".  largely possibility not change.
Quick question: Do you think the rename dialog should be "confirm" and the dataset.tsx dialog should be "create"?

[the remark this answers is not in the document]
tjennison-google: Given that it's a generic dialog the functions and variables should have generic names rather than create or rename that are specific to certain usages. I'm not sure what you mean about the button mostly being create/confirm. That's the two use cases we have but there isn't a reason to limit it to those options.

I would use "Rename" and "Create" for the buttons. Confirm is a more generic name to use in the code.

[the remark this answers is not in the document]
ginaygoogle: I understand your idea and agree with you. Thanks for letting me know the graininess for a property.

ui/src/overview.tsx
```diff
@@ -150,6 +174,9 @@ function ParticipantCriteria(props: { group: Group; criteria: Criteria }) {
       >
         Delete Criteria
       </MenuItem>,
+      <MenuItem key="2" onClick={showRenameCriteria}>
```
ginaygoogle, line 177: not sure whether we should change the key = "2" to some string name like  "rename_criteria". but the reason why I write 2 is because the above is 1.

ginaygoogle, line 177: not sure whether we should change the key = "2" to some string name like  "rename_criteria". but the reason why I write 2 is because the above is 1.

ui/src/components/dialog.tsx
tjennison-google: It doesn't look like you actually renamed the file. It should be `textInputDialog.tsx`.

ui/src/components/dialog.tsx
```diff
@@ -0,0 +1,73 @@
+import Button from "@mui/material/Button";
+import Dialog from "@mui/material/Dialog";
+import DialogActions from "@mui/material/DialogActions";
+import DialogContent from "@mui/material/DialogContent";
+import DialogTitle from "@mui/material/DialogTitle";
+import TextField from "@mui/material/TextField";
+import { ChangeEvent, ReactNode, useState } from "react";
+
+type NewDialogProps = {
```
tjennison-google, line 9: `TextInputDialogProps`

ui/src/components/dialog.tsx
```diff
@@ -0,0 +1,73 @@
...
+  title: string;
+  titleId: string;
```
tjennison-google, line 11: Remove, this should be unused now.

ui/src/components/dialog.tsx
```diff
@@ -0,0 +1,73 @@
...
+  textLabel: string;
+  className: string;
```
tjennison-google, line 13: Remove, this should be unused now.

ui/src/components/dialog.tsx
```diff
@@ -0,0 +1,73 @@
...
+      open={open}
+      onClose={() => {
+        setOpen(false);
+      }}
+      aria-labelledby="text-input-dialog-title"
+      maxWidth="sm"
+      fullWidth
+      className="text-input-dialog-name"
+    >
+      <DialogTitle id={props.titleId}>{props.title}</DialogTitle>
```
tjennison-google, line 48: `props.titleId` should be replaced by `text-input-dialog-title` to match the `aria-labelledby` above like it did before.

ui/src/components/dialog.tsx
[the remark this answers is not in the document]
tjennison-google: It doesn't look like Create was renamed to Confirm.

ui/src/overview.tsx
tjennison-google: It could be `rename-criteria` but it doesn't matter for a fixed list like this. Those keys are used to determine which items have been added or removed when the list changes.

tjennison-google (changes requested):

ui/src/components/dialog.tsx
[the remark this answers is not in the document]
ginaygoogle: Oh. I think i might misunderstand the comment. I change the buttonhint as a property then it can transfer, and we can set the button shows whatever we want. like create ,or  Confirm. Because this is a generic module. I just use the onCreate as the function name. Do you mean I should change the name from "onCreate" to "onConfirm" ?

ui/src/overview.tsx
ginaygoogle: Ok. Keep it consistant with 1.

ui/src/components/dialog.tsx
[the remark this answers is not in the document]
ginaygoogle: Maybe we can call it as onClick? It will be more generic.

ginaygoogle:
Screen Shot 2022-04-08 at 1 48 52 AM
Screen Shot 2022-04-08 at 1 48 58 AM
Screen Shot 2022-04-08 at 1 49 14 AM
Screen Shot 2022-04-08 at 1 49 21 AM

calbach (commented):
Sorry, I realize now several of my comments are on existing code that you were moving. If changes pertaining to existing code issues are easy to make, feel free to do so - otherwise we can try to address in a follow-up.

ui/src/components/textInputDialog.tsx
```diff
@@ -0,0 +1,72 @@
...
+import DialogTitle from "@mui/material/DialogTitle";
+import TextField from "@mui/material/TextField";
+import { ChangeEvent, ReactNode, useState } from "react";
+
+type textInputDialogProps = {
+  title: string;
+  titleId: string;
+  textLabel: string;
+  buttonHint: string;
+  callback: (name: string) => void;
```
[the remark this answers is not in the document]
ginaygoogle: Good catch. Let me use onConfirm.

ui/src/components/textInputDialog.tsx
```diff
@@ -0,0 +1,72 @@
...
+};
+
+export function useTextInputDialog(
+  props: textInputDialogProps
+): [ReactNode, () => void] {
+  const [open, setOpen] = useState(false);
+  const show = () => {
```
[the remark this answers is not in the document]
ginaygoogle: Yes. Agree.

ui/src/components/textInputDialog.tsx
```diff
@@ -0,0 +1,72 @@
...
+  title: string;
+  titleId: string;
+  textLabel: string;
+  buttonHint: string;
+  callback: (name: string) => void;
+};
+
+export function useTextInputDialog(
+  props: textInputDialogProps
+): [ReactNode, () => void] {
```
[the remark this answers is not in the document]
ginaygoogle: I explain "// Return a dialog and the callback function to show the dialog"

ui/src/cohortsSlice.ts
```diff
@@ -90,6 +90,30 @@ const cohortsSlice = createSlice({
...
+    renameCriteria: (
+      state,
+      action: PayloadAction<{
+        cohortId: string;
+        groupId: string;
+        criteriaId: string;
+        criteriaName: string;
+      }>
+    ) => {
+      const cohort = state.find((c) => c.id === action.payload.cohortId);
```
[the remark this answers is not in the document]
ginaygoogle: Hi CH. This style is which I copy from the above style which wants to make it consistent.

ui/src/components/textInputDialog.tsx
```diff
@@ -0,0 +1,71 @@
...
+  titleId: string;
+  textLabel: string;
+  buttonLabel: string;
+  onConfirm: (name: string) => void;
+};
+
+export function useTextInputDialog(
+  props: textInputDialogProps
+): [ReactNode, () => void] {
+  // Return a dialog and the callback function to show the dialog
```
calbach, line 20: I would place this comment above the method instead, on line 17.

calbach (approved):

ui/src/components/textInputDialog.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const [open, setOpen] = useState(false);
+  const show = () => setOpen(true);
+
+  const [name, setName] = useState(props.title);
+  const onNameChange = (event: ChangeEvent<HTMLInputElement>) => {
```
tjennison-google, line 25: onTextChange

ui/src/components/textInputDialog.tsx
```diff
@@ -0,0 +1,71 @@
...
+};
+
+export function useTextInputDialog(
+  props: textInputDialogProps
+): [ReactNode, () => void] {
+  // Return a dialog and the callback function to show the dialog
+  const [open, setOpen] = useState(false);
+  const show = () => setOpen(true);
+
+  const [name, setName] = useState(props.title);
```
tjennison-google, line 24: [text, setText]

ui/src/components/textInputDialog.tsx
```diff
@@ -0,0 +1,71 @@
...
+      maxWidth="sm"
+      fullWidth
+      className="text-input-dialog-name"
+    >
+      <DialogTitle id="text-input-dialog-title">{props.title}</DialogTitle>
+      <DialogContent>
+        <TextField
+          autoFocus
+          margin="dense"
+          id="name"
```
tjennison-google, line 51: id="text"

ui/src/components/textInputDialog.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+type textInputDialogProps = {
+  title: string;
+  titleId: string;
+  textLabel: string;
+  buttonLabel: string;
+  onConfirm: (name: string) => void;
+};
+
+// Return a dialog and the callback function to show the dialog
```
tjennison-google, line 17: This should end with a period.

ui/src/components/textInputDialog.tsx
```diff
@@ -0,0 +1,71 @@
...
+import Dialog from "@mui/material/Dialog";
+import DialogActions from "@mui/material/DialogActions";
+import DialogContent from "@mui/material/DialogContent";
+import DialogTitle from "@mui/material/DialogTitle";
+import TextField from "@mui/material/TextField";
+import { ChangeEvent, ReactNode, useState } from "react";
+
+type textInputDialogProps = {
+  title: string;
+  titleId: string;
```
tjennison-google, line 11: I'm not sure if my previous comment got lost somewhere, but when you resolved it you didn't actually fix it. `titleId` is unused and should be removed.

tjennison-google (approved):

PR #135, opened 2022-04-13
regeneration from M1 local

ginaygoogle:


tjennison-google:
The commit message isn't very descriptive. Use something that more specifically describes what you've actually done like:
```
Fix backend build on M1 macs

Update protoc version and write package locks.
```

tjennison-google (approved):

PR #141, opened 2022-04-22
TNG-19: add group UI and add edit group name

ginaygoogle:


ui/src/overview.tsx
```diff
@@ -118,8 +126,47 @@ function AddCriteriaButton(props: { group: string | GroupKind }) {
 }
 
 function ParticipantsGroup(props: { group: Group }) {
+  const dispatch = useAppDispatch();
+  const cohort = useCohort();
+  const [renameGroupDialog, showRenameGroup] = useTextInputDialog({
```
ginaygoogle, line 131: @tjennison-google  I found a bug in TNG-17 rename dialog. there are 3 parameters in the useTextInputDialog. But we need another parameter for the text. or it will cause the dialog default show "title" at the first time when we modify the name. Like this: 
Screen Shot 2022-04-21 at 10 43 51 PM

But this is only for the first time, next time, it will show the last changed name.
Screen Shot 2022-04-21 at 10 48 40 PM

 I checked All of Us page. It shows empty in the textInputDialog at the first time to modify the name.  
Screen Shot 2022-04-21 at 10 45 13 PM
 
Should I use the default empty blank that is consistent with All of us? 
--yes, I will create a patch PR for it
--no, should we show the actual name in the blank? like shows "Group1" while it modify at the first time. 

I also notice there is a red warning if I enter only spaces. I think I also need to add this corner case in the patch PR.
Screen Shot 2022-04-21 at 11 02 18 PM

ui/src/overview.tsx
```diff
@@ -81,7 +82,14 @@ function AddCriteriaButton(props: { group: string | GroupKind }) {
       groupId = props.group;
       dispatch(insertCriteria({ cohortId: cohort.id, groupId, criteria }));
     } else {
-      const action = dispatch(insertGroup(cohort.id, props.group, criteria));
+      const action = dispatch(
+        insertGroup(
+          cohort.id,
+          "Group" + String(cohort.groups.length + 1),
```
ginaygoogle, line 88: not sure whether this is a good way. my concern from 2 sides. one is Do you think this is long-- to create a variable. But this is only used once. The other hand is seems like a little hard code?

ui/src/app.css
```diff
@@ -38,6 +38,12 @@
   padding: 0.5em 0em;
 }
 
+.overview .group-title {
+  padding: 0.5em 0em;
+  background: #e2e6ea;
```
ginaygoogle, line 43: these colors are copied from All of us page

ui/src/overview.tsx
tjennison-google: Don't worry about the input box starting blank. AoU has the current name in the title instead but, in my opinion at least, it's better to have the current name editable.

Also don't worry about the red error. The confirm button is disabled when the input box is empty which is the most important part. It certainly wouldn't hurt to add it but that can easily go into a separate PR, though we would use MUI's input error UI which changes the color of the label and border or something like that.

ui/src/overview.tsx
tjennison-google: The AoU UI is a bit more subtle than this. It leaves the names blank at creation, and displays "Group X" as a fallback in that case. What that accomplishes is that when you have two groups and delete "Group 1", an unnamed "Group 2" will instead revert to displaying "Group 1". If you set the name at creation time then "Group 2" will stay the same and adding a new group will add a second "Group 2". We should use the same behavior.

As for the second question, what would you propose instead of creating the name like this? Unless we had an internationalization setup to make it work in different languages I'm not sure where would be a better place to put it.

ginaygoogle:
This is what the edit group name showing from create a group default as group + index. and when modify the name, the current name will show in the dialog blank. 
Screen Shot 2022-04-25 at 5 10 30 PM
Screen Shot 2022-04-25 at 5 10 43 PM
Screen Shot 2022-04-25 at 5 10 52 PM
Screen Shot 2022-04-25 at 5 11 03 PM
g)

ui/src/cohortsSlice.ts
```diff
@@ -36,18 +36,43 @@ const cohortsSlice = createSlice({
           cohort.groups.push(action.payload.group);
         }
       },
-      prepare: (cohortId: string, kind: GroupKind, criteria: Criteria) => ({
+      prepare: (
+        cohortId: string,
+        name: string,
```
tjennison-google, line 41: No need to add this parameter since it always starts empty.

ui/src/cohort.ts
```diff
@@ -39,6 +39,7 @@ export enum GroupKind {
 
 export interface Group {
   id: string;
+  name: string;
```
tjennison-google, line 42: Make this optional so it doesn't have to be filled since it always starts empty.

ui/src/overview.tsx
```diff
@@ -117,9 +120,54 @@ function AddCriteriaButton(props: { group: string | GroupKind }) {
...
+  const dispatch = useAppDispatch();
+  const cohort = useCohort();
+  const [renameGroupDialog, showRenameGroup] = useTextInputDialog({
+    title: "Edit Group Name",
+    initialText:
+      props.group.name === ""
+        ? "Group " + String(props.index + 1)
+        : props.group.name,
+    textLabel: "Group Name",
+    buttonLabel: "Confirm",
```
tjennison-google, line 133: `"Rename Group"`

The UI is easier to understand if we use more specific wording. That's why we added this as a parameter in the first place.

ui/src/overview.tsx
```diff
@@ -117,9 +120,54 @@ function AddCriteriaButton(props: { group: string | GroupKind }) {
...
 }
 
-function ParticipantsGroup(props: { group: Group }) {
+function ParticipantsGroup(props: { group: Group; index: number }) {
+  const dispatch = useAppDispatch();
+  const cohort = useCohort();
+  const [renameGroupDialog, showRenameGroup] = useTextInputDialog({
+    title: "Edit Group Name",
+    initialText:
+      props.group.name === ""
```
tjennison-google, line 129: Pull this entire expression out into a variable to ensure it stays consistent:
```
const groupName = props.group.name || "Group " + String(props.index + 1);
```

ui/src/overview.tsx
```diff
@@ -117,9 +120,54 @@ function AddCriteriaButton(props: { group: string | GroupKind }) {
...
+      <Grid container className="group-title">
+        <Grid item xs="auto">
+          <IconButton onClick={groupShow} component="span" size="small">
+            <MoreVertIcon fontSize="small" />
+          </IconButton>
+          {groupMenu}
+          {renameGroupDialog}
+        </Grid>
+        <Grid item>
+          <Typography variant="h5" fontWeight="bolder">
```
tjennison-google, line 164: Remove `fontWeight`, even though it may look better. We want to keep formatting out of the code as much as possible so it's easier to customize it for different use cases.

ui/src/cohort.ts
ginaygoogle: good catch

ui/src/cohortsSlice.ts
ginaygoogle: done

ui/src/overview.tsx
ginaygoogle: done

ui/src/overview.tsx
ginaygoogle: Agree. yesterday, I also found there are twice this expression. But i do not have a good idea to pass the variable from front to back. that's good

ui/src/cohortsSlice.ts
```diff
@@ -41,13 +41,33 @@ const cohortsSlice = createSlice({
           cohortId,
           group: {
             id: generateId(),
+            name: "",
```
tjennison-google, line 44: No need to fill in a default value now that it's optional.

tjennison-google (approved):

ui/src/cohortsSlice.ts
ginaygoogle: omg.... i just merged and have not noticed this comment. What should I do now....

PR #142, opened 2022-04-22
Patch/Correct the textInputDialog

ginaygoogle:
add text in textInputDialogProps, and make the dialog blank shows the current name

### DIFF
--- a/ui/cypress/integration/tests.js
+++ b/ui/cypress/integration/tests.js
@@ -6,6 +6,7 @@ describe("Basic tests", () => {
     cy.contains("Datasets");
 
     cy.get("button[id=insert-cohort]").click();
+    cy.get("#text").type("New Cohort");
     cy.get("button:Contains(Create)").click();
     cy.get("button:Contains(Add Criteria)").click();
     cy.get("li:Contains(Conditions)").click();
--- a/ui/src/components/textInputDialog.tsx
+++ b/ui/src/components/textInputDialog.tsx
@@ -8,6 +8,7 @@ import { ChangeEvent, ReactNode, useState } from "react";
 
 type textInputDialogProps = {
   title: string;
+  text?: string;
   textLabel: string;
   buttonLabel: string;
   onConfirm: (name: string) => void;
@@ -20,7 +21,9 @@ export function useTextInputDialog(
   const [open, setOpen] = useState(false);
   const show = () => setOpen(true);
 
-  const [text, setText] = useState(props.title);
+  const [text, setText] = useState(
+    typeof props.text !== "undefined" ? props.text : props.title
+  );
   const onTextChange = (event: ChangeEvent<HTMLInputElement>) => {
     setText(event.target.value);
   };
--- a/ui/src/components/textInputDialog.tsx
+++ b/ui/src/components/textInputDialog.tsx
@@ -8,7 +8,7 @@ import { ChangeEvent, ReactNode, useState } from "react";
 
 type TextInputDialogProps = {
   title: string;
-  text?: string;
+  initialText?: string;
   textLabel: string;
   buttonLabel: string;
   onConfirm: (name: string) => void;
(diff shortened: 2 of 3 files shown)


ui/src/components/textInputDialog.tsx
```diff
@@ -20,7 +21,9 @@ export function useTextInputDialog(
   const [open, setOpen] = useState(false);
   const show = () => setOpen(true);
 
-  const [text, setText] = useState(props.title);
+  const [text, setText] = useState(
+    typeof props.text !== "undefined" ? props.text : props.title
```
tjennison-google, line 25: Putting the title in the text box doesn't make sense since the title would be a weird thing to use as a name (e.g. "Edit Criteria Name", though I know that wouldn't happen in this case). Instead use `props.text || ""`.

ui/src/components/textInputDialog.tsx
```diff
@@ -8,6 +8,7 @@ import { ChangeEvent, ReactNode, useState } from "react";
 
 type textInputDialogProps = {
   title: string;
+  text?: string;
```
tjennison-google, line 11: How about `initialText` to make it clear that it only applies when the dialog first appears?

ui/src/components/textInputDialog.tsx
```diff
@@ -8,6 +8,7 @@ import { ChangeEvent, ReactNode, useState } from "react";
 
 type textInputDialogProps = {
```
tjennison-google, line 9: Sorry, I missed this on the previous PR. The first letter of `textInputDialogProps` should be capitalized since it's a type, just like components (so `TextInputDialogProps`). Functions and variables start with a lower case letter. Please make another quick PR to rename it.

tjennison-google (commented):
The PR title should be clearer about what has actually changed. You clarified it well in the description but it's better to have more information in the title since you can't see the description in a list of commits without clicking into it. I would go with `Add initialText parameter to TextInputDialog`, just a shortened version of what you put in the description.

ui/src/components/textInputDialog.tsx
ginaygoogle: Done in gy/dialogName this branch.

ui/src/components/textInputDialog.tsx
ginaygoogle: Yes. It think you are right. Only the initial text

ui/src/components/textInputDialog.tsx
tjennison-google: What more specifically was the problem? I would have thought that `props.text || ""` would have the same behavior as currently: empty for new cohorts and existing text for renaming criteria. Alternatively, if the integration test is expecting the text box to contain the title then the test should be fixed.

ginaygoogle: Yes... my fault

ui/src/components/textInputDialog.tsx
```diff
@@ -20,7 +21,9 @@ export function useTextInputDialog(
   const [open, setOpen] = useState(false);
   const show = () => setOpen(true);
 
-  const [text, setText] = useState(props.title);
+  const [text, setText] = useState(
+    (props.initialText = props.initialText || "")
```
tjennison-google, line 25: Why are you assigning to `props.initialText`? I don't see why it's necessary but even if it was it's confusing to do it inside of an expression.

ginaygoogle: OMG!!! My fault, so stupid am I...

ui/cypress/integration/tests.js
```diff
@@ -6,6 +6,7 @@ describe("Basic tests", () => {
     cy.contains("Datasets");
 
     cy.get("button[id=insert-cohort]").click();
+    cy.get("#text").type("New Cohort");
```
tjennison-google, line 9: I just noticed this doesn't have a `.click()` on it. What is it actually doing?

ginaygoogle: Because we change the text blank as empty while we do not assign any initialText. Like create a new Cohort. 
So we must type the name "New Cohort" in the dialog. on the line 10 it has `.click()` to create the new cohort.

tjennison-google: Oh, I see. I was thinking it meant like the element type or something but it actually types those characters in.

tjennison-google (approved):

PR #145, opened 2022-04-27
Update project for aou_synthetic underlay

tjennison-google:


api/src/main/resources/underlays/aou_synthetic.yaml
```diff
@@ -340,7 +340,7 @@ relationships:
 datasets:
   - name: aou_synthetic
     bigQueryDataset:
-      projectId: broad-tanagra-dev
+      projectId: verily-tanagra-dev
```
calbach, line 343: I suspect you'll need to regenerate the SQL goldenfile tests, I see a bunch of references

calbach (approved):

api/src/main/resources/underlays/aou_synthetic.yaml
tjennison-google: Yep, done.

PR #146, opened 2022-05-06
Add underlay for SD data

tjennison-google:
This is almost completely copied (with search/replace) from for the AoU underlay except for the concept_synonym table that isn't part of the SD dataset.

### DIFF
--- a/api/src/main/resources/application.yml
+++ b/api/src/main/resources/application.yml
@@ -16,4 +16,5 @@ spring:
 
 tanagra:
   underlay:
-    underlay-yaml-files: ["underlays/aou_synthetic.yaml", "underlays/synpuf.yaml"]
+    # TODO(tjennison): Split these for different environments.
+    underlay-yaml-files: ["underlays/aou_synthetic.yaml", "underlays/synpuf.yaml", "underlays/sd.yaml"]
--- a/api/src/main/resources/underlays/sd.yaml
+++ b/api/src/main/resources/underlays/sd.yaml
@@ -1,0 +1,2604 @@
+# An OMOP underlay schema for the SD dataset.
+
+name: sd
+entities:
+  - name: person
+    attributes:
+      - name: person_id
+        dataType: INT64
+      - name: year_of_birth
+        dataType: INT64
+      - name: gender_concept_id
+        dataType: INT64
+      - name: gender
+        dataType: STRING
+      - name: race_concept_id
+        dataType: INT64
+      - name: race
+        dataType: STRING
+      - name: ethnicity_concept_id
+        dataType: INT64
+      - name: ethnicity
+        dataType: STRING
+      - name: sex_at_birth_concept_id
+        dataType: INT64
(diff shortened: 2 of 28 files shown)


api/src/main/resources/underlays/sd.yaml
```diff
@@ -0,0 +1,2604 @@
+# An initial OMOP underlay schema for development.
```
calbach, line 1: nit: I'm guessing this was copy pasted from the other underlay, if so probably drop

api/src/main/resources/underlays/sd.yaml
```diff
@@ -0,0 +1,2604 @@
...
+        table: observation
+        column: visit_occurrence_id
+  - attribute:
+      entity: observation_occurrence
+      attribute: observation_source_value
+    simpleColumn:
+      columnId:
+        dataset: sd
+        table: observation
+        column: observation_source_value
```
calbach, line 1978: Has the schema been downcased now?

api/src/main/resources/application.yml
```diff
@@ -16,4 +16,4 @@ spring:
 
 tanagra:
   underlay:
-    underlay-yaml-files: ["underlays/aou_synthetic.yaml", "underlays/synpuf.yaml"]
+    underlay-yaml-files: ["underlays/aou_synthetic.yaml", "underlays/synpuf.yaml", "underlays/sd.yaml"]
```
calbach, line 19: I don't expect our dev instance to be able to load this (I'd say it **shouldn't** be able to), should we be splitting our the configs for local vs dev ?

api/src/main/resources/underlays/sd.yaml
```diff
@@ -0,0 +1,2604 @@
+# An initial OMOP underlay schema for development.
+
+name: sd
+entities:
+  - name: person
```
calbach, line 5: As a self-reminder for my next pass on this PR, I want to diff this against the aou synthetic underlay locally

workflow/src/main/resources/queries/sd/aa
```diff
@@ -0,0 +1,23 @@
+-- BuildTextSearchInformation search-strings input query for procedures.
```
calbach, line 1: is the filename "aa" a typo?

calbach (commented):
Could you please add to this PR description the following:

- Summary of changes to the underlay config, compared to whichever one was your starting point
- Summary of SQL changes, i.e. are these all forks of existing SQL, or are there new queries being added? Were there a lot of small changes needed, or was it mostly just a search/replace?

api/src/main/resources/application.yml
tjennison-google: Yes, we will want to split them. I've added a TODO here.

workflow/src/main/resources/queries/sd/aa
tjennison-google: Oops, removed. This was just a temp file.

PR #147, opened 2022-05-10
Neelismail/deleting groups

neelismail:


tjennison-google (approved):
My only feedback is on the PR itself and not the code.

Please update the PR title to be descriptive since that's the primary thing that's visible when looking at the history. It seems to have picked up the branch name instead.

The PR description is fine but will be too verbose for larger changes. In general, focus mostly on why things are changing rather than what specifically changed since that's already visible in the code itself. For larger changes you might want to included an overview of the notable parts of the PR but going file by file will be too much detail. For small changes, something quick like `Added a reducer and menu item.` is plenty, or even nothing beyond a descriptive title.

The most important things to include are things that aren't obvious. There's no question about why you added a reducer and menu item for this feature. But if you had to make changes to group creation to support deletion for some reason that would be helpful to comment on.

PR #148, opened 2022-05-11
Add support for selecting/deselecting all attributes in concept sets

neelismail:


ui/src/datasets.tsx
```diff
@@ -225,9 +227,48 @@ export function Datasets() {
...
+              <Checkbox
+                size="small"
+                fontSize="inherit"
+                name="select-all-values"
+                checked={selectAllChecked}
+                onChange={() =>
+                  updateExcludedAttributes((selection) => {
+                    conceptSetEntities.map((entity) => {
+                      entity.attributes.map((attribute) => {
+                        if (!selection?.get(entity.name)) {
```
tjennison-google, line 251: Instead of toggling each attribute individually, I think you can simplify this by either clearing the selection or creating a new selection including all of the attributes. Immerjs allows you to return a new object instead of updating the passed in one which may help.

If that's not actually an improvement, now that this logic is being copied to a second place refactor it into a separate function inside this component. Something like `setAttributeSelected(entity: string, attribute: string, selected: boolean)`. Then use it both here and below.

Note that you don't need to check `attributes?.has(attribute)` before adding or deleting. Duplicates and non-existent elements will automatically be handled by the Set. It was only checked below since Set doesn't have a built in toggle method.

ui/src/datasets.tsx
```diff
@@ -225,9 +227,48 @@ export function Datasets() {
...
+            </Stack>
+            <Stack direction="row">
+              <Checkbox
+                size="small"
+                fontSize="inherit"
+                name="select-all-values"
+                checked={selectAllChecked}
+                onChange={() =>
+                  updateExcludedAttributes((selection) => {
+                    conceptSetEntities.map((entity) => {
```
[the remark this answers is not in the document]
neelismail: Done

ui/src/datasets.tsx
neelismail: I thought your first suggestion was a smart idea. I refactored the logic for the checkbox so that `excludedAttributes` is either cleared or filled with all attributes depending on the previous state of the "select all" checkbox.

ui/src/datasets.tsx
```diff
@@ -151,6 +151,8 @@ export function Datasets() {
     )),
   });
 
+  const [selectAllChecked, setSelectAllChecked] = useState(true);
```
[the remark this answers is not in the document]
tjennison-google: The size of this list won't be large enough to cause any efficiency issues. However, it can just be a variable instead of a function since the result won't change until the component is re-rendered anyways.

ui/src/datasets.tsx
```diff
@@ -225,9 +236,41 @@ export function Datasets() {
...
-          <Stack direction="row" alignItems="baseline">
-            <Typography variant="h4">3. Values</Typography>
-            <Typography variant="h5">(Columns)</Typography>
+          <Stack
+            direction="row"
+            alignItems="center"
+            justifyContent="space-between"
+          >
+            <Stack direction="row" alignItems="center">
+              <Typography variant="h4" mr={1}>
```
tjennison-google, line 245: Is the margin here still necessary now that the checkbox is right aligned?

ui/src/datasets.tsx
```diff
@@ -225,9 +236,41 @@ export function Datasets() {
...
+                size="small"
+                fontSize="inherit"
+                name="select-all-values"
+                checked={allAttributesChecked()}
+                onChange={() =>
+                  updateExcludedAttributes((selection) => {
+                    selection.clear();
+                    if (allAttributesChecked()) {
+                      conceptSetEntities.forEach((entity) => {
+                        selection.set(entity.name, new Set<string>());
```
tjennison-google, line 261: You can create the set directly from the array: `new Set(entity.attributes)`

tjennison-google (commented):
A GitHub tip is that you can send your comments all at once by "Starting a Review". It's also nice because you can fix and respond to comments as you go but not actually send them until you've had a chance to push your changes. The trick is that you can only do it from the "Files Changed" tab.

ui/src/datasets.tsx
neelismail: I think this was an issue with the previous version too, but without the margin, there isn't a gap between "Values" and "(Columns)". Even if you add trailing and leading spaces, they appear as a connected word. I just added the margin for a stylistic reason, but can remove it if you feel it's unnecessary.

tjennison-google: Ah, I see. That's fine. I had misread anyways and thought this was after "(Columns)".

tjennison-google (approved):

PR #150, opened 2022-05-17
Adding support for multiple attribute ranges

neelismail:


ui/src/criteria/attribute.tsx
```diff
@@ -17,6 +23,12 @@ type Selection = {
   name: string;
 };
 
+type DataRange = {
+  id: string;
+  min: number | undefined;
```
tjennison-google, line 28: min/max previously allowed undefined because that was the case for non-range selection. There's no need to allow it now since an undefined dataRange indicates that. That will clean up places where you have to test min/max for undefined and can instead test dataRanges.

ui/src/criteria/attribute.tsx
```diff
@@ -46,11 +57,21 @@ type AttributeEditProps = {
...
+      initialDataRange.min = integerBoundsHint.min;
+      initialDataRange.max = integerBoundsHint.max;
+    }
+
     return {
       ...data,
       selected: !integerBoundsHint ? [] : undefined,
-      min: integerBoundsHint?.min,
-      max: integerBoundsHint?.max,
+      dataRanges: [initialDataRange],
```
tjennison-google, line 74: This makes it possible for both selected and dataRanges to be defined at the same time which is a confusing state. The above code is also unnecessarily verbose. I would use something like:
```
dataRanges: integerBoundsHint ? {
  id: generateId(),
  min: integerBoundsHint.min || 0,
  max: integerBoundsHint.max || 10000,
},
```

ui/src/datasets.tsx
```diff
@@ -151,6 +151,17 @@ export function Datasets() {
     )),
   });
 
+  const allAttributesChecked = () => {
```
tjennison-google, line 154: You need to rebase your PR.

PR #152, opened 2022-05-20
Define user data as part of the API

tjennison-google:
Also convert existing code to use the new definitions.

### DIFF
--- a/api/src/main/resources/api/service_openapi.yaml
+++ b/api/src/main/resources/api/service_openapi.yaml
@@ -8,7 +8,7 @@ servers:
   - url: /
 
 paths:
-  '/status':
+  "/status":
     get:
       security: []
       summary: Returns the operational status of the service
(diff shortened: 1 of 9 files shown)


api/src/main/resources/api/service_openapi.yaml
```diff
@@ -576,14 +583,128 @@ components:
...
-        $ref: '#/components/schemas/AttributeValue'
+        $ref: "#/components/schemas/AttributeValue"
+
+    UserData:
+      type: object
+      description: The root container for user generated data.
+      properties:
+        version:
+          type: integer
+          description: The version of the data to support forward and backward compatibility.
```
calbach, line 604: nit: would probably clarify: "version of UserData", else this can be misconstrued as a reference to the underlay or API or dataset.

api/src/main/resources/api/service_openapi.yaml
```diff
@@ -576,14 +583,128 @@ components:
...
     EntityInstanceStruct:
       type: object
       description: A map of an entity instance attribute names to their value for this instance.
       additionalProperties:
         type: object
         nullable: true
-        $ref: '#/components/schemas/AttributeValue'
+        $ref: "#/components/schemas/AttributeValue"
+
+    UserData:
```
calbach, line 598: Note: While cohort / conceptSet could (and probably should) be repurposed for a backend user artifact API, this UserData container seems likely to be specific to local storage indefinitely. May be worth a note in the description, but also won't matter too much until we have artifact APIs.

calbach (approved):

api/src/main/resources/api/service_openapi.yaml
tjennison-google: Done.

api/src/main/resources/api/service_openapi.yaml
tjennison-google: I'll leave it for now. It'll either stay as is or migrate into the UI depending on how storage evolves.

PR #153, opened 2022-05-20
Add basic local storage support

tjennison-google:
Storage is implemented by storing whenever the Redux store changes and
loading on app startup using a plugin to allow the implementation to be
swapped out. Depending on requirements goingn forwards, this may need to
be expanded to support editing individual objects, watching for updates
and conflict resolution, offline editing, etc.


ui/cypress/integration/tests.js
```diff
@@ -37,6 +37,8 @@ describe("Basic tests", () => {
     cy.get("input[name='queries-mode']").click();
     cy.contains("SELECT *");
 
+    cy.reload();
```
calbach, line 40: nit: possibly a comment here, if this is testing something around persistence.

Possibly also, should it check that the prior cohort appears in the list still?

ui/src/storage/local.ts
```diff
@@ -0,0 +1,23 @@
+import * as tanagra from "tanagra-api";
+import { registerStoragePlugin, StoragePlugin } from "./storage";
+
+const storageKey = "tanagra-userdata";
+
+@registerStoragePlugin()
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
calbach, line 7: nit: I don't see unused vars - is this needed?

ui/src/storage/storage.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+      storagePlugin.store(data);
+    }
+    return result;
+  };
+
+export function LoadingUserData(props: { children?: React.ReactNode }) {
+  const dispatch = useAppDispatch();
+  const status = useAsync<tanagra.UserData>(
+    useCallback(() => {
```
calbach, line 38: Hmm... `useCallback` inlined inside `useAsync`? Is this a typo? Or does this do something that I'm not understanding?

ui/src/storage/storage.tsx
```diff
@@ -0,0 +1,81 @@
...
+      if (!storagePlugin) {
+        throw new Error("No storage plugin configured.");
+      }
+
+      return storagePlugin.load().then((data?: tanagra.UserData) => {
```
calbach, line 43: In the event that this fails, it seems like ideally we would show an error message, and then alow the user to proceed with no stored artifacts; perhaps also disabling localstorage in general so we're not overwriting their data.

If we don't have error message handling like this, next best is just log the error and allow the user to continue.

ui/src/storage/local.ts
```diff
@@ -0,0 +1,23 @@
+import * as tanagra from "tanagra-api";
+import { registerStoragePlugin, StoragePlugin } from "./storage";
+
+const storageKey = "tanagra-userdata";
```
calbach, line 4: It might make sense to have the key instead be: `tanagra-userdata-${version}`. This would enable a user to continue using the app on the new version, even if we fail to migrate their old data version, without data loss. It does open up more complicated merge/migration scenarios, but it's probably better to be explicit in these cases.

This also would open up more options with rollbacks.

ui/src/storage/local.ts
```diff
@@ -0,0 +1,23 @@
...
+
+@registerStoragePlugin()
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+class LocalStoragePlugin implements StoragePlugin {
+  async store(data: tanagra.UserData): Promise<void> {
+    localStorage.setItem(
```
calbach, line 10: Possibly a future addition, but having a creation / lastModified timestamp on the individual resources in here might be quite useful, even if it isn't exposed to users directly.

ui/src/storage/local.ts
calbach: Right.. the old app not being able to read a new version is exactly why the current approach makes rollbacks impossible; any user that visited the app (and had their localstorage migrated) between initial deployment and rollback would be broken. If this is intended as a stop-gap until we have proper backend storage, then this is fine. Else we will eventually need to consider how the model can support this, even if it is messier.

tjennison-google: The problem is that they're broken in the rollback case too. The older app can't correctly parse the newer data and trying to can potentially destroy their data. Or they get rolled back to the previous version of their data losing any progress in the new version and start making a new branch of edits it's not clear how we could resolve.

I'm also not sure what you mean by proper backend storage. Since the storage in most instances won't be controlled by us (i.e. we can't touch it except through the app), handling it in the backend is just moving the problem. It's actually even more complicated because then there's another potential version difference between the backend and the UI. Handling storage through the backend may still be correct for other reasons but it doesn't matter for this issue.

Let's discuss further and see what we can come up with but in my experience there aren't any rollback paths that are actually practical for end users.

calbach:
LGTM

didn't we have review required via GitHub settings in this repo - did that change or was it just never configured?

calbach:
No problem, I did not think that was done intentionally. Thanks for adding the protection rule.

PR #161, opened 2022-06-14
Initial indexing based on underlay configuration

tjennison-google:
This adds support for indexing based on configuration in the underlay
for the WriteAllNodes workflow. There are several limitations still:
* While the effort is in progress, the workflows are awkwardly split to
  be able to run both individually and based on the underlay. The
  pipeline arguments don't have a good way to support this but once the
  underlay supports all of the workflows we can remove the workflow
  specific arguments from the command line arguments to each workflow
  and pass them as arguments instead.
* Only the WriteAllNodes workflow is supported so far. More will be
  added in future PRs.
* The underlay should be changed to contain more semantic information
  about how entities and attributes are used to allow things like table
  names and other parameters to be inferred directly from the underlay.
* There is no handling for output tables that already exist. It would be
  nice to automatically determine which workflows need to be run based
  on last modified timestamps but just deleting existing tables would be
  a good start.


lib/src/main/proto/bio/terra/tanagra/underlay/indexing.proto
```diff
@@ -0,0 +1,23 @@
+// Messages for underlays.
```
calbach, line 1: nit: comment update - underlay indexing?

api/src/main/resources/underlays/aou_synthetic.yaml
```diff
@@ -2602,3 +2602,12 @@ entityFiltersSchemas:
               int64Val: 2000000009
             - displayName: "Unknown"
               int64Val: 0
+entityIndexing:
+  - entity: condition
+    allNodesWorkflow:
+      outputTable: aou_synthetic_SR2019q4r4_indexes.condition_all_nodes
```
calbach, line 2608: Per your PR description, definitely need to figure out a way to connect this to the rest of the underlay; a fair amount of duplication and room for error with this approach.

calbach (approved):

api/src/main/java/bio/terra/tanagra/app/configuration/UnderlayConfiguration.java
```diff
@@ -61,6 +63,13 @@ public List<Underlay> getUnderlays() throws IOException {
     return result;
   }
 
+  public static Underlay loadFromYamlFile(String file) throws IOException {
```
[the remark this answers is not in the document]
tjennison-google: I've had to refactor this code into the lib subproject to properly satisfy some build issues. Please take another look.

calbach (approved):

PR #163, opened 2022-06-30
Implementing undo and redo functionality

neelismail:


node_modules/.package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
tjennison-google, line 1: There shouldn't be UI related files in the root of the repo. Same with the next two files.

ui/src/urlSlice.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const initialState = "/";
+
+const urlsSlice = createSlice({
+  name: "urlsSlice",
+  initialState,
+  reducers: {},
+  extraReducers: (builder) => {
+    builder.addMatcher(isUndoableAction, () => {
+      return getCurrentPageUrl();
```
tjennison-google, line 18: Good idea. This is a much cleaner approach than using middleware.

ui/src/criteria/attribute.tsx
```diff
@@ -410,8 +410,8 @@ function AttributeDetails(props: AttributeDetailsProps) {
   if (props.data.selected?.length) {
     return (
       <>
-        {props.data.selected.map(({ id, name }) => (
-          <Stack direction="row" alignItems="baseline" key={Date.now()}>
+        {props.data.selected.map(({ id, name }, index) => (
+          <Stack direction="row" alignItems="baseline" key={index}>
```
tjennison-google, line 414: This shouldn't be in this PR. Also, this is incorrect for changes other than simple adds or deletes at the end of the list. Any reordering or editing elsewhere in the list will result in keys mapping to different items.

ui/src/containers/UndoRedo.tsx
```diff
@@ -0,0 +1,60 @@
...
+        <UndoIcon
+          fontSize="medium"
+          sx={{ color: canUndo ? "white" : "gray" }}
+        />
+      </Button>
+      <Button
+        onClick={onRedo}
+        disabled={!canRedo}
+        component={RouterLink}
+        to={urlHistory.future[urlHistory.future.length - 1]}
```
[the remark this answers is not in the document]
neelismail: You're right, I updated my code to reflect this.

ui/src/containers/UndoRedo.tsx
```diff
@@ -0,0 +1,60 @@
...
+function UndoRedo({ canUndo, canRedo, onUndo, onRedo }: propsFromRedux) {
+  const urlHistory = useUrl();
+
+  return (
+    <Box>
+      <Button
+        onClick={onUndo}
+        disabled={!canUndo}
+        component={RouterLink}
+        to={urlHistory.present}
```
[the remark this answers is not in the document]
neelismail: It seems like `urlHistory.present` is correct because this is the URL before the state change and subsequent navigation occurs, so this is where we would like to navigate to when we undo a state change.

ui/src/rootReducer.ts
```diff
@@ -2,24 +2,40 @@ import { combineReducers, createAction } from "@reduxjs/toolkit";
...
+import undoable, { excludeAction } from "redux-undo";
 import * as tanagra from "tanagra-api";
 import underlaysReducer from "underlaysSlice";
+import urlSlice from "urlSlice";
+
+export const loadUserData = createAction<tanagra.UserData>("loadUserData");
+
+const undoableConfigs = {
+  initTypes: ["loadUserData"],
+  filter: excludeAction([loadUserData, "underlays/setUnderlays"]),
```
tjennison-google, line 14: On the other hand, it seems unnecessary to include `loadUserData` here since it's the initial state.

ui/src/router.tsx
```diff
@@ -96,3 +96,8 @@ function NotFound(props: { error: Error; resetErrorBoundary?: () => void }) {
     </>
   );
 }
+
+export const getCurrentPageUrl = (): string => {
```
tjennison-google, line 100: `export function getCurrentUrl(): string {` to be consistent with the rest of the code. Also drop `Page` from the name.

ui/src/router.tsx
```diff
@@ -96,3 +96,8 @@ function NotFound(props: { error: Error; resetErrorBoundary?: () => void }) {
...
+  const pathStartIndex = window.location.href.indexOf("#") + 1;
```
tjennison-google, line 101: This would be cleaner as:
```
const url = window.location.href;
return url.slice(url.indexOf("#") + 1);
```

PR #165, opened 2022-07-06
Ordering concepts sets on the concept set edit page by specified property described in underlay

neelismail:


ui/src/criteria/concept.tsx
```diff
@@ -511,6 +511,7 @@ function searchRequest(
       entityDataset: {
         entityVariable: "c",
         selectedAttributes: attributesForEntity(entity, columns),
+        orderByAttribute: "concept_name",
```
tjennison-google, line 514: Since roll up counts are in, we should order by those instead of name. That would be `person_count`. However, that doesn't work for brands so you'll need to add `orderBy` and `orderDirection` property to the `EntityConfig` in the `CriteriaConfig` and use `concept_name` for brands and `person_count` for everything else. `person_count` should be sorted in descending order.

api/src/main/resources/underlays/aou_synthetic.yaml
```diff
@@ -289,7 +289,7 @@ criteriaConfigs: >-
         {"key":"vocabulary_id","width":120,"title":"Vocab"},
         {"key":"concept_code","width":120,"title":"Code"}],
       "entities":[
-        {"name":"condition","selectable":true,"hierarchical":true}
+        {"name":"condition","selectable":true,"hierarchical":true,"orderBy":"person_count","orderDirection":"Desc"}
```
tjennison-google, line 292: All caps the enum values like `DESC` to match the API formatting.

ui/src/criteria/concept.tsx
```diff
@@ -44,6 +44,9 @@ type EntityConfig = {
   sourceConcepts?: boolean;
   attributes?: string[];
 
+  orderBy?: string;
+  orderDirection?: string;
```
tjennison-google, line 48: Make this a `tanagra.OrderByDirection` instead. Then it will automatically deserialize as the right type and you won't need the comparison below.

tjennison-google (approved):

PR #166, opened 2022-07-07
Expose rollup count attribute for all concept entities.

marikomedlock:
Added a `person_count` attribute to all concept entities in the `aou_synthetic` underlay. This attribute points to the rollup count index tables.

As part of this PR, I also generated the index tables in the `broad-tanagra-dev` and `verily-tanagra-dev` projects. The commands for generating them in the Broad project are included in `worklog.md`. To generate the index tables in the Verily project, I did the following:
- Requested membership in the on-demand VerilyGroup `tanagra-dev-breakglass`. This gives me permission to run Dataflow jobs and write to BigQuery in the Verily project.
- Replaced the project name in the shell commands `broad-tanagra-dev` -> `verily-tanagra-dev` and in the input SQL files.
- Replaced the service account email in the commands `tanagra@broad-tanagra-dev.iam.gserviceaccount.com` -> `workflow-runner@verily-tanagra-dev.iam.gserviceaccount.com`.
- Ran each command and confirmed the job finished successfully and created the index table.
[verilyProjectIndexingCmds.txt](https://github.com/DataBiosphere/tanagra/files/9095930/verilyProjectIndexingCmds.txt)


calbach (commented):
Code looks reasonable to me, I'll defer to Tim on discussion of the underlay manifestation of this.

marikomedlock:
> One thing to consider here is that there are actually two counts: per node counts and roll up counts. I think the other PR is generating rollup counts so we may want to call these `person_rollup_count` to leave `person_count` for the other one. On the other hand, if we're considering this as a temporary state then it doesn't matter too much.

Are we going to need to expose the per node counts in the future also? To make sure we're talking about the same thing, e.g. `endocrineDisorder` has 2 children, `diabetesType1` and `diabetesType2`. Per node count for `endocrineDisorder` includes people specifically diagnosed with `endocrineDisorder`, not one of its children. Rollup count for `endocrineDisorder` includes its own per node count and the per node counts of all of its descendants.

tjennison-google:
> > One thing to consider here is that there are actually two counts: per node counts and roll up counts. I think the other PR is generating rollup counts so we may want to call these `person_rollup_count` to leave `person_count` for the other one. On the other hand, if we're considering this as a temporary state then it doesn't matter too much.
> 
> Are we going to need to expose the per node counts in the future also? To make sure we're talking about the same thing, e.g. `endocrineDisorder` has 2 children, `diabetesType1` and `diabetesType2`. Per node count for `endocrineDisorder` includes people specifically diagnosed with `endocrineDisorder`, not one of its children. Rollup count for `endocrineDisorder` includes its own per node count and the per node counts of all of its descendants.

Yes, that's it. I suppose that's ultimately a product decision but they're both in the AoU UI.

marikomedlock:
> person_rollup_count

Gotcha. Okay, I'll leave this as `person_count` for now since I've already generated all the index tables and named them e.g. `condition_person_counts`. When/if we add the node counts, we can either call them `person_node_count` or rename both the attributes and index tables. Once indexing is automated, it should be much easier to tweak naming.

PR #168, opened 2022-07-12
Add count API endpoints.

marikomedlock:
Added 2 new API endpoints for counting entity instances.
- `counts:generateSqlQuery` returns the SQL query string
- `counts:search` returns the result set

The purpose of these endpoints is to allow the UI to show a breakdown of the selected cohort.
- How many people are in the cohort.
- How many people of each gender+race combination are in the cohort.

The `counts:search` result objects each contain a count and the set of attribute values that define the group. e.g. `{15, [gender:Male, race:American]}, {32, [gender:Female, race:Japanese]}`.

Included tests with example API calls for the `aou_synthetic` underlay: counting the total number of people, counting the number of people in a cohort broken down by gender+race.


api/src/main/java/bio/terra/tanagra/app/controller/EntityCountsApiController.java
```diff
@@ -0,0 +1,57 @@
...
+    String sql = queryService.generateSql(entityCounts);
+    return ResponseEntity.ok(new ApiSqlQuery().query(sql));
+  }
+
+  @Override
+  public ResponseEntity<ApiSearchEntityCountsResponse> searchEntityCounts(
+      String underlayName,
+      String entityName,
+      ApiSearchEntityCountsRequest body,
+      Integer pageSize,
```
calbach, line 46: pagination parameters appear to be ignored; I would probably omit these from the API spec if they are not supported yet

api/src/main/java/bio/terra/tanagra/service/query/QueryService.java
```diff
@@ -127,6 +155,62 @@ Query createQuery(EntityDataset entityDataset) {
...
+
+    // select fields for a count query include: COUNT(tableAlias), [group by attributes],
+    // [additional selected attributes]
+    List<Selection> selectFields = new ArrayList<>();
+    selectFields.add(
+        Selection.Count.builder()
+            .name(COUNT_ALIAS)
+            .entityVariable(entityCounts.primaryEntity())
+            .build());
+    entityCounts.groupByAttributes().stream()
```
calbach, line 183: Looks like the mappers are identical here, so I would probably go for:

```
Stream.concat(
  entityCounts.groupByAttributes().stream(),
  entityCounts.additionalSelectedAttributes().stream()
).map(
  ...
).forEach(selectedFields::add);
```

api/src/main/resources/api/service_openapi.yaml
```diff
@@ -616,6 +691,38 @@ components:
...
+    SearchEntityCountsRequest:
+      type: object
+      description: A request for counting groups of entity instances.
+      properties:
+        entityCounts:
+          $ref: "#/components/schemas/EntityCounts"
+
+    SearchEntityCountsResponse:
+      type: object
+      description: A view of entity counts for a search request.
```
calbach, line 703: If we are supporting pagination, should we make any statement about the order of results here?

api/src/main/java/bio/terra/tanagra/service/query/EntityCounts.java
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * <p>e.g. {20, [Male, American]}, {35, [Female, Japanese]}
+ */
+@AutoValue
+public abstract class EntityCounts {
+  /** The primary entity and variable that the counts are being created for. */
+  public abstract EntityVariable primaryEntity();
+
+  /**
+   * Any additional attributes to return as columns of the dataset. These must be attributes of the
```
calbach, line 24: Hmm.. how does this work? Wouldn't it require that we either group by this additional attribute, OR we are already grouping by an attribute that maps 1:1 with this selected attribute?

api/src/main/java/bio/terra/tanagra/service/query/EntityCounts.java
```diff
@@ -0,0 +1,85 @@
...
+
+    public abstract Builder primaryEntity(EntityVariable primaryEntity);
+
+    public abstract EntityVariable primaryEntity();
+
+    public abstract Builder additionalSelectedAttributes(List<Attribute> groupByAttributes);
+
+    public abstract ImmutableList<Attribute> additionalSelectedAttributes();
+
+    public abstract Builder groupByAttributes(List<Attribute> groupByAttributes);
```
calbach, line 53: Are there restrictions around this, e.g. only for indexed counts? If so - where is this enforced?

calbach (commented):
thanks, some high level questions on the API and behaviors

api/src/main/java/bio/terra/tanagra/app/controller/EntityCountsApiController.java
```diff
@@ -0,0 +1,57 @@
...
+      String pageToken) {
+    // TODO authorization check.
+    EntityCounts entityCounts =
+        apiConversionService.convertEntityCounts(underlayName, entityName, body.getEntityCounts());
+    QueryResult queryResult = queryService.retrieveResults(entityCounts);
+    ApiSearchEntityCountsResponse response =
```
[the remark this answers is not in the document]
marikomedlock: Removed the separate variable definition and returned directly.

api/src/main/java/bio/terra/tanagra/app/controller/EntityCountsApiController.java
marikomedlock: Removed. I had included these to be consistent with the existing entityInstances API endpoints, but agreed it's probably better to just omit them until we're ready to support them.

api/src/main/resources/api/service_openapi.yaml
marikomedlock: In practice, I expect us to not need pagination for this endpoint. The current use case is for a cohort breakdown. So if there are so many groups that we need to page through them, I think it would be hard to display them in a breakdown graphic.

Regardless, I could order the results by the same group by attributes. e.g. `GROUP BY gender, race ORDER BY gender, race`. @tjennison-google would that be useful or does the order not matter to the UI (assuming everything comes back in one request)?

api/src/main/java/bio/terra/tanagra/service/query/QueryService.java
marikomedlock: Ah that looks nicer, thanks!

api/src/main/java/bio/terra/tanagra/service/query/EntityCounts.java
marikomedlock: Yes, you can only group by an attribute that maps 1:1 with this selected attribute. My intent here was specifically to support returning e.g. gender name when grouping by gender concept id, so that the UI can use display names.

Now I'm thinking about it, I might be able to add validation for this -- make sure all the additional selected attributes are mapped to values foreign-keyed to one of the group by attributes. I'm going to investigate this some more.

api/src/main/java/bio/terra/tanagra/service/query/EntityCounts.java
marikomedlock: No restrictions, since this won't touch any of the indexed/precomputed rollup counts. These APIs are specifically for the dynamic rollup counts = how many men/women there are in my cohort. We can't precompute this because it depends on the definition of the cohort. By contrast, static count = how many people have a condition. We can precompute that because it doesn't depend on a cohort.

api/src/main/resources/api/service_openapi.yaml
tjennison-google: I don't think the order matters and I agree that we can ignore pagination for now unless we find a need for it. The UI would need the complete results to do anything with them so I don't think pagination is useful unless we hit a response size limit or something.

tjennison-google (commented):
Looks good to me from the API end.

api/src/main/java/bio/terra/tanagra/service/query/EntityCounts.java
marikomedlock: OK, this looks like a pretty involved validation change. I was hoping to check the `AttributeMappings` specified in the underlay, to validate that each `additionalSelectedAttribute` maps to a `LookupColumn` that uses a key identical to the column that one of the `groupByAttributes` is mapped to. However, the `AttributeMappings` are stored separately from the `Entity` object, so I don't have access to it in this `EntityCounts` class. Another (I think less ideal) option is to validate this during the SQL query construction. However, the way the visitor pattern is currently setup, when resolving a `Selection` expression, which includes `GROUP BY` and `SELECT` fields, we don't have references to other fields or their mappings.

I could change the code to make either of these approaches possible -- the relevant information is in the underlay, I just need to plumb it through to where we want to do the validation. However, I think a better approach is to add the idea of an attribute that is a bundle of fields (e.g. Gender attribute = `gender_concept_id` field and FK'ed `gender_concept_name`). When someone wants to select the Gender attribute, we always include both fields. If we implement this, then we wouldn't need the `additionalSelectedAttributes` API field, and this validation would be moot. It would also help when querying entity instances -- the UI would just request the Gender attribute, instead of having to remember to request both the `gender_concept_id` and the `gender_concept_name`. What do you think?

tjennison-google: I think the bundle approach is the way to go but we should probably approach that as part of the overall underlay work.

api/src/main/java/bio/terra/tanagra/service/query/EntityCounts.java
calbach: OK, I ask this because it seems like I could pretty easily construct a query of death by accident or intentionally by doing something dumb like: GROUP BY condition_concept_id

I'm wondering whether there should be some kind of reasonably high limit or constraint on the number of buckets this generates. e.g. only returns up to the top 10K buckets, or LIMIT the group by query to 10K and fail if we reach said limit in results. Something to avoid OOMs here.

calbach (approved):

api/src/main/java/bio/terra/tanagra/service/query/EntityCounts.java
marikomedlock: Good point. Right now, I think we'll only ever get a single page size of results back, similar to what happens when the UI requests e.g. a list of all conditions, so I don't think OOMs will be an immediate problem. Longer-term, here are 2 things we can do that will address this:
1. Include a limit field for all endpoints that return query results. Set the default to something that won't cause OOMs e.g. 100.
2. Specify in the underlay which attributes we want to allow grouping by. The UI will need this somehow anyway, so it knows what attributes to show in the breakdown panel. Then when the UI hits the count API endpoint, we can validate that all of groupByAttributes are allowable. Presumably this will avoid queries of death because a human has indicated that grouping by this attribute is reasonable.

PR #169, opened 2022-07-15
Add dev instance info to README.

marikomedlock:
Added a link to the dev instance and a note to request access to the appropriate VerilyGroup to the README. Goal is just to have this info in a central place where people can find it. I decided to put this here instead of in a Google doc because I think it's more likely we won't forget to keep this up-to-date. I will add a `go/` link to this section of the README also.


tjennison-google (approved):
Given that this repo is public, is this the best place for this information? Simon mentioned in chat that there's a doc to point people to which is maybe sufficient?

PR #173, opened 2022-07-25
Add a business logic layer between the API and UI

tjennison-google:
This layer will allow the UI to remain generic by pushing any underlay
specific logic into one place, and ideally into configuration. This will
make it easier to evolve the underlay and API as part of the underlay
overhaul. It also makes it much easier to understand what logic
is actually being applied when it's not scattered throughout the UI and
will make it possible to test the logic and UI separately which will
make both sets of tests easier to write.

This change involves three major pieces:
* Moving logic about the structure of the data (occurrences,
  hierarchies, etc.) to configuration in the underlay and adding an
  interface to make requests based on it.
* Restructuring the UI configuration in the underlay to contain
  multiple elements, including the new configuration.
* Changing the concept criteria to use the new interface and validate
  the configurantion and interface are sufficient.

What's not yet complete:
* Datasets page, aattribute criteria, and a few other minor items have
  not been migrated.
* There are some TODOs for places that where the new layer was
  insufficient and some improvements to the current API or data model
  will be needed.
* There are some TODOs left to clean up some adjacent code.

### DIFF
--- a/api/src/main/java/bio/terra/tanagra/app/controller/UnderlaysApiController.java
+++ b/api/src/main/java/bio/terra/tanagra/app/controller/UnderlaysApiController.java
@@ -70,6 +70,6 @@ public class UnderlaysApiController implements UnderlaysApi {
     return new ApiUnderlay()
         .name(underlay.name())
         .entityNames(entityNames)
-        .criteriaConfigs(underlay.criteriaConfigs());
+        .uiConfiguration(underlay.uiConfiguration());
   }
 }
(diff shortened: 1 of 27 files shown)


kemp-verily (commented):
LGTM for typescript - I didn't look at any of the Java code and make no claims about the business logic changes here (assuming Mariko is reviewing for that)

marikomedlock (approved):
Backend changes lgtm. Discussed offline with Tim.

PR #174, opened 2022-07-27
Fixing bug with undo / redo

neelismail:


ui/src/hooks.ts
```diff
@@ -59,5 +59,9 @@ export function useConceptSet() {
 }
 
 export function useUrl() {
```
tjennison-google, line 61: Rename this to better describe what it actually does now. Something like `useUndoRedoURLs`.

tjennison-google (approved):

PR #175, opened 2022-07-29
Expand business logic layer to attribute criteria

tjennison-google:
* Adds a function to fetch hints to the data source.
* Converts hint fetching to an asynchronous operation to support hints
  not always being in the underlay.
* Fixes concept.test.tsx to use the data source and removes the global
  store setup.
* Passes the data source to criteria creation instead of the underlay
  now that both criteria have been converted.

### DIFF
--- a/ui/src/criteria/attribute.tsx
+++ b/ui/src/criteria/attribute.tsx
@@ -13,15 +13,23 @@ import Input from "@mui/material/Input";
 import Stack from "@mui/material/Stack";
 import Typography from "@mui/material/Typography";
 import { CriteriaPlugin, generateId, registerCriteriaPlugin } from "cohort";
-import { useUnderlay } from "hooks";
+import Loading from "components/loading";
+import { DataValue } from "data/configuration";
+import {
+  attributeValueFromDataValue,
+  EnumHintOption,
+  IntegerHint,
+  Source,
+  useSource,
+} from "data/source";
+import { useAsyncWithApi } from "errors";
 import produce from "immer";
-import React, { useState } from "react";
+import React, { useCallback, useState } from "react";
 import * as tanagra from "tanagra-api";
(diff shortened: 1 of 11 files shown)


ui/src/criteria/attribute.tsx
```diff
@@ -407,18 +382,18 @@ type AttributeDetailsProps = {
...
           <Stack direction="row" alignItems="baseline" key={Date.now()}>
-            <Typography variant="body1">{id}</Typography>&nbsp;
+            <Typography variant="body1">{value}</Typography>&nbsp;
             <Typography variant="body2">{name}</Typography>
           </Stack>
         ))}
       </>
     );
-  } else if (props.data.dataRanges?.length) {
+  } else if (props.data.dataRanges.length > 0) {
```
kemp-verily, line 396: drop `else if` since you're returning above?

kemp-verily (approved):
LGTM for typescript (one nit)

marikomedlock (approved):
LGTM.

RE "hints not always being in the underlay" from PR description: Won't the UI hints always be in the underlay? Some may not be in the initial config supplied by the user, but then we should generate them as part of the indexing process, so wouldn't they be present in the expanded config used by the API?

tjennison-google:
> LGTM.
> 
> RE "hints not always being in the underlay" from PR description: Won't the UI hints always be in the underlay? Some may not be in the initial config supplied by the user, but then we should generate them as part of the indexing process, so wouldn't they be present in the expanded config used by the API?

My assumption is that we won't store all of the hints for all of the individual entities instances (e.g. blood pressure) in the underlay. In that case this provides the opportunity to fetch them, whether they're precomputed or computed on the fly. Maybe that will also end up applying to the demographic data as well or maybe not but this is more flexible in any case.

PR #178, opened 2022-08-05
Generalizing demographic charts on overview page

neelismail:


ui/src/overview.tsx
```diff
@@ -359,83 +429,132 @@ function DemographicCharts({ cohort }: DemographicChartsProps) {
...
-        .get(genderAgeRange)
-        .set(race, currCountForRace + count);
+      chartConfigs.forEach((config, index) => {
+        // Determine the primary data groupings for the chart.
+        const primaryPropertyComponents: string[] = [];
+        config.primaryProperties.forEach((property) => {
+          // Temporarily check if property is age since API only supports year of birth.
+          const propertyName =
+            property.key === "age" ? "year_of_birth" : property.key;
+          const propertyValues = Object.entries(
```
tjennison-google, line 450: Is this not just reading a named property of the object? Isn't this equivalent to `demographicData[i]?.[propertyName]`?

ui/src/overview.tsx
```diff
@@ -359,83 +429,132 @@ function DemographicCharts({ cohort }: DemographicChartsProps) {
...
+            demographicData[i]?.definition ?? {}
+          ).find((data) => data[0] === propertyName);
+
+          if (propertyValues) {
+            const dataIntegerVal = propertyValues[1].int64Val;
```
tjennison-google, line 455: This can be simplified as: `const value = propertyValues[1]?.int64Val ?? propertyValues[1]?.stringVal ?? propertyValues[1]?.boolVal`.

I think you'll be able to do the bucketing using `value` without checking the type because Javascript, but if not, use `typeof value === "number"`.

ui/src/overview.tsx
```diff
@@ -359,83 +429,132 @@ function DemographicCharts({ cohort }: DemographicChartsProps) {
...
+            const dataStringVal = propertyValues[1].stringVal;
+            const dataBoolVal = propertyValues[1].boolVal;
+            // The UI currently assumes that property.value represents an age bucket if defined.
```
tjennison-google, line 458: The bucketing can be used generically. Only the age calculation needs to be hardcoded so do it based on the property. Combined with the previous comment it would be something like:
```
let value = ...;
if (property.key === "age") {
  value = new Date().getFullYear() - value;
}
```

Is the `parseInt`/`toString()` actually accomplishing something?

ui/src/overview.tsx
```diff
@@ -359,83 +429,132 @@ function DemographicCharts({ cohort }: DemographicChartsProps) {
...
-        .get(race);
-
-      demographicsByGenderAgeRace
-        .get(genderAgeRange)
-        .set(race, currCountForRace + count);
+      chartConfigs.forEach((config, index) => {
+        // Determine the primary data groupings for the chart.
+        const primaryPropertyComponents: string[] = [];
+        config.primaryProperties.forEach((property) => {
+          // Temporarily check if property is age since API only supports year of birth.
```
tjennison-google, line 447: Format this as `TODO(neelismail): Remove age handling once the API supports them.`. The standard formatting allows automated tools to process them.

ui/src/overview.tsx
```diff
@@ -359,83 +429,132 @@ function DemographicCharts({ cohort }: DemographicChartsProps) {
...
+              primaryPropertyComponents.push(dataStringVal);
+            } else if (dataBoolVal) {
+              primaryPropertyComponents.push(`${dataBoolVal}`);
+            }
+          }
+        });
+
+        // Add the primary property grouping to the chart data if it hasn't already
+        const primaryPropertyString = primaryPropertyComponents.join(" ");
+        if (!chartsData[index].has(primaryPropertyString)) {
```
tjennison-google, line 489: This code would be cleaner if it looped over charts -> demographics rather than demographics -> charts like it does now. You wouldn't need checks like this because you would only be building and pushing the data for each chart once.

ui/src/overview.tsx
```diff
@@ -359,83 +429,132 @@ function DemographicCharts({ cohort }: DemographicChartsProps) {
...
-      demographicsByGenderAgeRace: Array.from(
-        demographicsByGenderAgeRace,
-        ([key, value]) => {
-          return {
-            ageRange: key,
-            ...Object.fromEntries(value),
-          };
+      chartsData: chartsData.map((chart, index) => {
+        if (stackedProperties[index].size > 0) {
+          return Array.from(chart, ([key, value]) => ({
```
tjennison-google, line 545: I haven't hunted through why this is being done but it feels like it shouldn't be necessary. Can you process the data into the correct format on the first pass?

ui/src/overview.tsx
```diff
@@ -359,83 +429,132 @@ function DemographicCharts({ cohort }: DemographicChartsProps) {
...
-      }),
-      demographicRaces: Array.from(demographicRaces),
-      demographicsByGenderAgeRace: Array.from(
-        demographicsByGenderAgeRace,
-        ([key, value]) => {
-          return {
-            ageRange: key,
-            ...Object.fromEntries(value),
-          };
+      chartsData: chartsData.map((chart, index) => {
```
tjennison-google, line 543: The chart data needs to be sorted (alphabetically is fine) so that the bars appear in a consistent order.

ui/src/overview.tsx
```diff
@@ -321,27 +399,92 @@ function DemographicCharts({ cohort }: DemographicChartsProps) {
...
+        entityCountPropertyValue.stringVal ??
+        entityCountPropertyValue.boolVal;
+
+      if (value) {
+        // TODO(neelismail): Remove age handling once the API supports them.
+        if (property.key === "age" && typeof value === "number") {
+          value = new Date().getFullYear() - value;
+        }
+
+        const definedValue = value;
```
neelismail, line 425: For some reason, I have to create this `definedValue` in order to prevent JavaScript from complaining that `value may be undefined` within the `if (property.buckets) { ... }` block. I thought this was especially strange because on line 419 I do a check to ensure `value` is defined. Do you have any suggestions about how this should be handled?

ui/src/overview.tsx
```diff
@@ -321,27 +399,93 @@ function DemographicCharts({ cohort }: DemographicChartsProps) {
...
+      underlay.uiConfiguration.demographicChartConfigs.groupByAttributes;
+    const additionalSelectedAttributes = new Set<string>();
+    underlay.uiConfiguration.demographicChartConfigs.chartConfigs.forEach(
+      (config) => {
+        config.primaryProperties.forEach((property) => {
+          // TODO(neelismail): Remove guard for age property key when API provides age support
+          if (
+            !additionalSelectedAttributes.has(property.key) &&
+            (!groupByAttributes.includes(property.key) ||
+              (property.key === "age" &&
```
tjennison-google, line 458: Instead of handling "age" everywhere, treat it like everything else and replace it at the end.

neelismail: Done.

ui/src/underlaysSlice.ts
```diff
@@ -20,6 +20,30 @@ export type Underlay = {
...
+  primaryProperties: ChartConfigProperty[];
+  stackedProperty?: ChartConfigProperty;
+};
+
+export type ChartConfigProperty = {
+  key: string;
+  buckets?: Range[];
+};
+
+export type Range = {
```
tjennison-google, line 43: This should also be renamed to `Bucket`.

tjennison-google (approved):
Approved with that last rename.

PR #181, opened 2022-08-11
Displaying counts for groups and criteria on overview page

neelismail:


ui/src/overview.tsx
```diff
@@ -190,6 +194,39 @@ function ParticipantsGroup(props: { group: tanagra.Group; index: number }) {
     ],
   });
 
+  const fetchGroupCount = useCallback(async () => {
+    const groupForFilter: tanagra.Group = {
+      id: props.group.id,
```
tjennison-google, line 199:
```
{
  ...props.group,
  kind: tanagra.GroupKindEnum.Included,
}
```

ui/src/overview.tsx
```diff
@@ -190,6 +194,39 @@ function ParticipantsGroup(props: { group: tanagra.Group; index: number }) {
...
+      kind: tanagra.GroupKindEnum.Included,
+      criteria: props.group.criteria,
+    };
+
+    const cohortForFilter: tanagra.Cohort = {
+      id: cohort.id,
+      name: cohort.name,
+      underlayName: cohort.underlayName,
+      groups: [groupForFilter],
```
tjennison-google, line 208: Inline `groupForFilter`. It short enough there's no need for a separate variable.

ui/src/overview.tsx
```diff
@@ -266,6 +318,39 @@ function ParticipantCriteria(props: {
...
+      id: props.group.id,
+      kind: tanagra.GroupKindEnum.Included,
+      criteria: [props.criteria],
+    };
+
+    const cohortForFilter: tanagra.Cohort = {
+      id: cohort.id,
+      name: cohort.name,
+      underlayName: cohort.underlayName,
+      groups: [groupForFilter],
```
tjennison-google, line 332: Inline `groupForFilter`.

ui/src/overview.tsx
```diff
@@ -190,6 +194,39 @@ function ParticipantsGroup(props: { group: tanagra.Group; index: number }) {
...
+    };
+
+    const searchEntityCountsRequest: tanagra.SearchEntityCountsRequest = {
+      entityCounts: {
+        entityVariable: "p",
+        additionalSelectedAttributes: [],
```
tjennison-google, line 214: Can you just omit these instead of passing empty arrays?

ui/src/overview.tsx
```diff
@@ -298,9 +383,13 @@ function ParticipantCriteria(props: {
               {props.criteria.name}
             </Link>
             <Divider orientation="vertical" variant="middle" flexItem />
-            <Typography variant="body1">
-              {0 /* TODO(tjennison): Fetch from backend. */}
-            </Typography>
+            {
+              <Loading status={criteriaCountState}>
```
tjennison-google, line 387: Does this display a sensibly sized spinner and not a giant one like elsewhere in the UI?

ui/src/overview.tsx
```diff
@@ -211,8 +248,21 @@ function ParticipantsGroup(props: { group: tanagra.Group; index: number }) {
...
+          key=""
+          display={"flex"}
+          flexDirection={"row"}
+          justifyContent={"space-between"}
+          alignItems="center"
+        >
           <AddCriteriaButton group={props.group.id} />
+          {
+            <Loading status={groupCountState}>
+              <Typography variant={"body1"} fontWeight={"bold"}>
```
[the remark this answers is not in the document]
neelismail: Done.

ui/src/overview.tsx
neelismail: Done.

ui/src/overview.tsx
neelismail: Done.

ui/src/overview.tsx
neelismail: I checked and they're optional, so I've omitted them.

ui/src/overview.tsx
neelismail: Done.

ui/src/overview.tsx
neelismail: Would it make sense to modify the existing `Loading` component to have an optional `size` prop, so smaller spinners can be specified? Or would you prefer me to just render a Material UI spinner directly in these cases?

tjennison-google: You should edit the `Loading` component so the handling is consistent. However, it's also needs to consider what happens with errors in the small space. Look at what the `TreeGrid` does, using an icon for errors that can be clicked to reload and showing the error message as a mouse over.

neelismail: I've modified the `Loading` component to render customized spinners and error messages if a `size` prop with value `small` is supplied. The spinners and error messages are displaying appropriately on the UI, but let me know if you have any concerns about the implementation.

ui/src/components/loading.tsx
```diff
@@ -38,12 +41,24 @@ export default function Loading(props: Props) {
...
 
-function showStatus(visible: boolean, status?: Status): ReactNode {
+function showStatus(
+  visible: boolean,
+  status?: Status,
+  size?: string
+): ReactNode {
   if (status?.error && !status?.isPending) {
-    return (
+    return size === "small" ? (
```
tjennison-google, line 57: Ternary operators that span more than a couple lines are difficult to read so avoid them when it's reasonable to do so. In this case it's easy:
```
if (size === "small") {
  return ...;
}
return ...;
```

ui/src/components/loading.tsx
```diff
@@ -38,12 +41,24 @@ export default function Loading(props: Props) {
     return <>{props.children}</>;
   }
 
-  return <Box className="loading">{showStatus(visible, props.status)}</Box>;
+  return (
+    <Box className={props.size === "small" ? "" : "loading"}>
```
tjennison-google, line 45: How about `loading-small` instead of an empty name?

ui/src/components/loading.tsx
```diff
@@ -38,12 +41,24 @@ export default function Loading(props: Props) {
...
-function showStatus(visible: boolean, status?: Status): ReactNode {
+function showStatus(
+  visible: boolean,
+  status?: Status,
+  size?: string
+): ReactNode {
   if (status?.error && !status?.isPending) {
-    return (
+    return size === "small" ? (
+      <Tooltip title={status.error?.message} arrow={true}>
```
tjennison-google, line 58: Pull the error message logic below into a variable that can be used here as well to avoid the potentially empty tooltip.

neelismail: Done.

ui/src/components/loading.tsx
```diff
@@ -38,21 +41,36 @@ export default function Loading(props: Props) {
...
 }
 
-function showStatus(visible: boolean, status?: Status): ReactNode {
+function showStatus(
+  visible: boolean,
+  status?: Status,
+  size?: string
+): ReactNode {
   if (status?.error && !status?.isPending) {
+    const errorMessage = status.error
```
tjennison-google, line 57: `status.error.message`

ui/src/components/loading.tsx
```diff
@@ -38,21 +41,36 @@ export default function Loading(props: Props) {
...
         <Typography variant="h2">Error</Typography>
-        {status.error?.message ? (
-          <Typography paragraph>
-            An error has occurred: <b>{status.error.message}</b>
-          </Typography>
-        ) : (
-          <Typography paragraph>An unknown error has occurred.</Typography>
-        )}
+        <Typography paragraph>
+          <b>{errorMessage}</b>
```
tjennison-google, line 72: The bold was formatting the error portion of the message. Just remove it.

tjennison-google (approved):

ui/src/components/loading.tsx
neelismail: Done.

ui/src/components/loading.tsx
neelismail: Done.

PR #182, opened 2022-08-12
[PF-1731] Replace Java setup action

zloery:
The `joschi/setup-jdk@v2` Github action is deprecated and the recommendation is to move to the official github setup-java action with the `temurin` distribution: see https://github.com/joschi/setup-jdk#-notice


marikomedlock (approved):
Thanks Zach for taking care of this! Sorry for the delay in reviewing, I missed this while I was OOO.

PR #185, opened 2022-08-24
Migrate filtering and data requests into business logic layer

tjennison-google:
This migrates the core filtering and data (entity instance) request
functionality into the business logic layer. This is the last (and
largest) major piece of work required to untagle the UI from the API.
This entails:
* Creating a local filter abstraction that operates in UI logical terms.
  These are converted to underlying backend types when making requests.
  Handling of empty/null filters has been streamlined.
* Adding methods to fetch entity instances and SQL queries from the
  backend, refactoring some of the existing code to be shared.
* Updating the datasets page to work in terms of the new filters,
  centralizing the last bespoke entity instance handling.
* Changing the criteria plugins to use the new filters.
* Updating the prepackaged concept sets to use the new filters.
* Enabling the primary entity and occurrences to be handled
  interchangeably in some cases.
* Fixing incorrect key names in the underlay defintions.

The overview pages are in a temporary state pending the addition of the
count API to the layer.

### DIFF
--- a/ui/src/app.tsx
+++ b/ui/src/app.tsx
@@ -1,5 +1,6 @@
 import { EntitiesApiContext, UnderlaysApiContext } from "apiContext";
 import Loading from "components/loading";
+import { FilterType } from "data/filter";
 import { useAsyncWithApi } from "errors";
 import { useAppDispatch } from "hooks";
 import { enableMapSet } from "immer";
@@ -8,7 +9,6 @@ import { useCallback, useContext } from "react";
 import { HashRouter } from "react-router-dom";
 import { AppRouter } from "router";
 import { LoadingUserData } from "storage/storage";
-import * as tanagra from "tanagra-api";
 import { setUnderlays } from "underlaysSlice";
 import "./app.css";
 
@@ -20,30 +20,17 @@ const prepackagedConceptSets = [
   {
     id: "_demographics",
     name: "Demographics",
-    entity: "person",
+    occurrence: "",
   },
   {
     id: "_analgesics",
     name: "Analgesics",
-    entity: "ingredient_occurrence",
+    occurrence: "ingredient_occurrence",
     filter: {
-      relationshipFilter: {
-        outerVariable: "ingredient_occurrence",
-        newVariable: "concept",
-        newEntity: "ingredient",
-        filter: {
-          binaryFilter: {
-            attributeVariable: {
-              variable: "concept",
-              name: "concept_id",
-            },
-            operator: tanagra.BinaryFilterOperator.DescendantOfInclusive,
-            attributeValue: {
-              int64Val: 21604253,
-            },
-          },
-        },
-      },
+      type: FilterType.Classification,
+      occurrenceID: "ingredient_occurrence",
+      classificationID: "ingredient",
+      keys: [21604253],
     },
   },
 ];
--- a/ui/src/data/filter.ts
+++ b/ui/src/data/filter.ts
@@ -1,0 +1,87 @@
+import { isValid } from "util/valid";
+import { DataKey, DataValue } from "./configuration";
+
+export enum FilterType {
+  Unary = "UNARY",
+  Array = "ARRAY",
+  Classification = "CLASSIFICATION",
+  Attribute = "ATTRIBUTE",
+}
+
+type BaseFilter = {
+  type: FilterType;
+};
+
+export enum UnaryFilterOperator {
+  Not,
+}
+
+export type UnaryFilter = BaseFilter & {
+  operator: UnaryFilterOperator;
+  operand: Filter;
+};
+
+export function isUnaryFilter(filter: Filter): filter is UnaryFilter {
+  return filter.type == FilterType.Unary;
+}
+
+export type ArrayFilterOperator = {
+  // Currently only AND (all) and OR (1+) equivalent values are supported but
+  // this could allow for various XOR and x-of-y type operations in the future.
+  // As a convenience, both min and max being undefined is equivalent to an AND.
+  min?: number;
+  max?: number;
+};
+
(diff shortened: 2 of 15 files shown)


ui/src/data/filter.ts
```diff
@@ -0,0 +1,87 @@
...
+  operator: UnaryFilterOperator;
+  operand: Filter;
+};
+
+export function isUnaryFilter(filter: Filter): filter is UnaryFilter {
+  return filter.type == FilterType.Unary;
+}
+
+export type ArrayFilterOperator = {
+  // Currently only AND (all) and OR (1+) equivilent values are supported but
```
kemp-verily, line 29: typo (equivalent)

ui/src/data/source.tsx
```diff
@@ -439,3 +571,226 @@ function processEntitiesResponse(
...
+      .filter(isValid);
+    if (operands.length === 0) {
+      return null;
+    }
+
+    let operator = tanagra.ArrayFilterOperator.Or;
+    if (!isValid(filter.operator.min) && !isValid(filter.operator.max)) {
+      operator = tanagra.ArrayFilterOperator.And;
+    } else if (filter.operator.min !== 1 || isValid(filter.operator.max)) {
+      throw new Error("Only AND and OR equivalent operators are support.");
```
kemp-verily, line 640: support -> supported

ui/src/data/source.tsx
```diff
@@ -439,3 +571,226 @@ function processEntitiesResponse(
...
+
+  if (isArrayFilter(filter)) {
+    const operands = filter.operands
+      .map((o) => generateFilter(source, o, fromPrimary))
+      .filter(isValid);
+    if (operands.length === 0) {
+      return null;
+    }
+
+    let operator = tanagra.ArrayFilterOperator.Or;
```
kemp-verily, line 636: nit: consider extracting this so the value isn't dangling -- as an added bonus you can drop the else chain and just early return... will matter more as you add more operators potentially and the chain gets more complex.

kemp-verily (commented):
lg w/nits for react/typescript only

ui/src/data/source.tsx
tjennison-google: Done.

ui/src/data/filter.ts
tjennison-google: Done.

ui/src/data/source.tsx
tjennison-google: Done.

ui/src/app.tsx
```diff
@@ -20,30 +20,17 @@ const prepackagedConceptSets = [
...
-            attributeValue: {
-              int64Val: 21604253,
-            },
-          },
-        },
-      },
+      type: FilterType.Classification,
+      occurrenceID: "ingredient_occurrence",
+      classificationID: "ingredient",
+      keys: [21604253],
```
marikomedlock, line 33: Why do you need this specific ingredient concept id here?

ui/src/data/filter.ts
```diff
@@ -0,0 +1,87 @@
...
+  attribute: string;
+  values?: DataValue[];
+  ranges?: { min: number; max: number }[];
+};
+
+export function isAttributeFilter(filter: Filter): filter is AttributeFilter {
+  return filter.type == FilterType.Attribute;
+}
+
+export type ClassificationFilter = BaseFilter & {
```
marikomedlock, line 71: Is a `ClassificationFilter` equivalent to a `RelationshipFilter` in the API? If so, consider using the same terminology in the UI/API code.

ui/src/data/source.tsx
```diff
@@ -391,43 +538,29 @@ function processEntitiesResponse(
...
+      const data = processEntityInstance(
+        attributeID,
+        instance,
+        (key, value) => {
+          if (key === pathAttribute) {
+            if (value?.stringVal) {
+              // TODO(tjennison): There's no way to tell if the IDs should
+              // be numbers or strings and getting it wrong will cause
+              // queries using them to fail because the types don't match.
+              // Figure out a way to handle this generically.
```
marikomedlock, line 551: We should include the data type in the entity attributes definition -- I think that will address this. The indexing workflows need that anyway (at least longer-term, right now they're just assuming data types).

ui/src/data/source.tsx
```diff
@@ -256,6 +360,53 @@ export class BackendSource implements Source {
...
+    return {
+      entityVariable: entity.entity,
+      selectedAttributes: requestedAttributes,
+      filter,
+      limit: 50,
+    };
+  }
+}
+
+function isInternalAttribute(attribute: string): boolean {
```
marikomedlock, line 408: Consider renaming this "isGeneratedAttribute" to be consistent with how we refer to these attributes in the API and indexing layers.

ui/src/app.tsx
tjennison-google: This is a prepackaged concept set and that's the `concept_id` for Analgesics. This was a temporary location for these and I can relocate them to the underlay at this point. I'll do that in another PR though.

ui/src/data/source.tsx
marikomedlock: Ah ok. Not this PR, but we should probably move that type of decoding logic to the backend also. I think the sql_function_wrapper and display value for attributes that I've added to the new underlay config version should help with some of this.

ui/src/data/filter.ts
marikomedlock: Gotcha. Can you add some comments inline describing which of these is the `ClassificationFilter`?

marikomedlock (approved):
LGTM provided inline documentation is added per comments.

ui/src/data/filter.ts
tjennison-google: Done.

PR #186, opened 2022-08-25
Add counts API to business logic layer

tjennison-google:


ui/src/overview.tsx
```diff
@@ -494,51 +465,36 @@ function DemographicCharts({ cohort }: DemographicChartsProps) {
...
-      // TODO(neelismail): Remove age handling once the API supports them.
-      if (
-        isValid(value) &&
-        property.key === "age" &&
-        typeof value === "number"
-      ) {
-        value = new Date().getFullYear() - value;
-      }
+    let value =
+      filterCountValue[property.key === "age" ? "year_of_birth" : property.key];
```
marikomedlock, line 475: Right now, the breakdown chart attributes are hard-coded in the UI. Should we make these part of the entity underlay definition (e.g. flag attributes as "forBreakdown")?

marikomedlock (approved):

ui/src/overview.tsx
tjennison-google: The only thing that's hardcoded is the `age => year_at_birth` conversion. The remaining attributes and the fact that `age` is one of them are already coming from the underlay.

PR #187, opened 2022-08-25
Use server side limits when querying

tjennison-google:


marikomedlock (approved):
LGTM. Seeing this, I'm thinking we ought to have a default limit value of 100, so that the UI doesn't need to specify this. Thoughts?

tjennison-google:
It doesn't matter too much either way but it never hurts to have useful defaults. I would remove the limits from here in that case.

PR #190, opened 2022-08-31
Move demographics charts into a drawer

tjennison-google:
This paves the way for making better use of the available space for editing.

### DIFF
--- a/ui/src/overview.tsx
+++ b/ui/src/overview.tsx
@@ -1,3 +1,5 @@
+import BarChartIcon from "@mui/icons-material/BarChart";
+import ChevronRightIcon from "@mui/icons-material/ChevronRight";
 import ExpandMoreIcon from "@mui/icons-material/ExpandMore";
 import MoreVertIcon from "@mui/icons-material/MoreVert";
 import Accordion from "@mui/material/Accordion";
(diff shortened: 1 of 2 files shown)


ui/src/overview.tsx
```diff
@@ -50,17 +55,77 @@ import {
...
+    duration: theme.transitions.duration.leavingScreen,
+  }),
+  overflowX: "hidden",
+  width: `calc(${theme.spacing(7)} + 1px)`,
+  [theme.breakpoints.up("sm")]: {
+    width: `calc(${theme.spacing(8)} + 1px)`,
+  },
+});
+
+const Drawer = styled(MuiDrawer, {
```
cbookg, line 82: nit: I would normally keep the default import and define this something else (but up to you)

ui/src/overview.tsx
```diff
@@ -612,6 +678,10 @@ function DemographicCharts({ cohort }: DemographicChartsProps) {
     return value.length > 15 ? value.substr(0, 15).concat("…") : value;
   };
 
+  if (!open) {
```
cbookg, line 681: it seems weird to pass this into this component.  shouldn't you just not include this component if your drawer is not open? (as i noted above)

ui/src/overview.tsx
```diff
@@ -50,17 +55,77 @@ import {
...
+            <Toolbar />
+            <Box sx={{ float: "right" }}>
+              <IconButton
+                size="large"
+                onClick={() => setShowDemographics(!showDemographics)}
+              >
+                {showDemographics ? <ChevronRightIcon /> : <BarChartIcon />}
+              </IconButton>
+            </Box>
+            <DemographicCharts cohort={cohort} open={showDemographics} />
```
cbookg, line 125: {showDemographics && <DemographicsChar cohort={cohort} />}

instead of passing in the prop to return null

cbookg (approved):
looks good, except for the one note on conditional rendering

PR #191, opened 2022-09-06
Migrate to react-router v6

tjennison-google:
The UI needs to support nested routes, where only part of the page
changes for routes below a certain parent. This could be handled
manually like the current parent URL handling, but v6 handles nested
routes and parent URLs automatically.

Changes:
* Routes are specified using a nested layout.
* Navigation uses relative paths instead of absolute paths.
* history.push() changes to navigate().


ui/src/router.tsx
```diff
@@ -4,90 +4,68 @@ import Edit from "edit";
...
-  validate: boolean,
-  params: Record<string, string>
-) {
-  let count = 0;
-  const url = path?.replaceAll(/:(\w+)/g, (_, match) => {
-    count = params[match] && count >= 0 ? count + 1 : -1;
-    return params[match];
-  });
-  return !validate || count == Object.keys(params).length ? url : undefined;
+export function criteriaURL(groupId: string, criteriaId: string) {
```
marikomedlock, line 55: Is `groupId` the criteria type (e.g. condition, device)? An example input here as a comment would be helpful.

marikomedlock (approved):

ui/src/router.tsx
tjennison-google: Group IDs are random strings that identify the included and excluded groups of criteria. IDs are used instead of indexes or names to more easily support reordering, renaming, etc. The overall data structure is in `cohort.ts`. I don't think an example of random IDs would be particularly enlightening here.

kemp-verily (approved):
react changes lgtm

PR #192, opened 2022-09-07
Centralize editing within a single main view

tjennison-google:
This provides a more cohesive user experience and makes better use of the available space.

* Moves cohort groups to a single left pane, allowing consistent access while editing items in the central pane.
* Adds a group editing page with more space to edit groups and their criteria. This removes the need for a separate overview page and also the need for an editing drawer for modifiers.
* Deleted concept.test.tsx. It was fragile and constantly breaking on unrelated changes. Now that the business logic has been separated out was providing minimal benefit anyways and should be replaced with UI tests.


ui/src/criteria/attribute.tsx
```diff
@@ -323,7 +323,7 @@ function AttributeDetails(props: AttributeDetailsProps) {
     return (
       <>
         {props.data.selected.map(({ value, name }) => (
-          <Stack direction="row" alignItems="baseline" key={Date.now()}>
```
marikomedlock, line 326: Why the change from Date -> String?

tjennison-google:
Some screenshots of the UI:
image
image

There are still more changes to come, like removing the duplicated elements, but UI is functional in the new format at this point.

ui/src/criteria/attribute.tsx
tjennison-google: This was effectively using a random key before, which worked because the list was never visible while being edited. Now the keys actually need to stable of duplicate UI elements can be created.

marikomedlock (approved):
Thanks for the screenshots, that's really helpful.

PR #194, opened 2022-09-08
Convert concept criteria to single selection

tjennison-google:
* This will allow modifiers to be applied in a consistent fashion, particularly for Labs & Measurements that have independent values.
* This paves the way for improved criteria naming because each criteria will only need to display a single value.


kemp-verily (commented):
lg for react

marikomedlock (approved):

PR #195, opened 2022-09-09
Separate criteria plugin config and data

tjennison-google:
Since the config was stored with the data, loading saved criteria would continue to used outdated config. This caused issues like not displaying new columns added to the config. Instead, only the id of the config is stored in the criteria and the current version is passed into the plugins.

This also merges the plugin specific config into the same structure as the rest of the config. The separation meant plugins that wanted base config and custom config either had to store them separately or cast the custom data at every use.

### DIFF
--- a/api/src/main/resources/underlays/aou_synthetic.yaml
+++ b/api/src/main/resources/underlays/aou_synthetic.yaml
@@ -424,116 +424,117 @@ uiConfiguration: >-
     },
     "criteriaConfigs": [{
       "type":"concept",
+      "id": "tanagra-conditions",
       "title":"Conditions",
       "defaultName":"Contains Conditions Codes",
-      "plugin": {
-        "columns": [
-          {"key":"concept_name","width":"100%","title":"Concept Name"},
-          {"key":"concept_id","width":120,"title":"Concept ID"},
-          {"key":"standard_concept","width":180,"title":"SourceStandard"},
-          {"key":"vocabulary_id","width":120,"title":"Vocab"},
-          {"key":"concept_code","width":120,"title":"Code"},
-          {"key":"person_count","width":120,"title":"Roll-up Count"}
-        ],
-        "hierarchyColumns": [
-          {"key": "concept_name", "width": "100%", "title": "Condition"},
-          {"key": "concept_id", "width": 120, "title": "Concept ID"},
-          {"key":"person_count","width":120,"title":"Roll-up Count"}
-        ],
-        "occurrence": "condition_occurrence",
-        "classification": "condition"
-      }
+      "columns": [
+        {"key":"concept_name","width":"100%","title":"Concept Name"},
+        {"key":"concept_id","width":120,"title":"Concept ID"},
+        {"key":"standard_concept","width":180,"title":"SourceStandard"},
+        {"key":"vocabulary_id","width":120,"title":"Vocab"},
+        {"key":"concept_code","width":120,"title":"Code"},
+        {"key":"person_count","width":120,"title":"Roll-up Count"}
+      ],
+      "hierarchyColumns": [
+        {"key": "concept_name", "width": "100%", "title": "Condition"},
+        {"key": "concept_id", "width": 120, "title": "Concept ID"},
+        {"key":"person_count","width":120,"title":"Roll-up Count"}
+      ],
+      "occurrence": "condition_occurrence",
+      "classification": "condition"
     },
     {
       "type":"concept",
+      "id": "tanagra-procedures",
       "title":"Procedures",
       "defaultName":"Contains Procedures Codes",
-      "plugin":{
-        "columns": [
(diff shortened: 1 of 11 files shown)


api/src/main/resources/underlays/aou_synthetic.yaml
```diff
@@ -424,116 +424,117 @@ uiConfiguration: >-
...
+        {"key":"concept_code","width":120,"title":"Code"},
+        {"key":"person_count","width":120,"title":"Roll-up Count"}
+      ],
+      "hierarchyColumns": [
+        {"key": "concept_name", "width": "100%", "title": "Condition"},
+        {"key": "concept_id", "width": 120, "title": "Concept ID"},
+        {"key":"person_count","width":120,"title":"Roll-up Count"}
+      ],
+      "occurrence": "condition_occurrence",
+      "classification": "condition"
```
marikomedlock, line 444: Do these both correspond to the entity names?

api/src/main/resources/underlays/aou_synthetic.yaml
```diff
@@ -424,116 +424,117 @@ uiConfiguration: >-
...
-        ],
-        "hierarchyColumns": [
-          {"key": "concept_name", "width": "100%", "title": "Condition"},
-          {"key": "concept_id", "width": 120, "title": "Concept ID"},
-          {"key":"person_count","width":120,"title":"Roll-up Count"}
-        ],
-        "occurrence": "condition_occurrence",
-        "classification": "condition"
-      }
+      "columns": [
```
marikomedlock, line 430: Not for this PR, but should we be generating this from the list of entity attributes in the underlay? That might require additional properties of each attribute (e.g. display name, display width), but then the info would be inline instead of in a separate block.

api/src/main/resources/underlays/aou_synthetic.yaml
```diff
@@ -424,116 +424,117 @@ uiConfiguration: >-
     },
     "criteriaConfigs": [{
       "type":"concept",
+      "id": "tanagra-conditions",
```
marikomedlock, line 427: Just curious, why the "tanagra-" prefix?

ui/src/storage/storage.tsx
```diff
@@ -32,45 +30,58 @@ export const storeUserData: Middleware<unknown, RootState> =
...
 
-        // TODO(tjennison): Handle backwards compatability when we're closer to
-        // launch.
-        if (data.version != currentVersion) {
-          data.cohorts = [];
-          data.conceptSets = [];
-        }
+  // TODO(tjennison): Handle backwards compatability when we're closer to
+  // launch.
+  if (data.version != currentVersion) {
```
marikomedlock, line 54: This is checking if the cohort serialization is out of date with the current UI, yes? If the UI will always fetch these by calling core Tanagra endpoints, we should probably also include a check there.

marikomedlock (approved):

kemp-verily (commented):
lg for react/typescript

api/src/main/resources/underlays/aou_synthetic.yaml
tjennison-google: I've generally captured the underlying entities separately from the the concepts that use them. For occurrences, that's probably always a direct mapping that can be merged into the entity definitions. For classifications, they currently all map directly to entities but I think that won't always be the case. In particular I'm thinking of how the source/standard trees work whether they are two separate ways to classify entities like conditions.

api/src/main/resources/underlays/aou_synthetic.yaml
tjennison-google: I'm not sure about this. One example where that doesn't work would be if there was more than one way to view an entity. Perhaps some elements can be inlines and some would remain separate?

api/src/main/resources/underlays/aou_synthetic.yaml
tjennison-google: Plugin criteria ids would live in the same namespace so I wanted to avoid collisions on such common names.

ui/src/storage/storage.tsx
tjennison-google: In theory it makes sense for that to live close to the data but we'll have to think through how all of the cases might work. Some examples:
- Can there be a divergence in version between the backend and UI?
- What about import?
- Could we need to display UI during an update?

marikomedlock: Sounds good, yeah we'll need to walk through difference scenarios (low priority for now). Tentatively, I'd hope we could always ship UI+backend together so we wouldn't need to handle different cohort schema versions.

api/src/main/resources/underlays/aou_synthetic.yaml
marikomedlock: OK, let's revisit post-underlay overhaul.

PR #201, opened 2022-09-15
Remove custom criteria naming code and remove remaining references

tjennison-google:
This also improves concept set naming.

### DIFF
--- a/ui/src/cohort.ts
+++ b/ui/src/cohort.ts
@@ -61,6 +61,14 @@ export type DisplayDetails = {
   additionalText?: string[];
 };
 
+export function getCriteriaTitle<DataType>(
+  criteria: tanagra.Criteria,
+  plugin?: CriteriaPlugin<DataType>
+) {
+  const p = plugin ?? getCriteriaPlugin(criteria);
+  return `${criteria.config.title}: ${p.displayDetails().title}`;
+}
+
 // registerCriteriaPlugin is a decorator that allows criteria to automatically
 // register with the app simply by importing them.
 export function registerCriteriaPlugin(
@@ -83,7 +91,6 @@ export function createCriteria(
   return {
     id: generateId(),
     type: config.type,
-    name: config.defaultName,
     data: entry.initializeData(source, config),
     config: config,
   };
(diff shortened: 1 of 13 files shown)


ui/src/cohort.ts
```diff
@@ -61,6 +61,14 @@ export type DisplayDetails = {
   additionalText?: string[];
 };
 
+export function criteriaTitle<DataType>(
```
kemp-verily, line 64: optional, but i would use a verb prefix here (eg `get`)

kemp-verily (commented):
lg for react/typescript with a nit

ui/src/cohort.ts
tjennison-google: Done.

marikomedlock (commented):
What's the reason behind removing the ability to name criteria groups and concept sets? Code changes look fine.

tjennison-google:
> What's the reason behind removing the ability to name criteria groups and concept sets? Code changes look fine.

I'm moving the UI to single select instances within a criteria. Each criteria can then be automatically named based on the actual concept that is selected so there isn't a need to override it like with "Contains Condition Codes".

marikomedlock:
> I'm moving the UI to single select instances within a criteria. Each criteria can then be automatically named based on the actual concept that is selected so there isn't a need to override it like with "Contains Condition Codes".

That makes sense for single/binary filters (condition=diabetes). Could the generated names get unwieldy for array filters (condition=diabetes and ethnicity=japanese and ...)?

tjennison-google:
> > I'm moving the UI to single select instances within a criteria. Each criteria can then be automatically named based on the actual concept that is selected so there isn't a need to override it like with "Contains Condition Codes".
> 
> That makes sense for single/binary filters (condition=diabetes). Could the generated names get unwieldy for array filters (condition=diabetes and ethnicity=japanese and ...)?

That would be groups, which can still be renamed. This is just the case for `condition=(diabetes or cancer)` which would instead be `condition=diabetes or condition=cancer`. I need to make a better habit of including screenshots in these PRs.

image

marikomedlock (approved):
nit: In the PR description, do you mean "removes" instead of "improves"?

tjennison-google:
> nit: In the PR description, do you mean "removes" instead of "improves"?

I did actually mean "improves". They had bad default names and no ability to rename them. I've edited the description to clarify.

PR #202, opened 2022-09-16
Add separate page for creating criteria

tjennison-google:
* Instead of the "Add Criteria" button opening a menu, it navigates to a separate page. The options from the menu are present on the new page along with a search bar.
* Support searching across criteria types and display the returned criteria merged together and sorted by roll up count. Criteria can be added directly from the search results.
* Allow customization of individual columns of a TreeGrid to allow the criteria type column to be before the name.

### DIFF
--- a/ui/src/addCriteria.tsx
+++ b/ui/src/addCriteria.tsx
@@ -1,0 +1,202 @@
+import Box from "@mui/material/Box";
+import Button from "@mui/material/Button";
+import Chip from "@mui/material/Chip";
+import Stack from "@mui/material/Stack";
+import Typography from "@mui/material/Typography";
+import ActionBar from "actionBar";
+import { insertCriteria } from "cohortsSlice";
+import Loading from "components/loading";
+import { Search } from "components/search";
+import {
+  TreeGrid,
+  TreeGridData,
+  TreeGridId,
+  TreeGridItem,
+} from "components/treegrid";
+import { DataEntry, DataKey } from "data/configuration";
+import { MergedDataEntry, useSource } from "data/source";
+import { useAsyncWithApi } from "errors";
+import { useAppDispatch, useCohortAndGroup, useUnderlay } from "hooks";
+import { useCallback, useMemo, useState } from "react";
+import { useNavigate } from "react-router-dom";
+import { cohortURL, criteriaURL } from "router";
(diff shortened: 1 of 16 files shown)


tjennison-google:
image
image

marikomedlock (approved):

ui/src/addCriteria.tsx
```diff
@@ -0,0 +1,202 @@
...
+    [source, cohort.id, group.id]
+  );
+
+  const criteriaConfigMap = useMemo(
+    () => new Map(criteriaConfigs.map((cc) => [cc.id, cc])),
+    [criteriaConfigs]
+  );
+
+  const search = useCallback(async () => {
+    // This prevents prevents briefly showing the empty results from the empty
```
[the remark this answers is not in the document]
tjennison-google: Fixed.

ui/src/addCriteria.tsx
```diff
@@ -0,0 +1,202 @@
+import Box from "@mui/material/Box";
```
kemp-verily, line 1: optional, but we tend to use
```
import { Box, Button, Chip } from "@mui/material"
```
style imports in Axon (Enterprise Terra UI)

you seem to have a mix here

ui/src/addCriteria.tsx
```diff
@@ -0,0 +1,202 @@
...
+            ),
+          },
+          entry: entry,
+        };
+        data[key] = item;
+      });
+    }
+
+    setData(data);
+  }, [source, query]);
```
kemp-verily, line 135: should this not also depend on `criteriaConfigs` (and possibly `criteriaConfigMap`, though I think that's redundant)?

ui/src/addCriteria.tsx
tjennison-google: We've gone with the individual imports based on the docs saying in can decrease startup time in development. It looks like there's another option to enable that more concise syntax using a Babel plugin so we could explore that at some point.

For now, I've changed the one instance using multiple imports to be consistent the others.

ui/src/addCriteria.tsx
tjennison-google: Done. It doesn't actually matter because the underlays are never changed without reloading but it's good to be correct to prevent future problems.

PR #205, opened 2022-09-20
Overhaul styling (fonts, colors, spacing, layout)

tjennison-google:
Some notable changes in the code:
* Added an MUI theme to control the overall style. This is a much cleaner interface for customizing the MUI components than using CSS. Future work will remove all built in style from app.css and set up a method for plugins to override the theme.
* Fixed ellipses in the TreeGrid.
* Tightened some of the column sizes to better accomodate longer concept names and reasonably support 1280 pixel width displays.
* Improve horizontal scrolling of tables by making the scroll bar stick to the bottom of the screen instead of the bottom of the table which might be way below the bottom of the screen.
* Removed existing inline concept editing. It was unnecessary with the change to single select. In the future, this is where modifier controls will go.

### DIFF
--- a/ui/src/criteria/concept.tsx
+++ b/ui/src/criteria/concept.tsx
@@ -1,8 +1,8 @@
 import AccountTreeIcon from "@mui/icons-material/AccountTree";
+import Box from "@mui/material/Box";
 import Button from "@mui/material/Button";
 import IconButton from "@mui/material/IconButton";
 import Stack from "@mui/material/Stack";
-import Typography from "@mui/material/Typography";
 import { CriteriaPlugin, registerCriteriaPlugin } from "cohort";
 import Checkbox from "components/checkbox";
 import Loading from "components/loading";
@@ -92,7 +92,7 @@ class _ implements CriteriaPlugin<Data> {
   }
 
   renderInline() {
-    return <ConceptInline data={this.data} />;
+    return <ConceptInline />;
   }
 
   displayDetails() {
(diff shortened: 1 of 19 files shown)


tjennison-google:
image
image
image

ui/src/datasets.tsx
```diff
@@ -133,6 +133,10 @@ export function Datasets() {
   });
 
   const allAttributesChecked = () => {
+    if (conceptSetOccurrences.length === 0) {
```
kemp-verily, line 136: behaviour change?

ui/src/criteria/concept.tsx
```diff
@@ -92,7 +92,7 @@ class _ implements CriteriaPlugin<Data> {
   }
 
   renderInline() {
-    return <ConceptInline data={this.data} />;
+    return <ConceptInline />;
```
kemp-verily, line 95: unrelated change?

kemp-verily (approved):
looks like a huge improvement!

ui/src/datasets.tsx
tjennison-google: This just fixes the default look of the checkbox when there are no concept sets selected vs. when all of the values of a concept set are unselected.

ui/src/criteria/concept.tsx
tjennison-google: `props` became an unused variable with the removal of the concept in the inline display so I had to remove it entirely.

marikomedlock (approved):
Screenshots look really great.

ui/src/criteria/concept.tsx
```diff
@@ -365,29 +374,12 @@ function ConceptEdit(props: ConceptEditProps) {
...
-    </>
+    </Box>
   );
 }
 
-type ConceptInlineProps = {
-  data: Data;
-};
-
-function ConceptInline(props: ConceptInlineProps) {
```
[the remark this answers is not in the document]
tjennison-google: Yes, though there's currently nothing to actually edit there. Eventually that's where modifiers will go. Once I removed the current content I ended up having to remove most of it entirely because it became a chain of unused variables.

ui/src/theme.tsx
```diff
@@ -0,0 +1,185 @@
+import { createTheme } from "@mui/material/styles";
+
+export const theme = createTheme({
```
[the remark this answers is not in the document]
tjennison-google: I think it will be an option for a javascript plugin that either wraps or replaces the argument to createTheme. Some high level things like logos and primary colors might make sense as part of the underlay or other config but full styling would require too much customization I think. I don't want to be recreating CSS in the underlay. Obviously anything related to the underlying data will still remain in the underlay though.

PR #206, opened 2022-09-21
Improve group filtering UI

tjennison-google:
* Reorganize the group filter definition. The filter is moved into a separate object to group its parameters. Repurpose the "kind" field to indicate the overall kind (any, all, in future temporal, x-of-y, etc.) and move the existing exclusion logic to a boolean.
* Change UI to use badges instead of icons. This also clarifies the relationship between individual criteria.
* Rename groups to requirements to indicate that they're all required to match.

### DIFF
--- a/ui/src/overview.tsx
+++ b/ui/src/overview.tsx
@@ -1,9 +1,8 @@
 import BarChartIcon from "@mui/icons-material/BarChart";
-import CheckIcon from "@mui/icons-material/Check";
 import ChevronRightIcon from "@mui/icons-material/ChevronRight";
-import ClearIcon from "@mui/icons-material/Clear";
 import Box from "@mui/material/Box";
 import Button from "@mui/material/Button";
+import Chip from "@mui/material/Chip";
 import Drawer from "@mui/material/Drawer";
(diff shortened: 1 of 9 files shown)


ui/src/overview.tsx
```diff
@@ -184,7 +186,7 @@ function AddGroupButton() {
   return (
     <>
       <Button onClick={onAddGroup} variant="contained" className="add-group">
-        Add Group
+        Add requirement
```
marikomedlock, line 189: I like this vocabulary change group->requirement. Good idea.

ui/src/overview.tsx
```diff
@@ -296,7 +300,10 @@ function ParticipantCriteria(props: {
       groups: [
         {
           id: props.group.id,
-          kind: tanagra.GroupKindEnum.Included,
+          filter: {
```
marikomedlock, line 303: Is this the default group filter?

marikomedlock (approved):

ui/src/overview.tsx
tjennison-google: Not quite but that would be a bit simpler and get the same result. Done.

PR #207, opened 2022-09-21
Show small loading spinner immediately

tjennison-google:
Since the small spinner is used inline unlike the normal one, delaying its appearance causes elements to shift around it until it appears.

### DIFF
--- a/ui/src/components/loading.tsx
+++ b/ui/src/components/loading.tsx
@@ -27,6 +27,13 @@ export default function Loading(props: Props) {
       return;
     }
 
+    // Show the small spinner immediately since it's used inline and the delay
+    // causes extra shifting of the surrounding elements.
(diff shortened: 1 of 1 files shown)


tjennison-google:
> What is the small vs large spinner being used for?

The large spinner is for page wide loading, like concepts or the demographics charts. The small one is used inline for the criteria and group counts. For the large spinners, the pages are otherwise empty so the delayed loading means just less flashing of the spinner if the loads are quick. For the small spinners, since they're inline, it causes extra shifting of the text and surrounding elements.

marikomedlock (approved):
Thanks for the explanation. Can you add that in a comment somewhere for future reference?

tjennison-google:
> Thanks for the explanation. Can you add that in a comment somewhere for future reference?

Done.

PR #210, opened 2022-09-22
Clean up chart formatting

tjennison-google:
Also overlay the close button on top of the charts to avoid either having to integrate it into the chart layout or waasting all of the space below it.


tjennison-google:
image

marikomedlock (approved):

PR #212, opened 2022-09-22
Don't update group counts on all cohort changes

tjennison-google:


ui/src/overview.tsx
```diff
@@ -226,7 +226,7 @@ function ParticipantsGroup(props: {
     }
 
     return (await source.filterCount(filter))[0].count;
-  }, [underlay, cohort]);
+  }, [underlay, cohort.id, cohort.name, cohort.underlayName, props.group]);
```
marikomedlock, line 229: So if any of these properties change, then you don't update the counts? Why do you need both `underlay` and `cohort.underlayName`?

tjennison-google: Good catch. There isn't actually a dependency on `underlay` so it shouldn't be there.

marikomedlock (approved):

PR #213, opened 2022-09-22
Fix range spacing and remove incorrect error tooltips

tjennison-google:
In addition to changing the amount by which the input arrows adjust the value, the "step" property also indicates that only step values are actually valid. These behaviors can't be separated without significant custom code so for now the step has just been removed.

### DIFF
--- a/ui/src/criteria/attribute.tsx
+++ b/ui/src/criteria/attribute.tsx
@@ -193,7 +193,7 @@ function AttributeSlider(props: SliderProps) {
 
   return (
     <Box sx={{ width: "30%", minWidth: 400, mt: 0.5 }}>
-      <Grid container spacing={1} direction="row">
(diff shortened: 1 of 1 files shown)


marikomedlock (commented):
"without significant custom code" refers to code we'd have to pull from the underlay to render this attribute?

tjennison-google:
> "without significant custom code" refers to code we'd have to pull from the underlay to render this attribute?

If there's a big range, like 1000, the arrows only changing by 1 at a time is too small to be very effective. Setting `step` to 10 will cause the arrows to adjust up or down by 10 at a time but then the input will say that anything not a multiple of 10 is invalid. There's no way to separate those behaviors with a bunch of custom code. It only shows up as a tooltip but I think living with arrow adjustments at 1 is better than showing confusing messages.

marikomedlock (approved):

PR #217, opened 2022-09-23
Simplify top level layout using grids

tjennison-google:


kemp-verily (commented):
did you decide against using the `<Grid>` component?  i like it for flexible layout over the core CSS tooling like this (but either is fine)

lgtm if you don't want to use `<Grid>`

tjennison-google:
> did you decide against using the `<Grid>` component? i like it for flexible layout over the core CSS tooling like this (but either is fine)
> 
> lgtm if you don't want to use `<Grid>`

This uses some of the CSS grid features for sizing that the MUI grid doesn't support because it actually uses flexbox under the hood.

marikomedlock (approved):

PR #219, opened 2022-09-23
Add placeholders to empty lists

tjennison-google:


tjennison-google:
image

tjennison-google:
The placeholder image is from the design guide but perhaps not what we want long term...

ui/src/datasets.tsx
```diff
@@ -147,6 +154,14 @@ export function Datasets() {
     return true;
   };
 
+  const cohorts = useMemo(
+    () =>
+      unfilteredCohorts.filter(
```
marikomedlock, line 159: `unfilteredCohorts` is all cohorts the user has access to, right? Would you expect this to go away once the UI is fetching cohorts from the backend?

marikomedlock (approved):

ui/src/datasets.tsx
tjennison-google: No, this filtering is based on the underlay. It might still go away once the UI is fetching cohorts from the backed though.

marikomedlock: Can you clarify what this logic is doing then? I thought `filter` was just getting the subset of cohorts that match the current underlay name.

tjennison-google: Yeah, that is what it's doing. I guess I mistook your question for being about access control when this is just because the UI only operates on one cohort at a time.

ui/src/criteria/concept.tsx
```diff
@@ -274,104 +275,112 @@ function ConceptEdit(props: ConceptEditProps) {
...
-              name: !!name ? String(name) : "",
-            };
-
-            if (props.config.multiSelect) {
-              const index = props.data.selected.findIndex(
-                (sel) => item.node.data.key === sel.key
-              );
+        {!data.root?.children?.length ? (
+          <Empty
+            minHeight="300px"
```
kemp-verily, line 280: optional: maybe make `empty.png` the default?

tjennison-google: That doesn't really end up better because we still need to be able to specify whether the image is displayed or not with either an extra parameter or extra state for this one. I think that outweighs specifying the image in a couple places, plus we we might show different images in the future.

PR #222, opened 2022-09-26
Indexing job for writing text search string.

marikomedlock:
- Added the indexing job for writing text search string - id pairs for entities with a text search mapping defined.
- A future improvement is to write this string to a new field in the denormalized entity instances index table.

### DIFF
--- a/api/src/main/java/bio/terra/tanagra/indexing/job/BuildTextSearchStrings.java
+++ b/api/src/main/java/bio/terra/tanagra/indexing/job/BuildTextSearchStrings.java
@@ -1,0 +1,84 @@
+package bio.terra.tanagra.indexing.job;
+
+import bio.terra.tanagra.exception.InvalidConfigException;
+import bio.terra.tanagra.indexing.EntityJob;
+import bio.terra.tanagra.query.Query;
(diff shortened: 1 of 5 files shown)


api/src/main/java/bio/terra/tanagra/indexing/job/BuildTextSearchStrings.java
```diff
@@ -0,0 +1,84 @@
...
+  public void run() {
+    run(false);
+  }
+
+  private void run(boolean isDryRun) {
+    Query selectIdTextPairs = getEntity().getSourceDataMapping().queryTextSearchString();
+    String sql = selectIdTextPairs.renderSQL();
+    LOGGER.info("select id-text pairs SQL: {}", sql);
+
+    BigQueryDataset outputBQDataset = getOutputDataPointer();
```
tjennison-google, line 44: Can some of the code in these functions be refactored? It seems like they were very similar (identical?) across the last three PRs.

marikomedlock: Good idea. Moved chunks of this code into the base class.

tjennison-google (approved):

PR #224, opened 2022-09-27
Indexing job for writing ancestor-descendant id pairs.

marikomedlock:
- Added the indexing job for writing ancestor-descendant id pairs for entities with hierarchies.
  - This is the first indexing job class that uses Apache Beam, so this PR includes adding the Beam dependencies to the `api` sub-project `build.gradle` file.
  - The new `WriteAncestorDescendantIdPairs` class is very similar to the deleted `FlattenHierarchy` class. It's repackaged to fit the indexing job interface.
- Moved the pagination proto definition into the `lib` Gradle sub-project along with the other proto definitions. This allowed me to remove the proto Gradle plugin from the `api` `build.gradle` file, so it's only used in a single place now. The Java classes generated from the pagination proto definitions have not changed.
- Fixed a bug reading in raw SQL from a multi-line file.

Coming next:
- Handle the path and rollup count Beam jobs.
- Remove the `workflow` sub-project completely once we can trigger all indexing jobs from the new underlay config in the `api` project.


api/src/main/java/bio/terra/tanagra/indexing/job/WriteAncestorDescendantIdPairs.java
```diff
@@ -0,0 +1,193 @@
...
+                "read parent-child query result",
+                BigQueryIO.readTableRows()
+                    .fromQuery(sql)
+                    .withMethod(BigQueryIO.TypedRead.Method.EXPORT)
+                    .usingStandardSql())
+            .apply(
+                MapElements.into(
+                        TypeDescriptors.kvs(TypeDescriptors.longs(), TypeDescriptors.longs()))
+                    .via(WriteAncestorDescendantIdPairs::relationshipRowToKV));
+    // TODO: Allow overriding the default max hierarchy depth.
```
tarekmamdouh, line 139: just curious, are the TODO tracked by tasks somewhere, or you will be tackling them in later PRs?

tarekmamdouh (approved):

api/src/main/java/bio/terra/tanagra/indexing/job/WriteAncestorDescendantIdPairs.java
marikomedlock: There's no task tracking yet, because it's such a small team (backend is just me). I'm planning to tackle them in later PRs, but not urgently. We may start using JIRA once we onboard more engineers -- in that case, I'll start noting the ticket number in the comment as well.

PR #225, opened 2022-09-28
Indexing job for writing num_children - path table.

marikomedlock:
- Added the indexing job for writing `num_children` and `path` to a separate table for entities with hierarchies.
  - This is the second indexing job class that uses Apache Beam, so this PR includes some refactoring of the shared Beam initialization code into the BQ indexing job base class.
  - The new `BuildNumChildrenAndPaths` class is very similar to the deleted `BuildPathsForHierarchy` class. It's repackaged to fit the indexing job interface.
- Changed the schema of the num_children-path index table to call the FK column "id" instead of "node", so it's clearer that this is the link to the main table of entity instances.
 
Coming next:
- Handle the rollup count Beam job.
- Remove the `workflow` sub-project completely once we can trigger all indexing jobs from the new underlay config in the `api` project.


api/src/main/java/bio/terra/tanagra/indexing/BigQueryIndexingJob.java
```diff
@@ -51,17 +81,37 @@ protected JobStatus checkTableExistenceForJobStatus(TablePointer outputTable) {
...
+      "--region=" + DEFAULT_REGION,
+      "--serviceAccount=" + getAppDefaultSAEmail(),
+    };
+    // TODO: Use PipelineOptionsFactory.create() instead of fromArgs().
+    // PipelineOptionsFactory.create() doesn't seem to call the default factory classes, while
+    // fromArgs() does.
+    return PipelineOptionsFactory.fromArgs(args).withValidation().as(BigQueryOptions.class);
+  }
+
+  protected String getAppDefaultSAEmail() {
```
tarekmamdouh, line 106: Would this method be used by other pipelines? Maybe extract it to its own service if that's the case. I don't feel it belongs in this class.

marikomedlock: All existing pipelines extend this base class. Specifically, it's needed to build the Dataflow options object for kicking off a Pipeline. We expect to add new indexing jobs later that e.g. run on Azure instead of BQ/Dataflow, and those would likely need a new base class. I'm not sure whether those would also need to pull the app default SA email, but if they do, then I think it would make sense to pull this out into a shared class at that point.

tarekmamdouh: should be fine for now I think, just noticed it's in the abstract BigQueryIndexingJob and not in the specific implementation

tarekmamdouh (approved):

PR #229, opened 2022-09-29
Documentation and aou_synthetic OMOP config files.

marikomedlock:
- Wrote documentation about the indexing process and the entity model. Reorganized the markdown files a bit and linked them from the top-level `README.md` in the `api` package.
- Defined all config files for the `aou_synthetic` underlay. I tested all indexing jobs successfully using these config files.
- Added an `INDEX_ALL` command to kick off all indexing jobs for an underlay, including all entities and entity groups.
- Handled the case where the primary table for an entity mapping is a raw SQL query. This is the case for the `ingredient` entity in the `aou_synthetic` underlay. Instead of pulling a table schema from BigQuery, we now pull a query schema.
- Added a few new binary filter operators to support some filters in the `aou_synthetic` config files: `NOT_EQUALS`, `IS NULL`, `IS NOT NULL`.


api/src/main/java/bio/terra/tanagra/utils/GoogleBigQuery.java
```diff
@@ -48,6 +49,20 @@ public GoogleBigQuery(GoogleCredentials credentials, String projectId) {
             .build()
             .getService();
     this.tableSchemasCache = new HashMap<>();
+    this.querySchemasCache = new HashMap<>();
+  }
+
+  public Schema getQuerySchemaWithCaching(String query) {
```
tarekmamdouh, line 55: this method is called in BigQueryDataset, which I assume (and I might be wrong) that it might be called from different threads but querySchemasCache is not thread safe.
If my assumption is correct, then querySchemasCache better be a ConcurrentHashMap

tarekmamdouh:
thank you for the documentation, it's very elaborate

tarekmamdouh (approved):

api/src/main/java/bio/terra/tanagra/utils/GoogleBigQuery.java
marikomedlock: Updated. It's not called from multiple threads now because the rudimentary CLI commands are always single-threaded. But in the future, I'd like to kickoff multiple indexing jobs in parallel to speed things up. You're right that I'd need a `ConcurrentHashMap` in that case, so this change will help for the future.

PR #230, opened 2022-09-30
Add CLEAN commands. Override default Dataflow job name.

marikomedlock:
- Added commands to cleanup the outputs of indexing jobs: `CLEAN_ENTITY`, `CLEAN_ENTITY_GROUP`, `CLEAN_ALL`. This is useful for debugging and testing.
- Run the indexing jobs in alphabetical order.
- Override the default Dataflow job name so the indexing job name and entity/group name are visible in the Cloud Console.

### DIFF
--- a/api/src/main/java/bio/terra/tanagra/indexing/BigQueryIndexingJob.java
+++ b/api/src/main/java/bio/terra/tanagra/indexing/BigQueryIndexingJob.java
@@ -15,10 +15,19 @@ import com.google.cloud.bigquery.TableId;
 import com.google.common.annotations.VisibleForTesting;
 import java.io.IOException;
 import java.util.Optional;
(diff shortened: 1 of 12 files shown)


marikomedlock:
> Do you need to worry about cancelling in-progress jobs when cleaning?

Yes, this is a problem and not just for cleaning. I haven't implemented any kind of context storage yet, so we're not saving the Dataflow job id, which means we can't check to see if there's already a job running. So you can kickoff the same job twice or clean while the first job is still running. Currently, you have to check the Dataflow jobs page in Cloud Console before running any indexing.

I hope to fix this in the future (e.g. write the job ids to disk as part of a CLI context file, or manage all this from service endpoints and store state in an internal db), but probably not until I get through all the SDD requirements.

PR #235, opened 2022-10-03
Remove visible references to "concepts"

tjennison-work:
* Prefer codes over IDs.
* Remove "concept" from titles.

### DIFF
--- a/api/src/main/resources/underlays/aou_synthetic.yaml
+++ b/api/src/main/resources/underlays/aou_synthetic.yaml
@@ -429,8 +429,7 @@ uiConfiguration: >-
       "conceptSet": true,
       "category": "Domains",
       "columns": [
-        {"key":"concept_name","width":"100%","title":"Concept name"},
-        {"key":"concept_id","width":100,"title":"Concept ID"},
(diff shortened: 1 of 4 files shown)


marikomedlock (commented):
Is removing the `concept_id` field a request from UX or VUMC? That's been useful to me when working with the AoU CB, though I'm certainly not a representative user.

tjennison-google:
> Is removing the `concept_id` field a request from UX or VUMC? That's been useful to me when working with the AoU CB, though I'm certainly not a representative user.

Neither, it just frees up more space for other columns. Why has it been useful? Just for debugging? Isn't it just an internal ID and the Code is the actual meaningful identifier?

marikomedlock:
> > Is removing the `concept_id` field a request from UX or VUMC? That's been useful to me when working with the AoU CB, though I'm certainly not a representative user.
> 
> Neither, it just frees up more space for other columns. Why has it been useful? Just for debugging? Isn't it just an internal ID and the Code is the actual meaningful identifier?

Yes for debugging and comparing Tanagra with the current AoU-CB when talking to VUMC folks. Maybe we can just check with them at our next checkin?

marikomedlock (approved):

PR #237, opened 2022-10-03
Rename "concept" criteria to "classification"

tjennison-google:
Replacing/renaming criteria will need to exist as part of versioning them but for now just globablly update the version.


marikomedlock (approved):
Does `criteria` include both concepts and demographics, and `classification` is just the concepts?

PR #241, opened 2022-10-05
Proper "back" behavior in classification criteria

tjennison-google:
* The in-app back button returns to the list view from the hierarchy view in classification criteria.
* The browser bar/forward buttons work when searching and across both views.

### DIFF
--- a/ui/src/actionBar.tsx
+++ b/ui/src/actionBar.tsx
@@ -5,7 +5,7 @@ import IconButton from "@mui/material/IconButton";
 import Toolbar from "@mui/material/Toolbar";
 import Typography from "@mui/material/Typography";
 import { useAppSelector } from "hooks";
-import * as React from "react";
+import { useEffect, useState } from "react";
 import { Link as RouterLink, useLocation, useParams } from "react-router-dom";
 import UndoRedo from "./components/UndoRedo";
 
@@ -20,6 +20,9 @@ export default function ActionBar(props: ActionBarProps) {
   );
 
   const location = useLocation();
+
+  const [backURL, set] = useState<string>();
+  setBackURL = set;
 
   return (
     <Box className="action-bar">
@@ -37,7 +40,7 @@ export default function ActionBar(props: ActionBarProps) {
             color="inherit"
             aria-label="back"
             component={RouterLink}
-            to={".."}
+            to={backURL ?? ".."}
             sx={{
               visibility: location.pathname === "/" ? "hidden" : "visible",
             }}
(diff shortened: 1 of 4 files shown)


kemp-verily (commented):
can you have @cbookg take a look?  he will probably have more useful feedback (he's already implemented some of this in the Axon UI)

ui/src/actionBar.tsx
```diff
@@ -21,6 +21,9 @@ export default function ActionBar(props: ActionBarProps) {
 
   const location = useLocation();
 
+  const [backURL, set] = useState<string | undefined>();
```
cbookg, line 24: you just need `<string>` here and the types will include undefined because you have no default value

ui/src/actionBar.tsx
```diff
@@ -66,3 +69,24 @@ export default function ActionBar(props: ActionBarProps) {
     </Box>
   );
 }
+
+let setBackURL: ((url?: string) => void) | undefined;
```
cbookg, line 73: i know you have a comment, but this should just be something the hook gets from the context.  any reason to not just do this?

cbookg:
> can you have @cbookg take a look? he will probably have more useful feedback (he's already implemented some of this in the Axon UI)

Not sure what you mean by this.  I don't think I've implemented anything similar.  If the goal here is to keep the search criteria in the query string such that you can go "back", this is something the TVC workspace list does *not* do but needs to be fixed as well.

ui/src/components/search.tsx
```diff
@@ -6,17 +6,35 @@ import InputAdornment from "@mui/material/InputAdornment";
...
 export type SearchProps = {
   placeholder?: string;
-  onSearch: (query: string) => void;
+  onSearch?: (query: string) => void;
+  initialValue?: string;
+  searchKey?: string;
 };
 
 export function Search(props: SearchProps) {
+  const setSearchParams = useSearchParams()[1];
```
cbookg, line 19: i had to look up this api to understand, this might be a more clear form:
const [_, setSearchParams]  = useSearchParams()

tjennison-google: Unfortunately it doesn't treat `_` specially and warns about it being unused.

ui/src/actionBar.tsx
tjennison-google: That will touch a bunch of other paces in the code, including refactoring the multiple action bar instances into a single location and also providing a way to set the title in each location.

ui/src/components/search.tsx
cbookg: From this site it looks like you can do it, but just dropping the underscore and having the comma:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

cbookg: under "ignorign some returned values"

tjennison-google: That does work. Done.

ui/src/criteria/classification.tsx
```diff
@@ -132,6 +132,37 @@ function keyForNode(node: ClassificationNode): DataKey {
   return key;
 }
 
+type SearchData = {
```
marikomedlock, line 135: - `SearchData.query` stores the text string that the user typed into the search box on the list view.
- `SearchData.hierarchy` stores the id that is selected in the hierarchy view, which is the row the user clicked on in the list view to get there.

Am I following this correctly? And if so, could you please add a comments about this for future reference.

tjennison-google: Done.

marikomedlock (approved):

PR #242, opened 2022-10-06
Underlays V2 endpoints.

marikomedlock:
New `/v2/underlays` endpoints to list and get an underlay. These endpoints use the new config files. I've left the existing `/underlays` endpoints as is, so as not to break the UI. Once the UI can be switched over to use all new endpoints, I'll go back and delete all the original ones.

### DIFF
--- a/api/src/main/java/bio/terra/tanagra/underlay/UnderlaysService.java
+++ b/api/src/main/java/bio/terra/tanagra/underlay/UnderlaysService.java
@@ -16,43 +16,35 @@ import org.springframework.stereotype.Component;
 /** Service to handle underlay operations. */
 @Component
 public class UnderlaysService {
-  private final UnderlayConfiguration underlayConfiguration;
-  private Map<String, Underlay> underlaysMap;
+  private final Map<String, Underlay> underlaysMap;
 
   @Autowired
   public UnderlaysService(UnderlayConfiguration underlayConfiguration) {
-    this.underlayConfiguration = underlayConfiguration;
+    // read in underlays from resource files
+    Map<String, Underlay> underlaysMapBuilder = new HashMap<>();
+    FileIO.setToReadResourceFiles();
+    for (String underlayFile : underlayConfiguration.getUnderlayFiles()) {
+      Path resourceConfigPath = Path.of("config").resolve(underlayFile);
+      FileIO.setInputParentDir(resourceConfigPath.getParent());
+      try {
+        Underlay underlay = Underlay.fromJSON(resourceConfigPath.getFileName().toString());
+        underlaysMapBuilder.put(underlay.getName(), underlay);
+      } catch (IOException ioEx) {
+        throw new SystemException(
+            "Error reading underlay file from resources: " + resourceConfigPath, ioEx);
+      }
+    }
+    this.underlaysMap = underlaysMapBuilder;
   }
 
   public List<Underlay> getUnderlays() {
-    return getUnderlaysMap().values().stream().collect(Collectors.toUnmodifiableList());
+    return underlaysMap.values().stream().collect(Collectors.toUnmodifiableList());
   }
 
   public Underlay getUnderlay(String name) {
-    if (!getUnderlaysMap().containsKey(name)) {
+    if (!underlaysMap.containsKey(name)) {
       throw new NotFoundException("Underlay not found: " + name);
     }
-    return getUnderlaysMap().get(name);
-  }
-
-  private Map<String, Underlay> getUnderlaysMap() {
-    // TODO: Only read in the underlay config files once at startup, instead of for every call.
-    if (underlaysMap == null) {
-      underlaysMap = new HashMap<>();
-      // read in underlays from resource files
-      FileIO.setToReadResourceFiles();
-      for (String underlayFile : underlayConfiguration.getUnderlayFiles()) {
-        Path resourceConfigPath = Path.of("config").resolve(underlayFile);
-        FileIO.setInputParentDir(resourceConfigPath.getParent());
-        try {
-          Underlay underlay = Underlay.fromJSON(resourceConfigPath.getFileName().toString());
-          underlaysMap.put(underlay.getName(), underlay);
-        } catch (IOException ioEx) {
-          throw new SystemException(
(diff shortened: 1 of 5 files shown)


api/src/main/resources/api/service_openapi.yaml
```diff
@@ -180,6 +180,43 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/UnderlayListV2"
+        500:
+          $ref: '#/components/responses/ServerError'
+
+  "/v2/underlays/{underlayName}":
+    parameters:
+      - $ref: "#/components/parameters/UnderlayNameV2"
```
tjennison-google, line 201: What is the point of having a reference here instead of including it inline? Should there be a new underlay component that uses it?

api/src/main/resources/api/service_openapi.yaml
```diff
@@ -876,3 +941,43 @@ components:
...
+      properties:
+        name:
+          description: Name
+          type: string
+        description:
+          description: Description
+          type: string
+        uiConfiguration:
+          description: UI configuration
+          type: string
```
tjennison-google, line 958: Can this be handled as generic object instead? The string was unavoidable when it had to pass through protos without actually being in the definition but it has significant downsides in that it can't be automatically formatted along with the rest of the underlay and errors aren't caught until it's actually parsed instead of at load time.

api/src/main/resources/api/service_openapi.yaml
```diff
@@ -876,3 +941,43 @@ components:
...
+
+    #########################UNDERLAY CONFIG V2 [START]#################################
+    UnderlayV2:
+      type: object
+      description: Underlay
+      properties:
+        name:
+          description: Name
+          type: string
+        description:
```
tjennison-google, line 953: Is this intended as a display name (e.g. `CMS SynPUF`) rather than a description (e.g. `Medicare Claims Synthetic Public Use Files (SynPUFs) were created to allow interested parties to gain familiarity using Medicare claims data while protecting beneficiary privacy.`)?

api/src/main/java/bio/terra/tanagra/underlay/UnderlaysService.java
```diff
@@ -0,0 +1,58 @@
...
+
+  public Underlay getUnderlay(String name) {
+    if (!getUnderlaysMap().containsKey(name)) {
+      throw new NotFoundException("Underlay not found: " + name);
+    }
+    return getUnderlaysMap().get(name);
+  }
+
+  private Map<String, Underlay> getUnderlaysMap() {
+    // TODO: Only read in the underlay config files once at startup, instead of for every call.
```
tjennison-google, line 39: Aren't the underlays already being loaded somewhere else? Can you at least share the code even if you can't avoid loading every time?

api/src/main/resources/api/service_openapi.yaml
marikomedlock: I defined the parameter as a reference so I could reuse it for the other endpoints (e.g. getEntity, listEntities all take an `underlayName` parameter) without redefining it inline everywhere.

Not sure what you mean about a new underlay component -- can you clarify about that please?

api/src/main/resources/api/service_openapi.yaml
marikomedlock: Do you mean if we could define a schema here for the UI config as returned by the API? Yes, and that sounds like a good idea. In that case then, I think I'm going to just remove this for now until we have our discussion about what the UI config should look like.

api/src/main/resources/api/service_openapi.yaml
marikomedlock: I meant display name, yes. I'll update the parameter name to `displayName`.

api/src/main/resources/api/service_openapi.yaml
tjennison-google: I had meant an UnderlayV2 but I forgot to update this once I got there. Anyways, I understand where else you're planning to use it and I see the format of the parameter is unfortunately a bit different from what's in UnderlayV2 so it can't be reused there.

api/src/main/java/bio/terra/tanagra/underlay/UnderlaysService.java
marikomedlock: Ah actually this comment is out of date. We only read them in once. I'll move the reading in to the constructor to make that more explicit.

I can't share code because this is reading in the new underlay config files, and the previous code reads in the yaml files with the proto-generated classes.

api/src/main/resources/api/service_openapi.yaml
tjennison-google: I had just meant a `map<string, Object>` or similar using `additionalProperties: true` but defining a schema would be fine too.

api/src/main/resources/api/service_openapi.yaml
marikomedlock: Do you mean reuse the `underlayName` parameter definition for the `name` property definition of the `UnderlayV2` schema? I'm not sure you can do that in OpenAPI -- I thought parameters needed to be separate from the schema objects.

tjennison-google: I had meant that to reuse at least the description but I see now they're not quite the same format.

api/src/main/java/bio/terra/tanagra/underlay/UnderlaysService.java
tjennison-google: I was thinking that the indexer reads them too but maybe that code isn't quite the same?

marikomedlock: Right the call to `Underlay.fromJSON` is made both from here and the indexer, so that's the code reuse. The code you see here mostly handles the differences between the API and the indexer:
- API needs to read in multiple underlays. Indexer only works with one at a time.
- API reads the index files from the JAR `resources`. Indexer typically reads them from disk (though it can also handle reading from the JAR).

tjennison-google (approved):

PR #247, opened 2022-10-13
Instances V2 endpoint: Hierarchy fields.

marikomedlock:
Added the hierarchy fields to the `/v2/instances` endpoint. This uses the new underlay config. The API specifies 3 possible hierarchy fields: `IS_ROOT`, `PATH`, `NUM_CHILDREN`. Now you can request a list of attributes and hierarchy fields to be included for each entity instance that's returned.

### DIFF
--- a/api/src/main/resources/api/service_openapi.yaml
+++ b/api/src/main/resources/api/service_openapi.yaml
@@ -1158,6 +1158,19 @@ components:
           type: array
           items:
             type: string
+        includeHierarchyFields:
+          description: Hierarchy fields to include in the returned instances
+          type: object
+          properties:
+            hierarchies:
+              type: array
+              items:
+                type: string
+            fields:
+              type: array
+              items:
+                type: string
+                enum: ["PATH", "NUM_CHILDREN", "IS_ROOT"]
         filter:
           $ref: "#/components/schemas/FilterV2"
         orderBy:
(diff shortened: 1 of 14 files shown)


api/src/main/resources/api/service_openapi.yaml
```diff
@@ -1237,11 +1250,25 @@ components:
...
+            type: object
+            nullable: true
+            $ref: "#/components/schemas/ValueDisplayV2"
+        hierarchyFields:
+          description: Hierarchy field values for this instance
+          type: object
+          properties:
+            path:
+              type: string
+              nullable: true
```
tjennison-google, line 1267: Kind of weird that only this field is nullable instead of returning a default value like the other two.

api/src/main/resources/api/service_openapi.yaml
```diff
@@ -1158,6 +1158,19 @@ components:
           type: array
           items:
             type: string
+        includeHierarchyFields:
+          description: Hierarchy fields to include in the returned instances
+          type: object
+          properties:
+            hierarchies:
```
tjennison-google, line 1165: I don't understand the model here. What is this an array of? Why doesn't the response also have an array?

api/src/main/resources/api/service_openapi.yaml
marikomedlock: Agreed it's weird. Not sure how best to address this.

If `path` has a null value, then it's not part of the hierarchy. That's the condition the UI uses to determine whether to show the hierarchy icon, right?

How about adding another hierarchy field to handle this instead of overloading `path`? e.g. `isMember` is true if the instance is part of the hierarchy, `path` is no longer nullable.

api/src/main/resources/api/service_openapi.yaml
marikomedlock: This is an array of hierarchy names. Intent is to handle entities with multiple hierarchies (e.g. source, standard).

You're right, I forgot to handle this in the response. Will update.

marikomedlock: Updated the response to also return an array of hierarchy field sets.

api/src/main/resources/api/service_openapi.yaml
marikomedlock: I took a stab at adding this `isMember` property PTAL.

api/src/main/resources/api/service_openapi.yaml
tjennison-google: Ah, I was thinking that might be the case but I could see how it worked. Makes sense now with the response changes.

api/src/main/resources/api/service_openapi.yaml
tjennison-google: That's right. I had forgotten an empty paths wasn't the same as a non-existent one.

I like this better anyways. It's more obvious what it means.

tjennison-google (approved):

PR #252, opened 2022-10-20
Eliminate "(any)" states from criteria

tjennison-google:
Instead of defaulting to "(any)" states, criteria aren't actually added to the cohort until a selection is made.

### DIFF
--- a/ui/src/apiContext.ts
+++ b/ui/src/apiContext.ts
@@ -71,7 +71,8 @@ class FakeUnderlaysApi {
         },
         {
           type: "attribute",
-          title: "Year at birth",
+          id: "tanagra-year_of_birth",
+          title: "Year of birth",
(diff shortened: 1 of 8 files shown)


ui/src/apiContext.ts
```diff
@@ -71,7 +71,8 @@ class FakeUnderlaysApi {
         },
         {
           type: "attribute",
-          title: "Year at birth",
+          id: "tanagra-year_of_birth",
```
marikomedlock, line 74: Can you use the attribute+entity as an id instead of manually specifying one here?

marikomedlock (approved):

ui/src/apiContext.ts
tjennison-google: No, there's no requirement for there to be only one of these for a given entity+attribute.

PR #255, opened 2022-10-24
Relationship fields: COUNT, DISPLAY_HINTS.

marikomedlock:
- Added a `RelationshipField` class similar to the `HierarchyField`. There are two types: `COUNT` AND `DISPLAY_HINTS`. Currently, only `COUNT` is populated with a valid number. `DISPLAY_HINTS` always returns a placeholder JSON string. This will be swapped for a map of objects in a follow-on PR.
- Wrote a new indexing job to copy the id pairs for every relationship over to a new table in the index dataset. Previously, we were joining against the source dataset for relationship filters. Now all queries only touch the index dataset.
- Updated the indexing job that computes rollup counts to copy the counts over to the entity table, so we don't have to join against a separate table every time.
- Removed the `AuxiliaryData` schema from relationship mappings in the underlay config files. Replaced it with a new `RollupInformation` schema to hold rollup count/display_hints field pointers for a relationship. I'm planning to make a similar change for the hierarchy mappings, and remove `AuxiliaryData` altogether, but will tackle that in a later PR.
- Added more indexing jobs for each entity group. Now we calculate rollup counts for the criteria-primary and criteria-occurrence relationships, for each criteria hierarchy. In the future, we should probably just calculate all possible rollup counts for every relationship, but since we're not using those fields now in the UI, I left them out for now so our indexing doesn't take longer.
- Updated all the config files for the `aou_synthetic` and `cms_synpuf` underlays and re-ran the indexing jobs for both.
- Wrapped each indexing job with a try/catch, so we do a best effort at running all the indexing jobs rather than quitting at the first failure.

### DIFF
--- a/api/src/main/resources/api/service_openapi.yaml
+++ b/api/src/main/resources/api/service_openapi.yaml
@@ -1204,6 +1204,22 @@ components:
               items:
                 type: string
                 enum: ["IS_MEMBER", "PATH", "NUM_CHILDREN", "IS_ROOT"]
+        includeRelationshipFields:
+          description: Relationship fields to include in the returned instances.
+          type: array
+          items:
+            type: object
+            description: Related entity and optional hierarchy
+            properties:
+              relatedEntity:
+                type: string
+              hierarchy:
+                type: string
+              fields:
+                type: array
+                items:
+                  type: string
+                  enum: ["COUNT", "DISPLAY_HINTS"]
         filter:
           $ref: "#/components/schemas/FilterV2"
         orderBy:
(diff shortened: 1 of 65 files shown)


api/src/main/resources/api/service_openapi.yaml
```diff
@@ -1204,6 +1204,25 @@ components:
               items:
                 type: string
                 enum: ["IS_MEMBER", "PATH", "NUM_CHILDREN", "IS_ROOT"]
+        includeRelationshipFields:
```
tjennison-google, line 1207: Why are these fields not just part of the hierarchy, especially when the hierarchy is a parameter anyways?

marikomedlock: I did it this way because we need a way to get per-node rollups that don't involve a hierarchy. Open to other suggestions though.

marikomedlock: Also, the `relatedEntity` property doesn't apply to the hierarchy fields (is_root, etc.)

tjennison-google: I see. I guess this is fine.

tjennison-google (approved):

PR #259, opened 2022-11-03
Make modifier display hints available in the API.

marikomedlock:
- Added an optional singleton relationship filter to the `getEntity` endpoint.
  - This endpoint returns the entity definition, which includes a list of attributes with their display hints. If no filter is specified, the display hints are calculated across all instances; This matches the previous behavior. If the filter is specified, then these display hints are updated to reflect the filter.
  - Currently, the optional filter only applies to occurrence entities where the `modifierAttributes` array in the `CRITERIA_OCCURRENCE` entity group is non-empty. In the future, we can lift this restriction and allow filtering on any related entity. **Among the existing OMOP underlays, `measurement_occurrence` is the only entity this applies to.**
- Updated the existing entity definition-level display hint logic (e.g. gender breakdown for all persons).
  - Enum values now include a count.
  - Numeric ranges now handle decimals.
- Re-expanded all config files with the updated display hints.


api/src/main/resources/api/service_openapi.yaml
```diff
@@ -1124,24 +1130,30 @@ components:
...
 
     DisplayHintNumericRangeV2:
       type: object
       description: Maximum and minimum values
       properties:
         min:
           description: Max value
-          type: integer
+          type: number
           format: int64
```
tjennison-google, line 1153: What does `format: int64` mean after this change? Can it just be removed?

api/src/main/resources/api/service_openapi.yaml
```diff
@@ -238,10 +238,16 @@ paths:
     parameters:
       - $ref: "#/components/parameters/UnderlayNameV2"
       - $ref: "#/components/parameters/EntityNameV2"
-    get:
```
tjennison-google, line 241: Is it difficult to create a new API? This seems like it's awkwardly shoehorned in here.

api/src/main/java/bio/terra/tanagra/api/QuerysService.java
```diff
@@ -208,6 +229,97 @@ public List<EntityInstanceCount> runInstanceCountsQuery(
...
+
+    Map<String, DisplayHint> displayHints = new HashMap<>();
+    Map<String, List<EnumVal>> runningEnumValsMap = new HashMap<>();
+    Iterator<RowResult> rowResultsItr = queryResult.getRowResults().iterator();
+    while (rowResultsItr.hasNext()) {
+      RowResult rowResult = rowResultsItr.next();
+      String attrName = rowResult.get(MODIFIER_AUX_DATA_ATTR_COL).getLiteral().getStringVal();
+
+      OptionalDouble min = rowResult.get(MODIFIER_AUX_DATA_MIN_COL).getDouble();
+      if (min.isPresent()) {
```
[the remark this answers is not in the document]
marikomedlock: Each row in the `MODIFIER_AUX_DATA` table applies to a single attribute. Each attribute can only have one kind of hint. We can have multiple different hints for a single entity, but it's at most one per attribute.

In your example for `measurement` instance=blood pressure, there could be 3 attributes each with a display hint:
- `systolic` attribute has a numeric range 70-90
- `diastolic` attribute has a numeric range 110-130
- `category` attribute has enum values low/normal/high

api/src/main/resources/api/service_openapi.yaml
marikomedlock: Hmm, I think in this case it doesn't matter to the service because this is an output schema. But for input schemas that the service reads, the format indicates whether to pull the value as a Java long (int64)/integer (int32)/float/double.

Changed this to double, since that's what we're outputing.

api/src/main/resources/api/service_openapi.yaml
marikomedlock: Not difficult. I just struggled with how to best make this available in the API. I had another version where I put it in a separate endpoint, let me push that up and see if it seems cleaner to you.

api/src/main/java/bio/terra/tanagra/api/QuerysService.java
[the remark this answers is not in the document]
tjennison-google: Hmm, apparently I don't understand what the data looks like. Starting with a simpler example like BMI, my understanding was that BMI is an instance of the measurement entity. Measurements have two attributes, a numeric value and a descriptive value (plus name etc.). To calculate the display hints, the set of measurement_occurrences where the measurement_id equals "BMI" would be queried. From there, the minimum value, the maximum value, and the set of unique descriptive values along with their counts would be collected. So each instance has a set of display hints.

From the example, I don't understand how blood pressure has unique attributes. I thought those would be the same across all measurement instances. It was also a bad example, I suppose, because it's probably actually a branch node in the hierarchy and systolic and diastolic are children and leaf nodes. The two leaf nodes would each have a a display hint associated with them containing a min, max, and enum values.

Where am I getting off track?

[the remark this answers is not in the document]
marikomedlock: I think we're on the same page about how display hints map to attributes, but maybe I'm not being clear about how this information is generated and stored in the index tables.

The `measurement_occurrence` entity has 2 attributes related to value: `value_numeric` (simple) and `value_enum` (key_display). Each attribute can have at most 1 display hint. For display hints filtered through the related `measurement` instance BMI, that could look like:
- `value_numeric` attribute has a numeric range of 1-100
- `value_enum` attribute has enum values low/normal/high
So there are 2 `measurement_occurrence` attribute display hints for the single related `measurement` instance.

This information is stored in the index tables across multiple rows. e.g. Here's the display hint rows for "Body height", which is a `measurement` instance with 2 types of hints.
Screen Shot 2022-11-03 at 1 42 16 PM

This part of the code is iterating through the rows in the index table, not `measurement` ids. Each index table row contains information about one enum value or a numeric range, not both.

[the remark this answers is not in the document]
tjennison-google: I see. That makes sense.

api/src/main/resources/api/service_openapi.yaml
marikomedlock: OK, got the alternate version working again. PTAL.

My main thought in implementing both of these alternatives was to have one way to lookup display hints. The hints might be across all instances (for demographics) or a subset defined by a related entity (for measurement detail). But the return information looks the same, so it seems like they should also be queryable in the same way.

api/src/main/resources/api/service_openapi.yaml
```diff
@@ -1419,6 +1431,60 @@ components:
           items:
             $ref: "#/components/schemas/InstanceCountV2"
 
+    DisplayHintV2:
```
marikomedlock, line 1434: Rather than have a list of these display hints, I could also have done a map (attribute name -> display hint) similar to the `InstanceV2` schema that we return for the `queryInstances` endpoint. But I thought the attribute type + data type might also be useful for display, so ended up bundling everything in this single `DisplayHintV2` object. Let me know if you prefer the other way.

tjennison-google: I liked this part better before when the hints were inside `AttributeV2`. This is fine but it feels like there's no reason not to get the static hint data along with the entities.

The only thing I didn't like about the previous version is that the one endpoint had two significantly different functions. One was to fetch static information about the underlay and the other was to query aggregate statistics.

marikomedlock: I think of all display hints as static, in the sense that they're pre-computed. Unlike the `counts` endpoint which I think of as dynamic because it has to be computed for each individual cohort.

It's just an implementation detail that we store some of these display hints in memory (via the loaded underlay config files) and some in BQ. And actually I'd like to change this -- move the unfiltered hints out of the underlay config and into BQ just like the filtered ones. I originally put them in the config just because it was faster to get working. So eventually, this will always execute a query.

My understanding is that AoU-CB also stores these hints (e.g. for demographics) in the BQ data (the `est_count` field in the `cb_criteria` table).

tjennison-google: Okay, well I don't feel that strongly either way then so go with whichever of these approaches you prefer.

tjennison-google (approved):

PR #263, opened 2022-11-07
Migrate to v2 API

tjennison-google:
* Add special columns naming to support ValueDisplay attributes and
  relationship fields. In the short term this minimizes the changes
  required to migrate to the new API because the TreeGrid doesn't need
  to change. Longer term those options could be separate fields of some
  kind but I'm still undecided if that actually adds any value.
* Remove classification filter and id type hacks that are no longer
  needed.


ui/src/data/source.tsx
```diff
@@ -531,203 +500,190 @@ function isInternalAttribute(attribute: string): boolean {
...
+    },
   };
 }
 
-function dataValueFromAttributeValue(
-  value?: tanagra.AttributeValue | null
-): DataValue {
-  return value?.int64Val ?? value?.stringVal ?? value?.boolVal ?? -1;
+function dataValueFromLiteral(value?: tanagra.LiteralV2 | null): DataValue {
+  switch (value?.dataType) {
```
marikomedlock, line 522: Do you also need to handle dates?

ui/src/data/source.tsx
```diff
@@ -531,203 +500,190 @@ function isInternalAttribute(attribute: string): boolean {
...
-          } else {
-            return false;
-          }
-          return true;
+      const path = instance.hierarchyFields?.[0]?.path;
+      if (isValid(path)) {
+        if (path === "") {
+          ancestors = [];
+        } else {
+          ancestors = path.split(".").map((id) => id as typeof data.key);
```
marikomedlock, line 654: Maybe the backend should return the path as an array of strings instead. Using a `"."` separator is an implementation detail of how we store the path on the backend and may (probably should) change in the future. Thoughts?

ui/src/data/configuration.ts
```diff
@@ -13,6 +11,9 @@ export enum SortDirection {
   Desc = "DESC",
 }
 
+export const ROLLUP_COUNT_ATTRIBUTE = "t_rollup_count";
+export const VALUE_SUFFIX = "_t_value";
```
marikomedlock, line 15: Just to make sure I'm following your PR description, the UI is basically mapping all selected information to a column, and these are the hardcoded name/suffix of the relationship field column and the display columns for value-display attributes. Is that correct?

ui/src/data/source.tsx
tjennison-google: Yes, thanks. I had forgotten you added them.

ui/src/data/source.tsx
tjennison-google: I agree. An array would be a more robust implementation.

ui/src/data/configuration.ts
tjennison-google: Yes. All of the attributes for an instances are stored in map and each column pulls out its value by key. I could instead have them configured with something like `"rollupCount": true` instead of `"column": "t_rollup_count"` but in a list in a configuration file those are basically the same and the former is maybe a bit worse because it breaks the consistency of the items.

There's no `_display` suffix because the display property is used by default if it exists. In most cases that's the desired behavior but there can be exceptions. "Vocab" is the only place `_value` is used so far because the vocabulary descriptions are quite verbose and nobody actually uses them.

marikomedlock (approved):

PR #266, opened 2022-11-10
Remove v1 code. Reorganize code into Gradle sub-projects.

marikomedlock:
- Removed `v1` endpoints and related code.
- Reorganized backend code into 3 Gradle sub-projects. This is basically the same separation that we had with the `v1` code (`api`, `lib`, `workflow`), but now there's much more shared code because both the indexing and the service endpoints are driven by the same config files.
  - `underlay` holds shared backend code (underlay, config file serialization, query engine)
  - `service` holds the service API code (endpoint definitions, `EntityFilter` classes)
  - `indexer` holds the indexing code (Apache Beam workflows and BigQuery jobs)
- Separated backend code tests into 3 GitHub actions, one for each Gradle sub-project. Only the `service` action builds the image (i.e. `jib` task).

**This is a big PR in terms of # files changed, but almost everything is just a rename or delete. The files that need close review are:**
- The sub-project build files (`indexer/build.gradle`, `underlay/build.gradle`, `service/build.gradle`)
- The top-level build settings file `settings.gradle`
- The UI files -- I attempted to fix a UI compile error by removing references to the now-delete `EntityCountsApi`, but don't know if that was the right approach.
- The GH actions files (`.github/workflows/`)

**Once this PR goes in, we need a follow-on PR in the verily-src/verily-tanagra-backend repo to update the sub-project name that it's building from api to service.**


tjennison-google (approved):
Thanks for pulling out the V1 reference in the UI I forgot.

PR #268, opened 2022-11-14
Access control interface and default implementation.

marikomedlock:
The goal of this PR is to clearly separate the service's Spring components from the authentication/authorization implementation on the backend, and to provide some starting code for collaborating with VUMC eng on the SDD access control implementation.

- Added a Spring component for doing authentication and authorization. Also added calls to this new component for each existing endpoint (except `/status`).
- Added an interface class for access control plugins to implement. Also included a default implementation that allows everything. Note that this PR doesn't include the plugin architecture -- this is just an interface for one of the plugins without any code for how we'll actually load other plugins besides the default.
- The service endpoints don't support OAuth credentials yet, so this PR just sets the credential to `null` everywhere. A future PR will add handling for Google OAuth credentials and pass them into the access control plugin.


service/src/main/java/bio/terra/tanagra/api/accesscontrol/ResourceType.java
```diff
@@ -0,0 +1,25 @@
+package bio.terra.tanagra.api.accesscontrol;
+
+import static bio.terra.tanagra.api.accesscontrol.Action.*;
+
+import java.util.List;
+
+/** Names of Tanagra resources and the actions that are possible for each. */
+public enum ResourceType {
+  UNDERLAY(List.of(READ, QUERY_INSTANCES, QUERY_COUNTS)),
```
tjennison-google, line 9: It could be worth considering GCP style permissions (e.g. `tanagra.underlay.read`). Familiar and perhaps easier to work without.

tjennison-google (approved):

service/src/main/java/bio/terra/tanagra/api/accesscontrol/ResourceType.java
marikomedlock: I think this is functionally equivalent, though maybe the naming could be made more similar. Both define a permission for an "action" on an "object" (e.g. READ an UNDERLAY), it's just specified with two enums rather than as a single string.

PR #269, opened 2022-11-15
Refactor ActionBar to be more customizable

tjennison-google:
I tried various more centralized approaches but ran into state or routing issues with each one. Instead, this applies an overall layout and each "page" is responsible for show it's own ActionBar. There are a few slightly tricky parts and a bit of duplicated code but both are minor and it's straightforward overall.

* Top level layout allocates space for the ActionBar on all "pages".
* No more duplicate ActionBars stacked on top of each other.
* ActionBar can be customized for different views, including upcoming cohort review.
* Refactor criteria based views as the amount of duplicated code increases.

### DIFF
--- a/ui/src/addCriteria.tsx
+++ b/ui/src/addCriteria.tsx
@@ -14,6 +14,7 @@ import {
   TreeGridId,
   TreeGridItem,
 } from "components/treegrid";
+import UndoRedo from "components/UndoRedo";
 import { DataEntry, DataKey } from "data/configuration";
 import { MergedDataEntry, useSource } from "data/source";
 import { useAsyncWithApi } from "errors";
@@ -145,7 +146,7 @@ export function AddCriteria() {
       }}
     >
       <Search placeholder="Search criteria or select from the options below" />
-      <ActionBar title={"Add criteria"} />
+      <ActionBar title={"Add criteria"} extraControls={<UndoRedo />} />
       {showResults ? (
         <Loading status={searchState}>
           {!data.root?.children?.length ? (
--- a/ui/src/router.tsx
+++ b/ui/src/router.tsx
@@ -49,7 +49,7 @@ export function cohortURL(cohortId: string, groupId: string) {
 }
 
 export function conceptSetURL(conceptSetId: string) {
-  return "conceptSets/" + conceptSetId;
+  return "conceptSets/edit/" + conceptSetId;
 }
 
 export function newConceptSetURL(configId: string) {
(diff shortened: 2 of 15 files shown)


ui/src/underlaySelect.tsx
```diff
@@ -13,8 +12,7 @@ export function UnderlaySelect() {
 
   return (
     <>
-      <ActionBar title="Select Dataset" />
-      <Toolbar />
+      <ActionBar title="Select Dataset" backURL={null} />
```
marikomedlock, line 15: Not for this PR, but we should probably come up with another name for Underlay or Dataset so there's no confusion. Maybe concept set too, since I assume that comes from OMOP's concept table and wouldn't really mean anything for another schema.

ui/src/addCriteria.tsx
```diff
@@ -145,7 +146,7 @@ export function AddCriteria() {
       }}
     >
       <Search placeholder="Search criteria or select from the options below" />
-      <ActionBar title={"Add criteria"} />
+      <ActionBar title={"Add criteria"} extraControls={<UndoRedo />} />
```
marikomedlock, line 149: Do you need `UndoRedo` in the `addCriteria` page? Or is this the "edit existing" criteria page and `newCriteria` is the one that adds one?

ui/src/router.tsx
```diff
@@ -49,7 +49,7 @@ export function cohortURL(cohortId: string, groupId: string) {
 }
 
 export function conceptSetURL(conceptSetId: string) {
-  return "conceptSets/" + conceptSetId;
+  return "conceptSets/edit/" + conceptSetId;
```
marikomedlock, line 52: Why the change to this route -- is there a `conceptSets/view/` for read-only?

ui/src/addCriteria.tsx
tjennison-google: Yes, the `UndoRedo` controls should be present everywhere within cohort editing. `addCriteria` is the page to select a new criteria to add either through searching or selecting from a list. `newCriteria` facilitates editing a temporary new criteria before it has actually been committed, whereas `edit` is for editing a criteria that's already in the data. They all use the `CriteriaHolder` to show the actual criteria editor and are just for sourcing the criteria appropriately.

ui/src/router.tsx
tjennison-google: This was a broken link I discovered while writing the PR. It should match the route on line 30.

ui/src/addCriteria.tsx
marikomedlock: Gotcha, thanks for the explanation.

PR #270, opened 2022-11-16
Cohort review endpoints definition.

marikomedlock:
Defined the cohort review endpoints. This PR does not include the implementation for these endpoints. I will tackle that in a separate PR.
- Review management endpoints: list, create, get, update, delete.
- Annotation management endpoints: list, create, get, update, delete.
- Query endpoint: get review instances = primary entity instance attribute values + annotation values.

Like the other query endpoints (instances, counts, hints), the review query endpoint does not include pagination. It just takes a limit. This means that the UI can only fetch the first page of results. I will add pagination to all query endpoints in a separate PR.

### DIFF
--- a/service/src/main/resources/api/service_openapi.yaml
+++ b/service/src/main/resources/api/service_openapi.yaml
@@ -486,6 +486,31 @@ paths:
         500:
           $ref: '#/components/responses/ServerError'
 
+  "/v2/studies/{studyId}/cohorts/{cohortId}/reviews/{reviewId}/instances":
+    parameters:
+      - $ref: '#/components/parameters/StudyIdV2'
+      - $ref: '#/components/parameters/CohortIdV2'
+      - $ref: '#/components/parameters/ReviewIdV2'
+    post:
+      summary: List all primary entity instances in a review and any associated annotation values
+      operationId: listInstancesAndAnnotations
+      tags: [ReviewsV2]
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/ReviewQueryV2'
+      responses:
+        200:
+          description: OK
+          content:
+            application/json:
+              schema:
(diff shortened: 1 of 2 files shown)


service/src/main/resources/api/service_openapi.yaml
```diff
@@ -1214,6 +1492,129 @@ components:
         - selectionData
         - uiConfig
 
+    ReviewV2:
+      type: object
+      properties:
```
tjennison-google, line 1497: This needs to store the cohort it was created with. It will also need a creation timestamp though that won't be useful until cohorts have modified timestamps.

marikomedlock: To get a review, the caller needs to specify the study and cohort ids as URL parameters, so they should already have that information. That's why I didn't include it as part of the response object. That said, I don't have an objection to returning the request parameters as part of the response -- would that make it easier for you to handle in the UI?

RE timestamps: OK I'll add `cohort.lastModified` and `review.created` timestamp fields.

tjennison-google:
It would also be helpful to have `required` specified where appropriate for the new objects.

service/src/main/resources/api/service_openapi.yaml
marikomedlock: On the backend, I'm planning to freeze the set of person ids included in the review when it was created, so that will use the cohort definition (i.e. the filter) as it exists at that time. I can also freeze the `selectionData` and `uiConfig`, though I don't expect the backend will need those again -- does the UI?

marikomedlock: Per offline conversation, UI does need these again so it can show a breakdown of the cohort at the time the review was created. Updated this to include a `CohortV2` reference.

tjennison-google (approved):

PR #272, opened 2022-11-17
Fix add criteria ordering

tjennison-google:


ui/src/data/source.tsx
```diff
@@ -395,7 +395,9 @@ export class BackendSource implements Source {
     const sources = lists.map(
       ([source, data]) => new MergeSource(source, data)
     );
-    const countKey = this.config.primaryEntity.entity + "_count";
+    // TODO(tjennison): Pass this as a parameter and base it on the criteria
+    // config ordering.
```
marikomedlock, line 399: Pass this as a parameter = via the UI config in the underlay?

marikomedlock (approved):

ui/src/data/source.tsx
tjennison-google: To this function. This should use the same `orderBy` information that's already there.

PR #273, opened 2022-11-17
Use hierarchy relationship fields when specified

tjennison-google:


ui/src/data/source.tsx
```diff
@@ -655,10 +661,11 @@ function processEntitiesResponse(
...
 
-      const rollupCount = instance.relationshipFields?.[0]?.count;
-      if (isValid(rollupCount)) {
-        data[ROLLUP_COUNT_ATTRIBUTE] = rollupCount;
-      }
+      instance.relationshipFields?.forEach((fields) => {
+        if (fields.hierarchy === hierarchy && isValid(fields.count)) {
+          data[ROLLUP_COUNT_ATTRIBUTE] = fields.count;
+        }
+      });
```
marikomedlock, line 668: Just to make sure I'm following, this change is basically to ignore the NO hierarchy count (i.e. `hierarchy=null`), is that right?

tjennison-google: It uses the `relationshipFields` instance with the hierarchy that matches the request. For most criteria that ignores the `hierarchy=null` instance but for criteria with no hierarchy (i.e. observations) it uses that one.

PR #274, opened 2022-11-18
Cohort endpoints implementation.

marikomedlock:
Implementation of `/cohorts` API endpoints. Includes logic for freezing a cohort and maintaining previous versions, which will be used for the implementation of the `/reviews` API endpoints, in a follow-on PR.

I did manual spot-testing on a local server. Automated tests coming in a follow-on PR.


service/src/main/java/bio/terra/tanagra/app/controller/CohortsV2ApiController.java
```diff
@@ -0,0 +1,204 @@
...
+      String studyId, String cohortId, ApiCohortUpdateInfoV2 body) {
+    accessControlService.throwIfUnauthorized(null, UPDATE, COHORT, new ResourceId(cohortId));
+
+    List<CriteriaGroup> criteriaGroups =
+        body.getCriteriaGroups() == null
+            ? null
+            : body.getCriteriaGroups().stream()
+                .map(
+                    apiCriteriaGroup -> {
+                      // Generate random 10-character alphanumeric strings for each criteria group
```
tjennison-google, line 127: It would be much simpler to store entire cohorts as JSON or protocol buffers. Storing them in separate tables like this is more useful when actually querying against them (e.g. Tanagra indexes).

service/src/main/java/bio/terra/tanagra/app/controller/CohortsV2ApiController.java
```diff
@@ -0,0 +1,204 @@
...
+  public ResponseEntity<ApiCohortV2> createCohort(String studyId, ApiCohortCreateInfoV2 body) {
+    accessControlService.throwIfUnauthorized(null, CREATE, COHORT, new ResourceId(studyId));
+
+    // Make sure underlay name and study id are valid.
+    underlaysService.getUnderlay(body.getUnderlayName());
+    studyService.getStudy(studyId);
+
+    // Generate random 10-character alphanumeric strings for the new cohort and user-facing IDs.
+    String newCohortId = RandomStringUtils.randomAlphanumeric(10);
+    String newUserFacingCohortId = RandomStringUtils.randomAlphanumeric(10);
```
tjennison-google, line 63: Why do cohorts have user facing IDs, since cohort IDs are generated by the backend? If it's for versioning, it would be clearer to name these `cohortId` and `revisionId` or something similar.

service/src/main/java/bio/terra/tanagra/app/controller/CohortsV2ApiController.java
```diff
@@ -0,0 +1,204 @@
...
+                      // that the UI-specified id doesn't need to match the DB schema on the backend
+                      // (e.g. for uniqueness). Right now, the API doesn't allow callers to update
+                      // individual criteria groups or criteria, just the whole cohort. This means
+                      // we can simplify the current implementation and just delete/re-insert all
+                      // criteria groups and criteria every time the cohort gets updated, rather
+                      // than diffing the criteria groups and updating/inserting/deleting. This is
+                      // why the IDs are randomly generated on each update. If we move away from the
+                      // simple implementation, then we'll have to expose the internal ids through
+                      // the service API.
+                      String newCriteriaGroupId = RandomStringUtils.randomAlphanumeric(15);
```
tjennison-google, line 138: Is there a reason this is 15 instead of 10 like above? This is also excessive. It's a larger key space than a 64 bit number when we could easily be using 32 bit keys (~6 alphanumeric characters).

service/src/main/resources/db/changesets/20221116_cohort_tables.yaml
```diff
@@ -0,0 +1,166 @@
...
+                    nullable: false
+                    unique: false
+              - column:
+                  name: version
+                  type: integer
+                  constraints:
+                    nullable: false
+                    unique: false
+              - column:
+                  name: is_most_recent
```
tjennison-google, line 44: Rather than sometimes having the most recent cohort being in a non-editable state, how about a single flag `is_frozen`. Each cohort would have one active version with `is_frozen = false`. Creating a review would make a copy of the active review but with `is_frozen = true`.

service/src/main/java/bio/terra/tanagra/app/controller/CohortsV2ApiController.java
marikomedlock: That would be simpler and right now the backend could treat the criteria groups/criteria like a blob, since it's just passing it to the UI. But when we move the filter-building logic to the backend, then it does need to understand these objects, so I've stored it in a deserialized way with this in mind. That's generally the rule I've tried to follow -- if it's opaque to the backend, store as a blob, if the backend needs to understand it, then store it with columns/tables.

service/src/main/java/bio/terra/tanagra/app/controller/CohortsV2ApiController.java
marikomedlock: Yes, it's for versioning. Cohort = filter on primary entity. User-facing cohort = group of one or more cohort revisions.

`cohortId` maps to a single row of the `cohort` table (= a single cohort revision), so I think that's clearer as is on the backend at least. Maybe renaming `userFacingCohortId` would be helpful? Suggestions welcome (`cohortRevisionGroup` was another name I considered).

service/src/main/java/bio/terra/tanagra/app/controller/CohortsV2ApiController.java
marikomedlock: Ah good catch. No, I meant to make them the same.

service/src/main/java/bio/terra/tanagra/app/controller/CohortsV2ApiController.java
tjennison-google: Even in your response you're describing them as "single cohort revisions". Perhaps `cohortId` and `cohortRevsionId`?

service/src/main/resources/db/changesets/20221116_cohort_tables.yaml
marikomedlock: That's a valid approach, too and the one I originally took. I changed it because I think it's slightly cleaner to:
- Only create a new cohort row when required (i.e. on update cohort, instead of on create review)
- Check if something is the most recent version with an explicit flag rather than comparing timestamps (i.e. `is_most_recent=true`, instead of `last_modified=MAX(select last_modified from cohort where user_facing_cohort_id=11)`. I then thought about adding another flag `has_been_edited`, but that's back to 2 boolean flags like I've got here.

service/src/main/java/bio/terra/tanagra/app/controller/CohortsV2ApiController.java
marikomedlock: `cohortId` is the primary key of the `cohort` table. To the backend, a cohort is just a filter on the primary entity. So one `cohortId` corresponds to one `Cohort` object. That's what I'm trying to preserve for clarity on the backend.

To the API, `cohortId` refers to the user-facing id, so this naming doesn't extend to callers of the API. It's just for the DB interactions and internal objects on the backend.

service/src/main/java/bio/terra/tanagra/app/controller/CohortsV2ApiController.java
marikomedlock: I don't see the benefit of adding a de/serialization step to DB interactions. I also think it's not a good idea to couple serialization formats throughout the stack (i.e. UI relies on same serialization structures as the internal objects as the backend DB) -- it causes a lot of headache if we ever need to update one of those or deal with upgrades.

marikomedlock: Per offline conversation, going to leave as is for now. In the future, we may consider storing JSON, but need a good way to handle schema upgrades first (maybe OpenAPI generated classes? protobufs?). Would be good to think about this after upgrading the Java version (so we can consider Java records as an option) and reevaluating the underlay config schema definition.

service/src/main/java/bio/terra/tanagra/app/controller/CohortsV2ApiController.java
marikomedlock: Per offline conversation, I made a couple changes related to this thread:
- Updated the API to define a separate `CohortRevision` object. The `Cohort` object contains a reference to the latest revision, and the `Review` object contains a reference to the frozen revision at the time the review was created.
- Replaced `user_facing_cohort_id` with `cohort_revision_group_id` in the backend table and class. There's still only a single table on the backend for cohort revisions because there's only the `underlayName` that applies across revisions (and even that could change in the future depending on how we handle underlying data upgrades). But if we add more fields to the set of cohort revisions in the future, it would make sense to split this off into a separate table.
- Replaced `getCohortUserFacing` and similarly named methods with `getCohortLatestVersion` to make the backend code easier to follow and map to the API calls.

I think there's still some ambiguity when we talk about a Tanagra cohort, because sometimes it refers to the latest revision in a collection of revisions, and sometimes it refers to a frozen revision. Will bring this up for discussion with the PMs regarding whether/how to make this clearer in the UI.

tjennison-google (approved):

service/src/main/java/bio/terra/tanagra/app/controller/CohortsV2ApiController.java
```diff
@@ -0,0 +1,219 @@
...
+      String studyId, String cohortId, ApiCohortUpdateInfoV2 body) {
+    accessControlService.throwIfUnauthorized(null, UPDATE, COHORT, new ResourceId(cohortId));
+
+    List<CriteriaGroup> criteriaGroups =
+        body.getCriteriaGroups() == null
+            ? null
+            : body.getCriteriaGroups().stream()
+                .map(
+                    apiCriteriaGroup -> {
+                      // Generate random 10-character alphanumeric strings for each criteria group
```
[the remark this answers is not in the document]
marikomedlock: Actually, I think this comment is accurate. We're generating a random id for each `criteria` and `criteriaGroup`, both of which have a user-facing id that shows up in the API schema objects, and an internal DB id that we use as the primary key in the related tables.

I'm happy to update this comment in a follow-on PR, so please continue this thread with the part of the comment that you think should be changed.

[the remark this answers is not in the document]
tjennison-google: Sorry, you're right. I forgot this part was for the criteria IDs specifically.

PR #277, opened 2022-11-23
Definition of concept set endpoints.

marikomedlock:
The `conceptSet` endpoints are basically the same as the `cohort` endpoints, with an additional `entity` parameter. For cohorts, the entity is always the primary entity, so there's no need for the caller to specify it. For concept sets, it can be any entity related to the primary entity (e.g. `condition_occurrence`).

This PR does not include the endpoints implementation, plan to tackle that in a separate PR.

### DIFF
--- a/service/src/main/resources/api/service_openapi.yaml
+++ b/service/src/main/resources/api/service_openapi.yaml
@@ -1548,11 +1548,11 @@ components:
           $ref: "#/components/schemas/ConceptSetDisplayNameV2"
         description:
           $ref: "#/components/schemas/ConceptSetDescriptionV2"
-        criteriaGroups:
+        criteria:
           type: array
-          description: Groups of criteria that define the entity filter
+          description: Single criteria that defines the entity filter
           items:
-            $ref: "#/components/schemas/CriteriaGroupV2"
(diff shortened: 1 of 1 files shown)


tjennison-google (commented):
Concept sets aren't the same as cohorts. Concept sets are only a single criteria in addition to the entity they refer to. There are no groups.

Concept sets represent the list of information to fetch about the selected participants: their diabetes conditions, their surgical operations, their abnormal blood pressure labs, etc. Filtering conditions that are diabetes and cancer is nonsense, and filtering conditions that are diabetes or cancer is redundant because they could just be listed separately.

tjennison-google (commented):
Also, should we brainstorm a more appropriate name before committing this API given that "concepts" aren't necessarily universal? "Attributes" and "attribute sets" would match our internal language. We could bring this up as a discussion question at our meeting.

marikomedlock:
RE concept sets vs cohorts: Makes sense. Changed concept sets to only have a single `Criteria` instead of a `CriteriaGroup` like cohorts.

RE naming: I'm on board with renaming this, among other things. But it's not much work to rename the backend and generated classes, so I'd rather not gate this PR on that decision, and just handle renaming separately.

PR #279, opened 2022-11-28
Add basic plugin system

arychj:
This only pulls in the basic features of the plugin system. I broke out the access control implementation so that this could be independently reviewed. @marikomedlock I'll handle the `ResourceType` / `ResourceId` changes we talked about in a separate PR.

### Adds
* basic plugin system
  * handles default plugins
  * plugin data sources
  * plugin configuration

### Fixes
* `DataSourceId` equality - AutoValue was removed, so the `DataSourceFactory` was unable to properly retrieve objects from its map


service/src/main/java/bio/terra/tanagra/plugin/PluggableConfiguration.java
```diff
@@ -0,0 +1,25 @@
+package bio.terra.tanagra.plugin;
+
+import java.util.Map;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@EnableConfigurationProperties
+@ConfigurationProperties(prefix = "tanagra.pluggable")
```
marikomedlock, line 10: This `name` + `parameters` structure is fine, but I think it should go in the underlay config so we can vary it per-dataset, not the Spring application config which would be per-deployment. Recommend putting it in the top-level `UFUnderlay` [class](https://github.com/DataBiosphere/tanagra/blob/main/underlay/src/main/java/bio/terra/tanagra/serialization/UFUnderlay.java) or add a sub-class for just the plugins.

service/src/main/java/bio/terra/tanagra/plugin/Plugin.java
```diff
@@ -0,0 +1,7 @@
+package bio.terra.tanagra.plugin;
+
+import javax.sql.DataSource;
+
+public interface Plugin {
+  void init(PluginConfig config, DataSource dataSource);
```
marikomedlock, line 6: Plugins should not use the main application Spring data sources. For plugins that need to store some internal state, here are 3 possible ways to do that:

1. Maintain its own DB as part of a companion service. Define an access control DB schema and have VUMC-specific fields. Call the companion service from an access control plugin in core codebase.
  - In this scenario, the companion service would have it's own Spring configuration file and you'd define the datasource connection/etc. info there. In the short term, if you want to code in the core Tanagra repo with plans to pull out the companion service into a separate repo later, I'd just add the datasource you want directly to the Spring yaml and we can just pull that part out when we refactor later. There's no need to make this configuration generic (i.e. plugin name + data source info) here because there should only be one other datasource and it would be temporary just for this initial implementation.

2. Use core Tanagra's existing DB (where cohorts/studies/etc are stored). Define an access control DB schema with no VUMC-specific fields. Query this DB from an access control plugin in core codebase. This is great in that it adds functionality to core Tanagra, but seems like potentially more work upfront to define and implement a generalizable access control model. Ideally we'd have product info on other VUMC datasets before doing this, but fine to make an educated first guess if you'd like.
  - In this scenario, you'd just use the existing data source info (no need to define another) and modify the existing schema. You can look at the `*Dao` classes for example code here.

3. Use core Tanagra's study metadata to store access control information. No DB schema to define, it's just a string-only key-value map. You can store whatever you want in there (e.g. list of emails of users who can access the study, serialized representation of some object), and then pull it out to determine access. This is sort of messy and less maintainable, but probably quicker to get working.
  - In this scenario, you wouldn't need to access the datasource info directly at all. Instead, you'd just save the study (e.g. by calling methods in `StudyService`) and piggyback on the existing db read/write code.

service/src/test/resources/application-plugin.yaml
```diff
@@ -0,0 +1,21 @@
+# A Spring application profile for testing with a local postgres database.
+# The database configuration here should match local-dev/local-postgres-init.sql
+tanagra:
+  jdbc:
+    dataSources:
+      # In a production instance, these values should be set with environment variables,
+      # either directly eg. TANAGRA_JDBC_DATA_SOURCES_0_PASSWORD, or injection eg. ${YOUR_ENV_VARIABLE}
+      - dataSourceId: test-data-source
+        url: jdbc:postgresql://127.0.0.1:5432/tanagra
```
marikomedlock, line 9: If you decide to store access control state in a separate DB from core Tanagra (option 1 in previous comment), then it should have a different URL from core Tanagra's DB.

marikomedlock (changes requested):

service/src/main/java/bio/terra/tanagra/plugin/Plugin.java
arychj: I disagree. There is no reason that Tanagra core can't provide raw data sources to a plugin, the alternative is basically copy/pasting a bunch of boilerplate code into individual plugins. I agree that it shouldn't be on Tanagra to manage these resources, but instantiating them based on configuration and handing them of to a plugin should be in scope.

marikomedlock: My thinking of how not-built-in plugins (e.g. a VUMC-specific one) should work is:
- The code should live in a separate repo. There is an implementation of at least one interface that is defined in core Tanagra (e.g. `AccessControlPlugin`). This repo outputs a JAR file.
- The VUMC underlay config file specifies the implementation class name (e.g. `SddAccessControlPlugin`).
- Core Tanagra is deployed with the other-repo JAR file on the classpath.
- At runtime, core Tanagra uses reflection to instantiate the implementation class from the JAR on its classpath.

My reasoning here is that plugins should not touch core Tanagra code, only config files. That means they don't modify the Spring application yaml to specify a custom data source.

So if a plugin wants to use a Spring data source, then I think it should either:
- Live in a companion service, where it can define that data source in its own application yaml.
- Be a built-in plugin (i.e. one that we package with core Tanagra) that uses core Tanagra's own DB. In that case, it can just use the data source already defined in core Tanagra's application yaml.

service/src/test/resources/application-plugin.yaml
marikomedlock: If you're going to use separate DBs, then you should test that. You can modify the script that start/stops a DB to add a new one, if this is the route you choose.

service/src/main/java/bio/terra/tanagra/plugin/Plugin.java
arychj: The VUMC-specific code should ultimately live in a separate resource yes, but it still has to be built against core Tanagra in order to import core Tanagra objects such as the plugin interface, `UserId`, `ResourceId`, and `ResourceType`. Why is passing a data source to a plugin, which is configured in an underlay loaded by the core any different?

service/src/test/resources/application-plugin.yaml
arychj: The database should be setup and torn down before each test so shouldn't matter what is being used. But I'll create a separate database if that's what you want.

service/src/main/java/bio/terra/tanagra/plugin/PluggableConfiguration.java
arychj: I'll move this to the underlay

service/src/main/java/bio/terra/tanagra/plugin/Plugin.java
marikomedlock: I think the reason we want a plugin architecture is so that you can override core Tanagra functionality without touching core Tanagra code, only underlay config files. Otherwise, we could just write everything into core Tanagra and use feature flags (this is option 2, and also a fine approach, but then we can just use the single data source). So if you have to modify core Tanagra's Spring configuration in order for your plugin to work, then we haven't got the division quite right.

I also think that the Spring data source should be defined in the same place as the DB schema and associated upgrade/migration utilities, which I expect would be in the separate repo along with the code that relies on it.

service/src/test/resources/application-plugin.yaml
marikomedlock: Currently, the Postgres DB is setup by a script, not the test code. So I don't think it will be setup/torn down before each test. Is that what you observed?

arychj: I didn't realize your current tests weren't using setup/teardown.

This datasource was used by the simple access control plugin and probably shouldn't be in this PR since I pulled that out of this PR. I'll remove remove it and add it to the new repo I'm going to have to create because Verily doesn't want Tanagra to have a built-in way to manage access control for multiple users.

service/src/main/java/bio/terra/tanagra/plugin/Plugin.java
arychj: Alright.

service/src/test/resources/application-plugin.yaml
arychj: Removed the ability for Tanagra to hand a datasource to a plugin, so this isn't relevant anymore.

marikomedlock: I disagree with this statement:
> Verily doesn't want Tanagra to have a built-in way to manage access control for multiple users

This is option 2 that I outlined above. As I said there, I think it would be great to add this functionality to core Tanagra. However, if something is in core Tanagra, then you should just use the same DB/data source that is already defined, and there is no need for separate datasource definitions.

service/src/main/java/bio/terra/tanagra/plugin/PluggableConfiguration.java
arychj: Done.

marikomedlock (changes requested):

service/src/main/java/bio/terra/tanagra/plugin/PluginService.java
```diff
@@ -0,0 +1,74 @@
...
+  public <T extends Plugin> T getPlugin(String underlayName, Class<T> c) {
+    String key = getKey(underlayName, c);
+    Plugin plugin = loadedPlugins.get(key);
+    if (plugin == null) {
+      plugin = loadPlugin(underlayName, c);
+
+      if (plugin == null) {
+        throw new PluginException(
+            String.format(
+                "Plugin '%s' not configured for underlay '%s', and no default is available",
```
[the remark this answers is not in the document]
arychj: I don't know. I don't have any requirements for this. I was trying to cover all edge cases.

service/src/main/java/bio/terra/tanagra/service/AccessControlService.java
```diff
@@ -27,7 +33,11 @@ public void throwIfUnauthorized(
       Action action,
       ResourceType resourceType,
       @Nullable ResourceId resourceId) {
-    if (!isAuthorized(credential, action, resourceType, resourceId)) {
+    // TODO: Update signature to take underlayName parameter once all routes have been configured to
+    // require an underlay.
```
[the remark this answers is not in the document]
arychj: The problem doesn't exist in `PluginService`, it exists in the classes calling `PluginService`. I can update `getPlugin()` to ignore the underlay parameter, but that would pretty much invalidate any tests done against the service.

service/src/main/java/bio/terra/tanagra/plugin/PluginType.java
```diff
@@ -0,0 +1,38 @@
...
+
+import bio.terra.tanagra.service.accesscontrol.AccessControlPlugin;
+import bio.terra.tanagra.service.accesscontrol.OpenAccessControlPlugin;
+import bio.terra.tanagra.service.identity.IdentityPlugin;
+import bio.terra.tanagra.service.identity.SingleUserIdentityPlugin;
+
+public enum PluginType {
+  ACCESS_CONTROL(AccessControlPlugin.class, OpenAccessControlPlugin.class),
+  IDENTITY(IdentityPlugin.class, SingleUserIdentityPlugin.class),
+  TEST(TestPlugin.class, TestPluginInternalImplementation.class);
```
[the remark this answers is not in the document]
arychj: I created the test plugin for testing loading of plugins as I didn't want to introduce a dependency on any specific "real" plugin in the plugin tests. I can remove this, but then I'll have to stub out a fake access control plugin in the plugin tests.

service/src/main/java/bio/terra/tanagra/plugin/PluginService.java
```diff
@@ -0,0 +1,74 @@
...
+public class PluginService {
+  private final UnderlaysService underlayService;
+  private final Map<String, Plugin> loadedPlugins = new HashMap<>();
+
+  @Autowired
+  public PluginService(UnderlaysService underlayService) throws PluginException {
+    this.underlayService = underlayService;
+  }
+
+  public <T extends Plugin> T getPlugin(String underlayName, Class<T> c) {
```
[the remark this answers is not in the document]
arychj: I did it this way because when getting when getting a plugin the developer doesn't actually care about the type, they care about the interface. If this were to take in the `PluginType` enum, the developer would then also have to know what type to cast the returned plugin to. As implemented, the developer only needs to specify the interface they care about, and they get back that interface.

service/src/main/java/bio/terra/tanagra/service/accesscontrol/AccessControlPlugin.java
```diff
@@ -1,17 +1,13 @@
...
- * Interface that all access control plugins must implement. In the future, we may consider moving
- * the credential decoding to a separate plugin, so deployments can override authentication and
- * authorization separately.
- */
-public interface AccessControlPlugin {
+import bio.terra.tanagra.plugin.Plugin;
+
+/** Interface that all access control plugins must implement. */
+public interface AccessControlPlugin extends Plugin {
   String getName();
```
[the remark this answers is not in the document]
arychj: I don't think this method is actually needed, but it was in your original implementation so I left it. I can remove it if you want.

service/src/test/resources/application-plugin.yaml
```diff
@@ -0,0 +1,17 @@
+# A Spring application profile for testing with a local postgres database.
+# The database configuration here should match local-dev/local-postgres-init.sql
+tanagra:
+  jdbc:
+    dataSources:
+      # In a production instance, these values should be set with environment variables,
+      # either directly eg. TANAGRA_JDBC_DATA_SOURCES_0_PASSWORD
+      - dataSourceId: test-data-source
+        url: jdbc:postgresql://127.0.0.1:5432/tanagra
```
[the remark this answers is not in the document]
arychj: This was an old config file that I had intended to remove but forgot about.

service/src/main/resources/config/broad/cms_synpuf/expanded/cms_synpuf.json
```diff
@@ -16,5 +16,13 @@
...
+  "plugins": {
+    "TEST": {
+      "type": "bio.terra.tanagra.plugin.TestPluginExternalImplementation",
+      "parameters": {
+        "test": "configured value"
+      }
+    }
+  }
```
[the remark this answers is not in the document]
arychj: I started trying to create a plugin-specific test underlay, but decided against it as doing so rapidly expanded into having to create a dozen additional supporting json files in order to the underlay to be valid enough to be loaded, which seemed unnecessary and difficult to maintain.

In the example you linked, the entities, entitygroups, and sql would also need to be duplicated into the test underlay in order for it to be valid. Is that what the ask is?

service/src/main/java/bio/terra/tanagra/service/AccessControlService.java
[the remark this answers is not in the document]
marikomedlock: Hmm maybe I don't understand how this will work then. If you hard-code the underlay to empty string, then won't this throw an exception for every endpoint that does an access control check?

My initial read was that you wanted a placeholder value for the underlay name so you can tackle plumbing it through to the controllers in a future PR. What I was trying to say is that I would just do the hard-coding (and the special casing that I think you'll need to handle an invalid underlay name) in `PluginService` rather than here. You can still pass the underlay name for your tests.

e.g. Have 2 overloaded methods there:
  - One takes an underlay name that you call in your tests
  - Another takes no arguments and handles returning the defaults until you've got the underlay name plumbed through everywhere.

marikomedlock (changes requested):

service/src/main/java/bio/terra/tanagra/service/AccessControlService.java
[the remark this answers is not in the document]
arychj: I don't think a default plugin should be able to be loaded if there is configured version of that plugin available. That means an underlay should be required to get a plugin.

I'll just delete all the calls to the plugin service in `AccessControlService` until the all the endpoints can be updated to require an underlay.

service/src/main/java/bio/terra/tanagra/plugin/PluginService.java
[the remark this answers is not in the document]
arychj: That ignores the advantages of generics and requires more boilerplate code be added in more places for each new plugin we create. Are generics disallowed in Tanagra?

service/src/main/java/bio/terra/tanagra/plugin/PluginType.java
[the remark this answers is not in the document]
arychj: I'm trying to test loading a configured plugin over a default implementation. Are you saying that's something which doesn't need coverage?

service/src/main/resources/config/broad/cms_synpuf/expanded/cms_synpuf.json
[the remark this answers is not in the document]
arychj: Before I put it in the synpuf underlay I tried creating my own underlay with just the underlay file and a single entity and it failed multiple times, first because there were not data pointers defined, then because there were no attributes defined. At that point I figured I was going to have to fully build out a nonsense underlay in order to get it to build, so I stuck it in synpuf. Maybe I missed something, I'll try again on Monday.

underlay/src/main/java/bio/terra/tanagra/serialization/UFPluginConfig.java
```diff
@@ -0,0 +1,56 @@
...
+    this.implementationClassName = builder.type;
+    this.parameters = builder.parameters;
+  }
+
+  @JsonPOJOBuilder(buildMethodName = "build", withPrefix = "")
+  public static class Builder {
+    private String type;
+    private Map<String, String> parameters;
+
+    public UFPluginConfig.Builder type(String type) {
```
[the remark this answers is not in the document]
marikomedlock: I think it's confusing to have a setter method with a different name than the property it is setting. Please update this.

marikomedlock (changes requested):

service/src/main/java/bio/terra/tanagra/plugin/PluginService.java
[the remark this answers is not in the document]
arychj: But the caller does care that it's an implementation of `AccessControlPlugin`? In order to access any members of the plugin the returned class needs to be of that type.

To be clear, what I'm suggesting here is:
```java
WhateverPlugin wp = pluginService.getPlugin(Interface.class, underlayName);
```

_not:_
```java
WhateverPlugin wp = pluginService.getPlugin(Implementation.class, underlayName);
```

If we get plugins using the enum, then the call to get a plugin would be (which has an unnecessary cast which could be handed elsewhere:
```java
WhateverPlugin wp = (WhateverPlugin)pluginService.getPlugin(PluginType.WHATEVER, underlayName);
```

Regarding adding a `getWhateverPlugin()` method to the plugin service for _every_ plugin... What I don't understand is why this is preferred. It is essentially replicating what the Java compiler is doing anyways via type erasure, is another boilerplate step that needs to be copy/pasted for each new plugin, and is more code that needs to be maintained.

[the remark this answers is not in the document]
marikomedlock: In your first and last code snippets, wouldn't it be `WhateverInterface` not `WhateverPlugin` as the variable type?

I think we're actually getting at the same thing -- the `Class<T> c` method argument is confusing and unnecessarily vague. It's specified as just a class name (no restrictions), when what we expect is that it's one of the (currently 2) interfaces that extend `Plugin`: `AccessControlPlugin` or `IdentityPlugin`. So I think you should change this method to make that clear and impossible for callers to mess up. Using the enum is one way -- agreed the casting required makes that less desirable. Defining two methods each with the correct return type, is another way. Or please propose something else too, I'm not wedded to these ideas.

RE adding new plugins: I think we're mixing up adding new plugin types and adding new plugin implementations. Adding a new plugin implementation for e.g. access control, wouldn't require any changes here. Adding a new plugin type would, but that's a much more invasive change and would require lots of other changes also.

service/src/main/java/bio/terra/tanagra/plugin/PluginType.java
```diff
@@ -0,0 +1,37 @@
...
+import bio.terra.tanagra.service.identity.SingleUserIdentityPlugin;
+
+public enum PluginType {
+  ACCESS_CONTROL(AccessControlPlugin.class, OpenAccessControlPlugin.class),
+  IDENTITY(IdentityPlugin.class, SingleUserIdentityPlugin.class);
+
+  private final Class<?> baseClassName;
+  private final Class<?> defaultImplementationClassName;
+
+  <TT extends Plugin, TD extends TT> PluginType(Class<TT> baseClassName, Class<TD> defaultType) {
```
marikomedlock, line 15: nit: parameter name should match property name

service/src/main/resources/config/test/plugin/test.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name" : "test",
```
marikomedlock, line 2: Glad to see you got this working! It looks great. Can you please move it under `src/test/resources` instead of `src/main/resources` though?

service/src/test/java/bio/terra/tanagra/plugin/PluginTest.java
```diff
@@ -0,0 +1,24 @@
...
+public class PluginTest extends BaseSpringUnitTest {
+  private static final String TEST_UNDERLAY = "test";
+
+  @Autowired private PluginService pluginService;
+
+  @Test
+  void defaultPluginLoading() {
+    AccessControlPlugin accessControlPlugin =
+        pluginService.getPlugin(TEST_UNDERLAY, AccessControlPlugin.class);
+    assertTrue(accessControlPlugin instanceof OpenAccessControlPlugin);
```
marikomedlock, line 22: Should this be `instanceof TestPluginExternalImplementation`? In your test underlay I see that as the plugin implementation class name.

Not sure if this is what you were going for, but perhaps you can test your plugin loading with 2 tests:
- No plugin specified, check it loads the default.
- TestImplementationClass specified, check that it throws a "class not found" or some other exception (confirming that it tried to load the non-default).
- (When we have another actual implementation class, not in this PR) RealImplementationClass specified, check that it loads and that the class name matches.

marikomedlock (commented):
This is looking great. Thanks for all the effort here.

service/src/main/java/bio/terra/tanagra/plugin/PluginService.java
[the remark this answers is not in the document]
arychj: My examples were confusing because Tanagra doesn't use `Ixxx` or `xxxInterface` naming for plugins, so `NamePlugin` is an interface, and then I was trying to differentiate between a the interface and the implementation in the method arguments and just further confused the issue. My bad.

I'm not sure what you mean that there are no restrictions on the `Class<T>` parameter of `getPlugin()`, it is type restricted to classes inheriting from `Plugin` via `<T extends Plugin>` and won't compile if anything else is passed to it.

I wasn't referring to implementations when talking about adding new plugins, I was talking about new plugin types. It seems like adding a `getAccessControlPlugin()`, `getIdentityPlugin()`, and `getWhateverPlugin()` for each plugin type is redundant as this functionality is already exposed by `getPlugin()`.

service/src/main/resources/config/test/plugin/test.json
arychj: Will `application-test.yml` load underlays from `src/test`?

service/src/test/java/bio/terra/tanagra/plugin/PluginTest.java
arychj: No, I had to remove the `TEST` plugin type from the `PluginTest` enum, so I can only use "real" plugins for testing.

I missed updating the plugin configuration in the underlay when I removed the test code. Thanks for catching.

underlay/src/main/java/bio/terra/tanagra/serialization/UFUnderlay.java
```diff
@@ -24,6 +25,7 @@ public class UFUnderlay {
   private final String primaryEntity;
   private final String uiConfig;
   private final String uiConfigFile;
+  private final Map<String, UFPluginConfig> plugins;
```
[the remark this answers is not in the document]
arychj: I tried doing this originally, but it creates a circular dependency between `tanagra.service` and `tanagra.underlay`. Are you suggesting moving `PluginType` to `tanagra.underlay` and referencing it from there?

service/src/main/java/bio/terra/tanagra/plugin/PluginService.java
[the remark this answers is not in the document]
marikomedlock: Yes, the class name is restricted to `T extends Plugin` but that means passing an implementation class name would be valid right, though I think that's not what you intended? It'd be great if there were a way to say `T extends Plugin but only immediate sub-interfaces no implementation classes`, but I don't know how to do that with Java type parameters. That's why I suggested just breaking it out into separate methods, but if you can figure out another way to make sure callers don't pass implementation class names at compile-time, I'm on board with that.

service/src/main/resources/config/test/plugin/test.json
marikomedlock: No it doesn't, but I think you can follow the example of loading test underlays from a different place in the `indexer` tests (e.g. [here](https://github.com/DataBiosphere/tanagra/blob/main/indexer/src/test/java/bio/terra/tanagra/indexing/deserialization/InvalidUnderlayTest.java)).

underlay/src/main/java/bio/terra/tanagra/serialization/UFUnderlay.java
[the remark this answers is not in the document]
arychj: That also creates a circular dependency.

service/src/main/resources/config/test/plugin/test.json
arychj: So I need to create a new application profile under test, set that as active, create an underlay config directory in test, and then override the parent directory in the test class?

marikomedlock: The example test class that I linked above does not contain a new application profile -- can you clarify why you'd like to do that?

Given that you're testing underlay config file validation, I was suggesting that you just add a new class in that same package and mimic how it loads underlays from a `src/test/resources` directory.

marikomedlock (commented):
comments inline

service/src/main/resources/config/test/plugin/test.json
arychj: I thought I needed to add the underlay to the `tanagra.underlay.files` parameter in the application profile in order for it to be loaded?

underlay/src/main/java/bio/terra/tanagra/serialization/UFUnderlay.java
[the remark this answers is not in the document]
arychj: `PluginType` references the base `Plugin` interface, the interfaces for the individual plugins, as well as the default implementations for every plugin. If I move `PluginType` to `plugin.underlay` it will need to import those, which creates the circular dependency. We could probably work around the interface issue, but the default implementations are always going to be a problem assuming you think they belong in core.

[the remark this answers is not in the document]
marikomedlock: Ah okay I understand the problem now, thanks. I guess we could just put the enum in that package and skip the class mapping. My intention around the class mapping was to avoid the caller having to know the class name, and instead letting you just pull it off the enum, but since you're not doing that here anyway, I don't see much point in that. For this PR, fine with me to do either: remove the class name from the enum since you're not using anyway and move `PluginType` to `underlay` package, or just leave it as is (`string -> string` map in the config schema) with a `TODO` comment to enforce this later.

service/src/main/resources/config/test/plugin/test.json
marikomedlock: For running the application, yes that's how you specify the underlays. But when testing the underlay config validation, you don't need to go through the Spring framework for this I don't think. The example test classes above don't rely on that property for the invalid underlays they load. e.g. `underlay/PrimaryEntityNotFound.json` is not in any application yaml file.

marikomedlock: Just put up a PR against this branch. PTAL

marikomedlock (approved):

PR #281, opened 2022-11-29
Add cohort review overview page

tjennison-google:
* Generalize action bar controls so that different ones can be displayed on different pages.
* Generalize text input using a form so keyboard controls (and eventually validation) work correctly.
* Add routing and links to/from cohort page.
* Add basic operations.

TODO

* Backend requests. This will require require a significant extension to the integration test infrastructure and I had already started this PR before the backend APIs were available.
* Stats.
* Form validation.


ui/src/router.tsx
```diff
@@ -42,12 +47,18 @@ export function AppRouter() {
...
 
+// TODO(tjennison): This is becoming spaghetti. Consider alternative ways to set
+// this up or perhaps alternative libraries.
 export function underlayURL(underlayName: string) {
   return underlayName;
 }
 
-export function cohortURL(cohortId: string, groupId: string) {
-  return "cohorts/" + cohortId + "/" + groupId;
+export function cohortURL(cohortId: string, groupId?: string) {
```
marikomedlock, line 56: Is this function called from somewhere other than `absoluteCohortURL` below? If not, maybe you can inline it there instead.

ui/src/cohortReview/cohortReviewList.tsx
```diff
@@ -0,0 +1,220 @@
...
+    initialText: selectedReview?.displayName,
+    textLabel: "Review name",
+    buttonLabel: "Rename",
+    onConfirm: onRenameReview,
+  });
+
+  return (
+    <>
+      <ActionBar
+        title={`Reviews of ${cohort.name}`}
```
marikomedlock, line 114: Nit: Cohort reviews or Reviews of Cohort {name}?

ui/src/data/source.tsx
```diff
@@ -970,3 +1035,86 @@ function normalizeRequestedAttributes(attributes: string[]) {
     ),
   ];
 }
+
+function fromAPICohort(cohort: tanagra.CohortV2): tanagra.Cohort {
```
marikomedlock, line 1039: Do you plan to keep this translation between the API and UI schemas long-term, or is it just a short-term bridge to keeping the existing code working? On the backend, I separate the API serialization from the internal objects, just wondering if you do the same thing for UI code.

ui/src/cohortToolbar.tsx
```diff
@@ -1,19 +1,38 @@
...
+} from "hooks";
 import { Link as RouterLink } from "react-router-dom";
 import { ActionCreators as UndoActionCreators } from "redux-undo";
+import { cohortReviewURL } from "router";
+
+export default function CohortToolbar() {
+  const underlay = useUnderlay();
+  const cohort = useCohort();
 
-function UndoRedo() {
```
marikomedlock, line 9: Why did you remove the `UndoRedo` toolbar function? I see that it's replaced with `CohortToolbar` in the other files as well. Is it just always included everywhere the `Cohorts` toolbar button is so you bundled them?

marikomedlock (approved):

ui/src/cohortReview/cohortReviewList.tsx
tjennison-google: I think that's unnecessarily verbose given the context of how you get to this page and the fact that the names are user selected.

ui/src/cohortToolbar.tsx
tjennison-google: Yeah, I need to specify the same set of buttons for the `ActionBar` everywhere within the cohort editor so it's easiest to group them within a component. I could have made a new one that also included the `UndoRedo` component but it's also only used within the cohort editor and I don't expect it to be used elsewhere so it made more sense to just merge them together.

PR #282, opened 2022-11-29
Dataset endpoints definition.

marikomedlock:
OpenAPI spec only. Implementation is not included in this PR.

Dataset = cohort + array of concept sets. Concept sets are updatable, cohort is not (thoughts?).

### DIFF
--- a/service/src/main/resources/api/service_openapi.yaml
+++ b/service/src/main/resources/api/service_openapi.yaml
@@ -695,6 +695,97 @@ paths:
         500:
           $ref: '#/components/responses/ServerError'
 
+  "/v2/studies/{studyId}/datasets":
+    parameters:
+      - $ref: '#/components/parameters/StudyIdV2'
+    get:
+      parameters:
+        - $ref: "#/components/parameters/OffsetV2"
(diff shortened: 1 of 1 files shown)


marikomedlock:
> I wish we had more time to think about the layout of this data. What value does creating a dataset even provide? It seems like it's mostly the list of attributes, since selecting a few named concept sets for export doesn't seem particularly onerous. But would it make more sense for the list of attributes to bind to concept sets and then there wouldn't even be a need for datasets?

Agreed about wanting more time to understand product requirements here instead of just copying what AoU-CB is doing. I guess the immediate reason we need to create datasets is that AoU-CB shows them to users as separate artifacts. I don't think SDD has that feature though. What if we thought of this instead as an export config? My (admittedly vague) understanding here is that AoU's dataset builder is primary used to export data somewhere. And I don't think there's anything you would do with a dataset other than export it.

I like the idea of the selected attributes being a part of the concept set definition. Then concept set = data that will be exported per person. Let's run it by the PMs.

marikomedlock:
Closing this PR because we agreed in our weekly check-in meeting (Verily + VUMC) to punt saving datasets until after beta launch.

PR #284, opened 2022-11-29
Define annotation value endpoints.

marikomedlock:
- Defined endpoints for creating, updating, and deleting annotation values. Annotation value = literal + (optional) review id.
- Endpoints for listing and getting annotation values are not included because they will always be fetched as part of a review instance via the `/v2/studies/{studyId}/cohorts/{cohortId}/reviews/{reviewId}/instances` endpoint.
- Updated the review instances endpoint to allow multiple values for a single annotation (key).

### DIFF
--- a/service/src/main/resources/api/service_openapi.yaml
+++ b/service/src/main/resources/api/service_openapi.yaml
@@ -604,11 +604,12 @@ paths:
         500:
           $ref: '#/components/responses/ServerError'
 
-  "/v2/studies/{studyId}/cohorts/{cohortId}/annotations/{annotationId}/values":
+  "/v2/studies/{studyId}/cohorts/{cohortId}/annotations/{annotationId}/reviews/{reviewId}/values":
     parameters:
       - $ref: '#/components/parameters/StudyIdV2'
       - $ref: '#/components/parameters/CohortIdV2'
       - $ref: '#/components/parameters/AnnotationIdV2'
+      - $ref: '#/components/parameters/ReviewIdV2'
     post:
       summary: Create a new annotation value
       operationId: createAnnotationValue
@@ -618,7 +619,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/AnnotationValueCreateInfoV2"
+              $ref: "#/components/schemas/AnnotationValueCreateUpdateInfoV2"
       responses:
         200:
           description: OK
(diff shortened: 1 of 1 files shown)


service/src/main/resources/api/service_openapi.yaml
```diff
@@ -1792,6 +1867,36 @@ components:
...
+      required:
+        - id
+        - value
+
+    AnnotationValueCreateInfoV2:
+      type: object
+      properties:
+        value:
+          $ref: "#/components/schemas/LiteralV2"
+        review:
```
tjennison-google, line 1889: After having some more time to think about it, I think we should tie annotation values to reviewIds like you originally suggested and use an alternative approach for "quick reviews" if we want to support them at some point. That approach would be some sort of default review with no fixed participant list that annotations could be assigned to instead of creating a named one.

Fetching annotation values would then return a list of 0 or more values depending which reviews they're present in. It would be convenient if the backend could return them in order of descending review creation time like how reviews are listed but the UI could also sort them.

The advantage of this approach is that it naturally handles versioned edits (e.g. review status) across reviews while still maintaining the ability to edit at any point without conflict.

marikomedlock: How about we keep the annotation value API paths as is then (i.e. `reviewId` not included) and just make the `reviewId` parameter required on the `AnnotationValue` object?

This is just a preference, but I usually think of API path parameters as hierarchical. So if we nest annotation values under a review, I'd expect them to only be visible in that review. Since we're planning on showing all annotation values (from all reviews) in each review, to me that indicates the annotation values are tied to the cohort not the review. And the review the annotation value was created in is more like a property.

In the future if we want to support adding annotation values w/o a review id ("quick reviews"), then we can just add another property to the `AnnotationValue` object e.g. `notTiedToReview`.

tjennison-google: That is fine with me and I agree that to some extent, at least, annotation values are tied to cohorts. How would reviewId being required work for the "quick reviews" case though? We'd just remove the requirement then?

I also don't have a particular opinion about the data representation, though one thing to consider is that deleting a review should delete all of the annotation values associated with it.

tjennison-google (approved):

service/src/main/resources/api/service_openapi.yaml
marikomedlock: Hmm that's a good point about deleting, so basically all annotation value crud operations are hierarchical under reviews. Then I guess it is better to go back to having `reviewId` in the path.

I'll keep the `reviewId` as a property of `AnnotationValue` though, so that it will be returned in the review `instances` endpoint.

Let's postpone designing for "quick reviews" until we get there.

PR #289, opened 2022-12-06
Authentication: JWT for IAP, Bearer token for local server

marikomedlock:
Added 3 types of authentication support:
- Bearer token. This is what a local server uses, including the backend Swagger UI page.
- JWT supplied by IAP in front of GKE. This is what the Verily dev deployment uses. I've tested it in our test deployment (https://verily-tanagra-test-backend-cluster.api.verily.com/).
- JWT supplied by IAP in front of AppEngine. I think this is what the VUMC deployment will use. I've not tested this, but included it as a first draft for the VUMC folks.

This PR just handles the authenticated Google user and logs the user email. I can see these logs in our test env in Cloud Console. I will plumb through this user email to the access control logic in a follow-on PR.


service/src/main/java/bio/terra/tanagra/app/AuthInterceptor.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  /**
+   * Returns true if the request is authenticated and can proceed. Publishes authenticated user info
+   * using Spring's SecurityContext.
+   */
+  @Override
+  public boolean preHandle(HttpServletRequest request, HttpServletResponse response, Object handler)
+      throws Exception {
+    // Clear the security context before we start, to make sure we're not using authentication
```
tjennison-google, line 45: I'm not very familiar with Spring so I don't understand the details of how this machinery works, but this is a terrifying comment. Are requests not completely isolated? What happens if there are two or more concurrent requests?

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -1928,3 +1932,17 @@ components:
           type: array
           items:
             type: string
+
+  securitySchemes:
```
tjennison-google, line 1936: Is this applicable to IAP or just bearer tokens?

service/src/main/java/bio/terra/tanagra/app/AuthInterceptor.java
marikomedlock: I tried to copy the AoU authentication strategy as much as possible. This method in particular is a straight [copy](https://github.com/all-of-us/workbench/blob/main/api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java#L203). I also spent a while looking into the Spring stuff and I think this is fine. Each request should be handled by a separate thread, and the default mode of this class is thread-local ([ref](https://docs.spring.io/spring-security/site/docs/3.0.x/apidocs/org/springframework/security/core/context/SecurityContextHolder.html)).

service/src/main/resources/api/service_openapi.yaml
marikomedlock: Good question. I think this OpenAPI definition only matters for our local server right now (in particular hitting the Swagger UI page, which I do often during development), and may matter in the future if another service uses our generated client library. I don't think it makes a difference in our dev deployment, though I have not tested that.

tjennison-google (approved):

PR #290, opened 2022-12-07
Datatable and form for studies admin tab

dolbeew:
Adds a new `StudyAdmin` component for the studies tab:
- Datatable for viewing a list of all studies
- Form for creating or editing studies
- Added separate `sd-admin/source.tsx` file for endpoints from the `StudyV2` api

### DIFF
--- a/ui/src/apiContext.ts
+++ b/ui/src/apiContext.ts
@@ -227,6 +227,8 @@ class FakeHintsApi {
   }
 }
 
+class FakeStudiesAPI {}
+
 function apiForEnvironment<Real, Fake>(
   real: { new (c: tanagra.Configuration): Real },
   fake: { new (): Fake }
@@ -259,4 +261,8 @@ export const EntitiesApiContext = apiForEnvironment(
 export const HintsApiContext = apiForEnvironment(
   tanagra.HintsV2Api,
   FakeHintsApi
+);
+export const StudiesApiContext = apiForEnvironment(
+  tanagra.StudiesV2Api,
+  FakeStudiesAPI
 );
--- a/ui/src/sd-admin/source.tsx
+++ b/ui/src/sd-admin/source.tsx
@@ -4,9 +4,17 @@ import * as tanagra from "tanagra-api";
 import { CreateStudyRequest, StudyV2, UpdateStudyRequest } from "tanagra-api";
 
 export interface AdminSource {
-  createStudy(createStudyRequest: CreateStudyRequest): Promise<StudyV2>;
+  createStudy(
+    displayName: string,
+    description: string,
+    properties: Array<object>
+  ): Promise<StudyV2>;
 
-  updateStudy(updateStudyRequest: UpdateStudyRequest): Promise<StudyV2>;
+  updateStudy(
+    studyId: string,
+    displayName: string,
+    description: string
+  ): Promise<StudyV2>;
 
   getStudiesList(): Promise<StudyV2[]>;
 }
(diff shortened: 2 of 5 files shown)


ui/src/apiContext.ts
```diff
@@ -227,6 +227,8 @@ class FakeHintsApi {
   }
 }
 
+class FakeStudiesAPI {}
```
dolbeew, line 230: Not sure how important it is to define a full fake api, just added an empty class for now

ui/src/sd-admin/source.tsx
```diff
@@ -0,0 +1,33 @@
+import { StudiesApiContext } from "apiContext";
+import { useContext, useMemo } from "react";
+import * as tanagra from "tanagra-api";
+import { CreateStudyRequest, StudyV2, UpdateStudyRequest } from "tanagra-api";
+
+export interface AdminSource {
```
dolbeew, line 6: Created a new source file since we're keeping the sd admin stuff separate. I can move these to the existing file if we think these endpoints will also be needed in core Tanagra.

ui/src/apiContext.ts
tjennison-google: This is the current method we're using to provide fake data for integration testing using Cypress. We'll need to replace these with a more comprehensive solution to support artifact storage but for now, at least, this is where you'd need to add data if you wanted to add tests.

ui/src/sd-admin/source.tsx
tjennison-google: I think this is the right choice for now.

ui/src/sd-admin/studyAdmin.tsx
```diff
@@ -0,0 +1,497 @@
...
+    sortable: true,
+    disableColumnMenu: true,
+    flex: 1,
+    renderHeader: () => (
+      <div style={{ lineHeight: "1.5rem" }}>
+        <div>Study Name</div>
+        <div>
+          <TextField
+            sx={{
+              "label+.css-apw9r9-MuiInputBase-root-MuiInput-root": {
```
tjennison-google, line 29: I've found the MUI customization options fairly confusing but I think you want something like: `& .MuiInputBase-root": ...` and `& .MuiInputLabel-root": ...`. Or perhaps without the space between the `&` and `.`. I not sure the generated style names are guaranteed to by stable or something like that.

The documentation for customizing part of the MUI controls is here: https://www.figma.com/file/Dvq2jerBXPxACCvAFLYVSJ/Tanagra?node-id=101%3A303&t=Zywx8cTzczl4dMWv-0

ui/src/sd-admin/studyAdmin.tsx
```diff
@@ -0,0 +1,497 @@
...
+
+  const getStudies = async () => {
+    const studies = await source.getStudiesList();
+    setStudies(studies);
+    setLoadingStudyList(false);
+  };
+
+  const updateStudy = async () => {
+    setLoadingStudy(true);
+    const updateStudyRequest: UpdateStudyRequest = {
```
tjennison-google, line 187: Optional: This is the sort of stuff I would put in AdminSource and why I created the Source in the first place, so that the UI doesn't have to know about the details of the API calls and would instead call simple functions like `updateStudyInfo(id, name, description)` or use objects (e.g. a Study) defined in a UI-centric manner.

tjennison-google (approved):

ui/src/apiContext.ts
dolbeew: Ok, good to know. I have not used Cypress but I'll take a look and try to follow up with some tests.

ui/src/sd-admin/studyAdmin.tsx
dolbeew: Yeah, I think I copied those values from the Chrome inspector but that makes sense that they would change. Both your suggestions seem to work, thanks.

ui/src/sd-admin/studyAdmin.tsx
dolbeew: Ah, that makes sense. Updated the `createStudy` function also.

dolbeew (commented):
Thanks!

PR #294, opened 2022-12-09
Add getMe endpoint for current user information.

marikomedlock:
- Added an endpoint `/v2/profile` for getting information about the logged in user. Currently the only information returned is the user's email.
- Explicitly specified the paths that we don't need to decode the authentication token for. Not sure if this is strictly required, but seems better/clearer to be explicit about this anyway.

### DIFF
--- a/service/src/main/java/bio/terra/tanagra/app/WebMvcConfig.java
+++ b/service/src/main/java/bio/terra/tanagra/app/WebMvcConfig.java
@@ -17,6 +17,6 @@ public class WebMvcConfig implements WebMvcConfigurer {
 
   @Override
   public void addInterceptors(InterceptorRegistry registry) {
-    registry.addInterceptor(authInterceptor);
+    registry.addInterceptor(authInterceptor).addPathPatterns("/v2/**", "/status");
   }
 }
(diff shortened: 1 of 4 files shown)


service/src/main/java/bio/terra/tanagra/app/WebMvcConfig.java
```diff
@@ -17,6 +17,6 @@ public WebMvcConfig(AuthInterceptor authInterceptor) {
 
   @Override
   public void addInterceptors(InterceptorRegistry registry) {
-    registry.addInterceptor(authInterceptor);
+    registry.addInterceptor(authInterceptor).addPathPatterns("/v2/**", "/status");
```
tjennison-google, line 20: I think it would be safer to require auth everywhere by default (`/**`?) and only exclude specific URLs if we need to.

tjennison-google (approved):

service/src/main/java/bio/terra/tanagra/app/WebMvcConfig.java
marikomedlock: Good idea. Changed to intercept all paths and added an exception for the static swagger resources that let us hit endpoints manually in a webpage.

PR #296, opened 2022-12-09
Added implementation for Cloud SQL JDBC connection

NickofNasa:
implementation added to allow for Cloud SQL Socket JDBC connections. This is neccissary for the VUMC App Engine Deployment of the tanagra service.


marikomedlock:
Per discussion today, holding off reviewing this until we confirm this change works for VUMC's deployment. Also please re-merge with main to kick off the tests -- coverage has improved significantly since this was first opened. Thanks. @NickofNasa

PR #297, opened 2022-12-09
Pass user id to access control service. Remove identity plugin.

marikomedlock:
- Passed the `UserId` object through to the access control service and plugin classes. This allows us to use the user email or google subject id to determine access.
- Removed the plugin that allows for custom logic to turn a credential object into a `UserId` (=subject id + email) object. Originally I'd thought this might vary across underlays, but having worked through the implementation, I think authentication is going to be per-deployment, not per-underlay. Currently, Tanagra supports 3 implementations of this credential->`UserId` logic: IAP in front of GKE, IAP in front of AppEngine, Bearer token (for local server).
- Bumped the Google `oauth2` library version to be compatible with the BQ library version. This was causing an exception when running queries.


service/src/main/java/bio/terra/tanagra/app/AuthInterceptor.java
```diff
@@ -122,10 +124,17 @@ public void postHandle(
...
       Object handler,
-      ModelAndView modelAndView)
-      throws Exception {
+      ModelAndView modelAndView) {
     // Clear the security context, just to make sure nothing subsequently uses the credentials
     // set up in here.
     SecurityContextHolder.clearContext();
   }
+
+  public static UserId getCurrentUserOrThrow() {
```
tarekmamdouh, line 133: I am not clear why this method is in the interceptor, although it's not being used by the interceptor. Would it be better to have an AuthUtils or AuthCommon class for these methods instead?

marikomedlock: I thought it was helpful to have all the auth stuff together, and since this is just one method. But happy to pull this out into a separate class if that's clearer.

marikomedlock: OK I moved this method to the `UserId` class instead. That class now has 2 static constructors:
- `fromToken` is called from the token utils classes to create the current user object
- `currentUser` is called from the controller classes to get the current user object
PTAL and see if this looks cleaner to you too.

tarekmamdouh: thank you @marikomedlock yes this looks much better. I believe that interceptors shouldn't really be referenced by any other service directly, so now things seem clearer.

tarekmamdouh (approved):
LGTM

PR #299, opened 2022-12-14
Add config property for disabling authentication.

marikomedlock:
Added a config property for disabling authentication. Updated the `service/local-dev/run_server.sh` script to optionally set this config property (default is still to enable authentication).

This change is intended for use during local UI development only. Since the UI doesn't currently include a login flow, all endpoints are returning `401`s now that I've added checks for the authentication headers to the backend. The 2 existing Tanagra deployments are both behind IAP, so the login flow is handled by the proxy and the UI doesn't need to do anything to stick the authentication token in the header. For local backend development, we can enable authentication because the Swagger UI page allows you to supply a bearer token that the backend knows how to handle.

### DIFF
--- a/service/local-dev/run_server.sh
+++ b/service/local-dev/run_server.sh
@@ -1,13 +1,34 @@
 #!/bin/bash
 
-export TANAGRA_DATABASE_NAME=tanagra_db
-export TANAGRA_DB_INITIALIZE_ON_START=true
-export TANAGRA_DB_USERNAME=dbuser
-export TANAGRA_DB_PASSWORD=dbpwd
+setupEnvVars() {
+  disableAuthChecks=$1 # 0=false, 1=true
 
-export TANAGRA_UNDERLAY_FILES=broad/aou_synthetic/expanded/aou_synthetic.json,broad/cms_synpuf/expanded/cms_synpuf.json
-export TANAGRA_FEATURE_ARTIFACT_STORAGE_ENABLED=true
-export TANAGRA_AUTH_IAP_GKE_JWT=false
-export TANAGRA_AUTH_BEARER_TOKEN=true
+  export TANAGRA_DATABASE_NAME=tanagra_db
+  export TANAGRA_DB_INITIALIZE_ON_START=true
+  export TANAGRA_DB_USERNAME=dbuser
+  export TANAGRA_DB_PASSWORD=dbpwd
+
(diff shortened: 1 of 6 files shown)


service/src/main/java/bio/terra/tanagra/app/AuthInterceptor.java
```diff
@@ -97,6 +97,9 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
         userAuth =
             new UserAuthentication(
                 userId, bearerToken.getToken(), UserAuthentication.TokenType.BEARER_TOKEN);
+      } else if (authConfiguration.isDisableChecks()) {
+        LOGGER.warn("Authentication checks are disabled");
```
melissachang, line 101: optional

Authentication checks are disabled. This should only happen for local development.

service/local-dev/run_server.sh
```diff
@@ -1,13 +1,34 @@
...
-export TANAGRA_DB_INITIALIZE_ON_START=true
-export TANAGRA_DB_USERNAME=dbuser
-export TANAGRA_DB_PASSWORD=dbpwd
-
-export TANAGRA_UNDERLAY_FILES=broad/aou_synthetic/expanded/aou_synthetic.json,broad/cms_synpuf/expanded/cms_synpuf.json
-export TANAGRA_FEATURE_ARTIFACT_STORAGE_ENABLED=true
-export TANAGRA_AUTH_IAP_GKE_JWT=false
-export TANAGRA_AUTH_BEARER_TOKEN=true
+setupEnvVars() {
+  disableAuthChecks=$1 # 0=false, 1=true
```
melissachang, line 4: opt Since this file is only for local development, should we take out `disableAuthChecks` and just do lines 16-17?

melissachang (approved):

dolbeew (approved):
Working locally for me. Thanks!

service/src/main/java/bio/terra/tanagra/app/AuthInterceptor.java
marikomedlock: done.

service/local-dev/run_server.sh
marikomedlock: This is only for local UI development. For local backend development, i.e. where we're testing with the Swagger page, we still want authentication enabled. I added the bearer token authentication method specifically to support backend testing on a local env (i.e. not behind IAP).

marikomedlock: Updated `CONTRIBUTING.md`

PR #302, opened 2022-12-16
Bump, tag, publish GHA.

marikomedlock:
- Added a `client` Gradle sub-project to build and publish the client library to Broad artifactory.
- Added a `publish.sh` script for manually publishing the library.
- First pass at a GHA to automatically bump the version and publish the library and service image. I don't think I can test this GHA until at least one version of it is in the main branch, so consider this a first draft and I'll tackle any problems in a follow-on PR.
- Started version at `0.0.6`.

I generally tried to follow what Terra services (wsm, rbs, user manager) are already doing, so most of this code is copied from one of those repos. One weird thing I noticed is that the client library seems to publish with a `-plain.jar` suffix. I don't know where it's getting the `plain` from, but decided to leave as is for now and come back later if this is a problem.

Not strictly part of this PR, but I also added the `BROADBOT_TOKEN` as a GH secret with this `terraform-ap-deployments` [PR](https://github.com/broadinstitute/terraform-ap-deployments/pull/856). This is needed for the bump-publish GHA.


.github/actions/bump-skip/action.yaml
```diff
@@ -0,0 +1,28 @@
+# This action must be done after the checkout action
+name: 'bump-skip'
+description: 'Set skip-out when we are doing a version bump'
+author: 'dd'
```
melissachang, line 4: Delete?

client/gradle/open-api.gradle
```diff
@@ -0,0 +1,40 @@
...
+                            "modelTests=false," +
+                            "modelDocs=false," +
+                            "dateLibrary=java8"
+            ]
+        }
+    }
+}
+
+// Internal dependencies
+//generateSwaggerCodeClient.dependsOn(':openapi:buildOpenApi')
```
melissachang, line 32: Delete lines 32, 38, 39?

melissachang (approved):

.github/actions/bump-skip/action.yaml
marikomedlock: done.

client/gradle/open-api.gradle
marikomedlock: done, thank you.

PR #303, opened 2022-12-21
Upload cms_synpuf expanded config

melissachang:
Before this PR, local UI had error:

image

After this PR, local UI works:

image


docs/INDEXING.md
```diff
@@ -56,13 +56,14 @@ Below you can see an [example](#omop-example) of the commands for an OMOP datase
...
 ```
-export GOOGLE_APPLICATION_CREDENTIALS=/credentials/indexing_sa.json
+export GOOGLE_APPLICATION_CREDENTIALS=$(PWD)/rendered/tanagra_sa.json
 ```
 Expand the defaults, scan the source data, and generate an expanded underlay config file that includes all this 
-information. The first argument `/config/input/omop.json` is a pointer to the user-specified underlay file.
-The second argument `/config/output/` is a pointer to the directory where Tanagra can write the expanded config files.
+information. The first argument is a pointer to the user-specified underlay file.
+The second argument is a pointer to the directory where Tanagra can write the expanded config files.
+Both arguments must be absolute paths. Example:
```
marikomedlock, line 64: Thank you for updating this!

marikomedlock (approved):

PR #305, opened 2022-12-21
Add created, created_by columns to artifacts

melissachang:
Add created, createdBy columns to Cohort, Concept Set, Cohort Review, Study.

Also added last_modified to artifacts that were missing it.

I noticed when I update cohort (change name), last_modified isn't updated. That's because [this method](https://github.com/DataBiosphere/tanagra/blob/952892907fdf67f8929a111a65fa94141a5dea8d/service/src/main/java/bio/terra/tanagra/db/CohortDao.java#L307) doesn't set last_modified. I'll follow up with Mariko after the holidays.

Manually tested:
- New fields returned

image

- created/created_by don't change after update
- For created_by: Existing rows (rows that existed before changeset) have "unknown". New rows are set to user email.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6 @@
+{
+  "name": "tanagra",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {}
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,0 @@
-{
-  "name": "tanagra",
-  "lockfileVersion": 2,
-  "requires": true,
-  "packages": {}
-}
--- a/service/src/main/java/bio/terra/tanagra/db/ReviewDao.java
+++ b/service/src/main/java/bio/terra/tanagra/db/ReviewDao.java
@@ -10,8 +10,6 @@ import bio.terra.tanagra.query.QueryResult;
 import bio.terra.tanagra.query.RowResult;
 import bio.terra.tanagra.service.artifact.Cohort;
 import bio.terra.tanagra.service.artifact.Review;
-import java.sql.Timestamp;
-import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
--- a/service/src/main/java/bio/terra/tanagra/db/ReviewDao.java
+++ b/service/src/main/java/bio/terra/tanagra/db/ReviewDao.java
@@ -47,7 +47,7 @@ public class ReviewDao {
               .displayName(rs.getString("display_name"))
               .description(rs.getString("description"))
               .size(rs.getInt("size"))
-              .created(rs.getTimestamp("created"));
+              .created(DbUtils.timestampToOffsetDateTime(rs.getTimestamp("created")));
 
   // SQL query and row mapper for reading a review instance.
   private static final String REVIEW_INSTANCE_SELECT_SQL =
--- a/service/src/main/java/bio/terra/tanagra/db/ReviewDao.java
+++ b/service/src/main/java/bio/terra/tanagra/db/ReviewDao.java
@@ -37,7 +37,7 @@ public class ReviewDao {
 
   // SQL query and row mapper for reading a review.
   private static final String REVIEW_SELECT_SQL =
-      "SELECT r.cohort_id, r.review_id, r.display_name, r.description, r.size, r.created, r.last_modified FROM review AS r "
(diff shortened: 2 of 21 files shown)


package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
marikomedlock, line 1: Is this an `npm` generated file? I would have expected it to put things in the `ui` directory, not at the top-level.

service/src/main/java/bio/terra/tanagra/db/ReviewDao.java
```diff
@@ -79,16 +81,17 @@ public void createReview(
...
     MapSqlParameterSource params =
         new MapSqlParameterSource()
             .addValue("cohort_id", cohort.getCohortId())
             .addValue("review_id", review.getReviewId())
             .addValue("display_name", review.getDisplayName())
             .addValue("description", review.getDescription())
             .addValue("size", review.getSize())
-            .addValue("created", Timestamp.from(Instant.now()));
+            .addValue("created", Timestamp.from(Instant.now()))
+            .addValue("created", review.getCreatedBy());
```
marikomedlock, line 94: `.addValue("createdBy", review.getCreatedBy());`?

service/src/main/java/bio/terra/tanagra/db/ReviewDao.java
```diff
@@ -47,7 +47,9 @@ public class ReviewDao {
               .displayName(rs.getString("display_name"))
               .description(rs.getString("description"))
               .size(rs.getInt("size"))
-              .created(rs.getTimestamp("created"));
+              .created(DbUtils.timestampToOffsetDateTime(rs.getTimestamp("created")))
+              .createdBy(rs.getString("created"))
```
marikomedlock, line 51: `.createdBy(rs.getString("createdBy"))`?

dolbeew (approved):
I'll defer to @marikomedlock on the api code but the ui looks good. Deployed and tested locally.

service/src/main/java/bio/terra/tanagra/db/StudyDao.java
```diff
@@ -58,15 +61,16 @@ public StudyDao(NamedParameterJdbcTemplate jdbcTemplate) {
...
+            + "VALUES (:study_id, :display_name, :description, CAST(:properties AS jsonb), :created_by)";
 
     MapSqlParameterSource params =
         new MapSqlParameterSource()
             .addValue("study_id", study.getStudyId())
             .addValue("display_name", study.getDisplayName())
             .addValue("description", study.getDescription())
-            .addValue("properties", DbSerDes.propertiesToJson(study.getProperties()));
+            .addValue("properties", DbSerDes.propertiesToJson(study.getProperties()))
+            .addValue("created_by", study.getCreatedBy());
```
marikomedlock, line 73: You're not setting the `created` field here because its set to default to `now()` in the liquibase file right? If so, a comment here explaining that would be helpful. ditto for the other dao classes

marikomedlock (approved):
RE `updateCohortHelper`: Good catch, thanks. Would you mind updating that field here too? I'm also happy to put in the fix in a separate PR, let me know.

LGTM once the 2 field names are fixed.

package-lock.json
melissachang: Hmm, I might have accidentally run `npm` in top-level directory. Deleted this file.

service/src/main/java/bio/terra/tanagra/db/ReviewDao.java
melissachang: Done, good catch. Review was the one artifact I didn't manually test, because I wasn't sure how to create a review in swagger.

service/src/main/java/bio/terra/tanagra/db/ReviewDao.java
melissachang: Done

service/src/main/java/bio/terra/tanagra/db/StudyDao.java
melissachang: Done. I agree comment is better, thanks.

melissachang:
> RE `updateCohortHelper`: Good catch, thanks. Would you mind updating that field here too? I'm also happy to put in the fix in a separate PR, let me know.

I thought I responded to this, looks like the comment got lost or something.

The fix to updateCohortHelper looked non-trivial, so I wanted to talk to you first. I can fix after the holidays, or you can fix before then if you want. :) Concept set had the same problem. I didn't test review or study.

PR #306, opened 2022-12-22
Don't push the Docker image to GCR.

marikomedlock:
Verily deployments are building their own Docker images, not pulling from the Broad public container registry. VUMC deployments are running a JAR on AppEngine, so don't use the Docker image at all.

If we ever have a non-Verily GKE deployment, we can consider enabling this then.


ddietterich (approved):
lgtm - are you able to test it?

marikomedlock:
> lgtm - are you able to test it?

Yes, I tested the action using the workflow from this branch. It publishes the library successfully and just fails on creating the release, as it should (I think) since that checks if there are any changes to `main`.

PR #309, opened 2023-01-05
Migrate from useAsync to SWR

tjennison-google:
SWR is a more robust library for handling web requests and will offer support for automatically revalidating data (e.g. when adding a cohort and refetching the list). and caching. Migrating the existing code first so that there won't be two libraries once artifact storage support is added.

Each invocation of useSWR requires a "key" object for cache management purposes. A future PR could consider moving some of the parameters out of the useCallback list since the key is passed to the fetch function but in many cases that doesn't make sense because the object (e.g. an underlay) shouldn't be included in the key in its entirety so some further thought is required.

Response error parsing has been moved into the sourse since the existing method is incompatible with SWR. It makes more sense for it to be there anyways.


ui/package.json
```diff
@@ -40,6 +40,7 @@
     "react-test-renderer": "^17.0.2",
     "recharts": "^2.1.12",
     "redux-undo": "^1.0.1",
+    "swr": "^2.0.0",
```
marikomedlock, line 43: Can you remove the dependency on useAsync (`react-async`?) with this switch?

ui/src/data/source.tsx
```diff
@@ -1122,3 +1134,18 @@ function saveLocalReviews(reviews: tanagra.ReviewV2[]) {
     })
   );
 }
+
+function parseAPIError<T>(p: Promise<T>) {
```
marikomedlock, line 1138: Do the `throw`s here also log to the console? If not, maybe logging the `response` could be helpful for debugging?

marikomedlock (approved):
Just curious, how did you choose the SWR library -- is it recommended by Verily or Terra or just something you had experience with before?

tjennison-google:
> Just curious, how did you choose the SWR library -- is it recommended by Verily or Terra or just something you had experience with before?

I choose it because Terra uses it.

PR #311, opened 2023-01-05
Add an option to use Verily underlays to the local dev script

tjennison-google:
Also use getopts to simplify handling multiple command line arguments.

### DIFF
--- a/service/local-dev/run_server.sh
+++ b/service/local-dev/run_server.sh
@@ -1,34 +1,51 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
-setupEnvVars() {
-  disableAuthChecks=$1 # 0=false, 1=true
+usage() { echo "$0 usage flags:" && grep " .)\ #" $0; }
 
-  export TANAGRA_DATABASE_NAME=tanagra_db
-  export TANAGRA_DB_INITIALIZE_ON_START=false
-  export TANAGRA_DB_USERNAME=dbuser
-  export TANAGRA_DB_PASSWORD=dbpwd
+usage
+echo
 
-  export TANAGRA_UNDERLAY_FILES=broad/aou_synthetic/expanded/aou_synthetic.json,broad/cms_synpuf/expanded/cms_synpuf.json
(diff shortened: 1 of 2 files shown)


service/local-dev/run_server.sh
```diff
@@ -1,34 +1,45 @@
-#!/bin/bash
+#!/usr/bin/env bash
+usage() { echo "$0 usage:" && grep " .)\ #" $0; exit 0; }
```
marikomedlock, line 2: Cool! I've not seen this before.

Could you please also add the usage to a comment at the top of this file, so it's easy to see how to run this?

marikomedlock (commented):
Can you please update `CONTRIBUTING.md` where it talks about this script?

service/local-dev/run_server.sh
tjennison-google: It actually pulls the comments from the cases below to build the usage and it looks like:
```
$ ./service/local-dev/run_server.sh
./service/local-dev/run_server.sh usage:                                                                                                                         
    a) # Disable authentication.                                                                                                                                 
    v) # Use Verily underlays.                                                                                                                                   
    h | *) # Display help.
```
Is it sufficient to have the flags documented that way? It's nice not to have duplicate documentation so if I was going to separately add a usage block at the top I'd want that to double as comments and output instead of also having this.

marikomedlock: I don't think you need to duplicate everything, but it wasn't immediately clear to me how to specify an option. I kept trying `./service/local-dev/run_server.sh a` (i.e. without the required dash). And the usage doesn't print out if you just call the script `./service/local-dev/run_server.sh`, that uses the defaults instead.

So maybe another approach is to change the script so `./service/local-dev/run_server.sh` prints out the usage, and somewhere note that these are flags, not arguments?

tjennison-google: How about this?

tjennison-google:
> Can you please update `CONTRIBUTING.md` where it talks about this script?

Done

marikomedlock (approved):
Thanks!

PR #312, opened 2023-01-06
Call VUMC admin service.

marikomedlock:
- VUMC admin service connection. This isn't being used by the service yet, it's just a POC of service-to-service communication.
  - Added a dependency on the VUMC admin service client library and called some of its endpoints in a test.
  - Added configuration to connect to a deployment (URL and OAuth client id). The only deployment of the admin service is in Verily's `test` environment, so both the production and test application configurations point to `https://tanagra-vumc-test.api.verily.com`. Once we have it deployed in our `dev` environment also, I'll change the configs to point to that instead.
  - Used application default credentials to generate a JWT for talking to the admin service. This means the service-to-service communication uses the calling service's SA, not end user credentials.
  - Added a new `/v2/vumc-admin-service-test` endpoint intended to help debug communication between the services. The VUMC admin service has a similar test endpoint, so we can check that they're both able to call the other.
- Added a new `/v2/version` endpoint that returns the git hash of the currently deployed code. This is intended to help with debugging (e.g. confirm we're talking to latest deployed code), in both the Verily and VUMC deployments.
- In `AuthInterceptor` we check whether a request needs to be authenticated or not. Previously, this was checking for an empty `security: []` property of the OpenAPI endpoint. I changed this to check for an `Unauthenticated` tag instead, both for slightly improved readability and because the tags are consistent in later Swagger versions, while the security property is not.
- Consolidated all manipulation of the Spring security context into the `SpringAuthentication` class. Previously, this was spread across `AuthInterceptor` and `UserId`.


service/src/main/java/bio/terra/tanagra/app/auth/AuthInterceptor.java
```diff
@@ -58,63 +61,50 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
...
     ApiOperation apiOp = AnnotationUtils.findAnnotation(method.getMethod(), ApiOperation.class);
     if (apiOp != null) {
-      Authorization[] authorizations = apiOp.authorizations();
-      for (Authorization auth : apiOp.authorizations()) {
-        if (!auth.value().isEmpty()) {
-          LOGGER.info("Authorization required by endpoint: {}", request.getRequestURL().toString());
-          isAuthRequired = true;
-          break;
+      for (String tag : apiOp.tags()) {
+        if (!tag.isEmpty() && OPENAPI_TAG_AUTH_NOT_REQUIRED.equals(tag)) {
```
tjennison-google, line 67: `!tag.isEmpty()` is redundant since then it won't equal `OPENAPI_TAG_AUTH_NOT_REQUIRED` anyways?

service/build.gradle
```diff
@@ -59,6 +59,12 @@ dependencies {
     implementation group: "org.apache.commons", name: "commons-text", version: "1.9"
     annotationProcessor group: "org.springframework.boot", name: "spring-boot-configuration-processor"
 
+    // VUMC Admin dependency
+    implementation group: "bio.terra", name: "tanagra-vumc-admin-client", version: "0.0.1d-SNAPSHOT"
```
tjennison-google, line 63: Do you know what the longer term plan is for this yet?

tjennison-google (commented):
Have we come to any decision on whether Tanagra would be forwarding admin service requests (or vice-versa)? Using ADCs to do a permission check is reasonable but forwarding requests using ADCs not so much.

marikomedlock:
> Have we come to any decision on whether Tanagra would be forwarding admin service requests (or vice-versa)? Using ADCs to do a permission check is reasonable but forwarding requests using ADCs not so much.

No, that will be VUMC/Erik's call. If it were my call, I think I would lean toward the AoU-Terra model and not forward requests. e.g. Admin service has its own `getStudy` endpoint that checks access, fetches the study object from core Tanagra, decorates it with any VUMC-specific properties, and returns this expanded study object to the caller. In that case, we don't want users to fetch the core study object directly, so it's fine to limit those endpoints to only the admin service SA.

service/src/main/java/bio/terra/tanagra/app/auth/AuthInterceptor.java
marikomedlock: Good point. Removed.

service/build.gradle
tjennison-google: Sorry, I should have been more specific. I was thinking more about how this is integrated into the core service.

tjennison-google (approved):

service/build.gradle
marikomedlock: I'm planning to add a new implementation of the `AccessControlPlugin` interface. Currently, we only have `OpenAccessControlPlugin` that allows all requests. `VumcAdminAccessControlPlugin` will call the `isAuthorized` and `listAuthorizedResources` endpoints of the admin service. And then I can use an underlay config property to specify which plugin class to call. (I'll tackle this in a future PR.)

That should work with the way things are setup now. However, once Erik starts developing the admin service, he may decide to remove those endpoints or change the service architecture, and we'd have to adapt here to those changes. e.g. One way this could change is if he wants core Tanagra to notify the admin service every time we create an artifact. In that case, we'd need to expand the access control plugin interface to also call VUMC admin service at create time. If it were my call, I would not take that approach, so that's why I held off implementing it until we know more details of Erik's plan here.

PR #315, opened 2023-01-09
Add studies landing page

dolbeew:
Adds user-facing studies page with list of study cards and table of resources. Uses the same basic layout as the Researcher Workbench homepage.
Screenshot 2023-01-09 at 12 51 06 PM


ui/src/router.tsx
```diff
@@ -19,23 +20,29 @@ export function AppRouter() {
       <Routes>
         <Route index element={<UnderlaySelect />} />
         <Route path=":underlayName">
-          <Route index element={<Datasets />} />
-          <Route path="cohorts/:cohortId/:groupId">
-            <Route index element={<Overview />} />
-            <Route path="add">
-              <Route index element={<AddCriteria />} />
-              <Route path=":configId" element={<NewCriteria />} />
+          <Route index element={<Studies />} />
```
dolbeew, line 23: I added the new route where the workflow made the most sense to me (select underlay -> select study -> select cohorts/concept sets). Not sure if this will be the final workflow.

tjennison-google: The final workflow will be to deploy the VUMC and core Tanagra pieces separately, either as separate services or plugins (or maybe both). In either case, I don't think the landing page will be part of the routing path in core Tanagra like this. Instead, the routing would be controlled by the deployment and would route to the landing page at `/` which would link to core Tanagra routed at some other URLs (e.g. `/underlay/cohort/12345/`) and open them in new tabs, like a Google Docs style navigation. That is the most practical way to build the app in a modular fashion.

For now, how about putting the landing page at something like `/landing` in the routing like how the admin page is setup. It can still link to the cohorts/concept sets and we can figure out the necessary config variables to mimic what the final setup might look like.

tjennison-google (approved):

PR #316, opened 2023-01-09
Point to admin service in dev. Update README for admin service.

marikomedlock:


service/build.gradle
```diff
@@ -60,7 +60,7 @@ dependencies {
     annotationProcessor group: "org.springframework.boot", name: "spring-boot-configuration-processor"
 
     // VUMC Admin dependency
-    implementation group: "bio.terra", name: "tanagra-vumc-admin-client", version: "0.0.2-SNAPSHOT"
+    implementation group: "bio.terra", name: "tanagra-vumc-admin-client", version: "0.0.3-SNAPSHOT"
```
tjennison-google, line 63: When exactly does this version need updating? I don't think I noticed where it comes from in your other PRs.

tjennison-google (approved):

service/build.gradle
marikomedlock: It only needs to be updated when the admin service API changes. Most changes in that repo won't require a version bump here. This one isn't needed really, I was just happened to be testing this.

You can see the most up to date version in the [admin service repo](https://github.com/DataBiosphere/tanagra-vumc-admin/) (client library gets published automatically on PR merge, same as this repo). But instead of monitoring that, I think we can just bump the client library version whenever Erik tells us he's changed something.

tjennison-google: So is it coming from the tag?

marikomedlock: Well technically it's the client library version, which you can see in the [Broad public Maven repo](https://broadinstitute.jfrog.io/ui/repos/tree/General/libs-snapshot-local/bio/terra/tanagra-vumc-admin-client) -- the core Tanagra client library is just above it (`tanagra-client`).

But I added a [GHA](https://github.com/DataBiosphere/tanagra-vumc-admin/blob/main/.github/workflows/bump-tag-publish.yaml) to both repos that on PR merge:
- Bumps the version in `settings.gradle`
- Publishes the client library to Maven with that same version number
- Creates a GitHub release tag with that same version number
- Attaches the OpenAPI yaml file to the GitHub release.

So source of truth for the latest client library version is the Maven repo, but when everything is working as intended, you should be able to just look at the latest GitHub release tag in the repo.

tjennison-google: I see. Thanks.

PR #320, opened 2023-01-13
Add exportInstances endpoint

melissachang:
- BigQuery writes results of SQL query to GCS file
- Service creates signed URL of GCS file. exportInstances returns signed URL

Manually tested.

Added to application.yml:

```
  export:
    gcs-bucket-project-id: broad-tanagra-dev
    gcs-bucket-name: broad-tanagra-dev-bq-export
```

If either property isn't set when export endpoing is called, an exception is thrown.


docs/CONTRIBUTING.md
```diff
@@ -102,3 +111,8 @@ To regenerate the golden files, run the tests with the `generateSqlFiles` Gradle
 ```
 ./gradlew cleanTest test --tests bio.terra.tanagra.* --info -PgenerateSqlFiles=true
 ```
+
```
marikomedlock, line 114: It looks like there's a `.CONTRIBUTING.md.swp` binary file that also got pushed -- do you need that one?

marikomedlock (approved):

melissachang:
@tjennison-google FYI backend is done, UI is unblocked

PR #321, opened 2023-01-13
Vumc GitHub publishing

NickofNasa:
Added VUMC github specific publishing block to the API service, this will allow publishing to the VUMC jar repo once a token is provided to the databiosphere repository.

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -105,8 +105,6 @@ dependencies {
         exclude group: "org.junit.vintage", module: "junit-vintage-engine"
     }
 
-    // Cloud SQL
-    implementation "com.google.cloud.sql:postgres-socket-factory:1.7.0"
 
     // Static analysis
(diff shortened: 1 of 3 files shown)


service/build.gradle
```diff
@@ -105,6 +105,9 @@ dependencies {
         exclude group: "org.junit.vintage", module: "junit-vintage-engine"
     }
 
+    // Cloud SQL
+    implementation "com.google.cloud.sql:postgres-socket-factory:1.7.0"
```
marikomedlock, line 109: Could you pull this dependency out into a separate PR please? I want to make sure we do some testing on our end before merging, and don't want to block the publishing change you made.

NickofNasa: removed this implementation from this pull request

melissachang (approved):

PR #322, opened 2023-01-17
Override max hierarchy depth. Parallelize indexing jobs.

marikomedlock:
- Allow overriding the max hierarchy depth, since this is typically much less than the default (64). I got the new max hierarchy depths from the [AoU-CB bash scripts](https://github.com/all-of-us/workbench/tree/main/api/db-cdr/generate-cdr): condition 20, ingredient 8, measurement 11, procedure 19.
- Allow running indexing jobs in parallel or serial. Serial is what we have currently. Parallel runs each entity's jobs concurrently, and then when all have finished, runs each entity group's jobs concurrently. There may also be concurrency within each entity/group -- jobs are defined in stages, any job in the same stage can be run concurrently.

Also some smaller fixes:
- Include the underlay name in the Apache Beam job name, so we can see it in the Dataflow Cloud Console.
- Handle null values in the modifers workflow.
- Add some Distinct steps to the ancestor-descendant workflow. This seems to have fixed the overflow cartesian product errors I was seeing when running this for ingredients.
- Print out a summary of the indexing jobs at the end.
- Remove the `invalid_reason IS NULL` condition from the brand definition in the SDD dataset, per VUMC feedback.

### DIFF
--- a/docs/INDEXING.md
+++ b/docs/INDEXING.md
@@ -80,7 +80,7 @@ Potential downside is more permissions needed by indexing service account.)
 Set the default application credentials to a service account key file that has read access to the source and read + 
 write access to the index data.
 ```
-export GOOGLE_APPLICATION_CREDENTIALS=/credentials/indexing_sa.json
+export GOOGLE_APPLICATION_CREDENTIALS=$HOME/tanagra/credentials/indexing_sa.json
 ```
 
 #### All Jobs
@@ -88,11 +88,11 @@ Do a dry run of all the indexing jobs. This provides a sanity check that the ind
 query inputs, are valid. This step is not required, but highly recommended to help catch errors/bugs sooner and without 
 running a bunch of computation first.
 ```
-./gradlew indexer:index -Dexec.args="INDEX_ALL /config/output/omop.json DRY_RUN"
+./gradlew indexer:index -Dexec.args="INDEX_ALL $HOME/tanagra/service/src/main/resources/config/output/omop.json DRY_RUN"
 ```
 Now actually kick off all the indexing jobs.
 ```
-./gradlew indexer:index -Dexec.args="INDEX_ALL /config/output/omop.json"
+./gradlew indexer:index -Dexec.args="INDEX_ALL $HOME/tanagra/service/src/main/resources/config/output/omop.json"
 ```
 This can take a long time to complete. If e.g. your computer falls asleep or you need to kill the process on your
 computer, you can re-run the same command again. You need to check that there are no in-progress Dataflow jobs in the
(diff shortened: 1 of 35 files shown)


indexer/src/main/java/bio/terra/tanagra/indexing/jobexecutor/JobRunner.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+
+public abstract class JobRunner {
+  protected static final long MAX_TIME_PER_JOB_MIN = 60;
+  protected static final long MAX_TIME_PER_JOB_DRY_RUN_MIN = 5;
+
+  public static final String ANSI_RESET = "\u001B[0m";
```
tjennison-google, line 14: I would call these `TERMINAL_ESCAPE_RESET` or something similar to better indicate what they actually are.

service/src/main/resources/config/broad/aou_synthetic/expanded/entity/condition.json
```diff
@@ -201,7 +201,8 @@
               "column" : "concept_id"
             }
           }
-        }
+        },
+        "maxHierarchyDepth" : 20
```
tjennison-google, line 205: For someone configuring a new dataset, how would someone know this value? And what would be the consequences of getting it wrong?

tjennison-google (approved):

service/src/main/resources/config/broad/aou_synthetic/expanded/entity/condition.json
melissachang: You mentioned a script (AoU-CB bash scripts). Could you point to them?

Maybe add this documentation to HierarchyMapping.java.

Instead of setting it per underlay, could we just set it to something like 50 for all underlays? I'm guessing the purpose is to avoid infinite loop... ?

melissachang (approved):

service/src/main/resources/config/broad/aou_synthetic/expanded/entity/condition.json
marikomedlock: Unfortunately, no easy answer to "how would someone know this value" -- basically they have to know enough about the dataset to know how deep the hierarchy is.

If you get it wrong by overshooting, then the hierarchy will still be generated correctly, but the indexing job will take longer. If you get it wrong by undershooting, then the hierarchy won't be generated correctly -- it will be missing levels.

Previously, the max depth was set to 64 for all underlays. That worked fine and is still the default, but lowering the number to something more realistic makes the indexing jobs run faster. This shouldn't ever cause infinite loops, it's just that Apache Beam (doesn't support conditional steps)[https://issues.apache.org/jira/browse/BEAM-106], so we'll always try to compute the max hierarchy depth, even if there are no more parent-child relationships to expand. The only way to reduce the size of the indexing job is to set the max depth lower beforehand.

marikomedlock: Added the link to the AoU-CB bash scripts to the PR description.

Added a comment to `HierarchyMapping.DEFAULT_MAX_HIERARCHY_DEPTH` about how overshooting is ok, but undershooting is not.

docs/INDEXING.md
```diff
@@ -117,6 +117,14 @@ or entity group:
 All the entities in a group should be indexed before the group. The `INDEX_ALL` command ensures this ordering, but keep 
 this in  mind if you're running the jobs for each entity or entity group separately.
 
+#### Concurrency
+By default, the indexing jobs are run concurrently as much as possible. You can force it to run jobs serially by
+appending `SERIAL` to the command:
+```
+./gradlew indexer:index -Dexec.args="INDEX_ALL /config/output/omop.json DRY_RUN SERIAL"
```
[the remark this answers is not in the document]
marikomedlock: Prefixed all paths with `$HOME/tanagra` in this file to make it clear they are absolute paths.

service/src/main/resources/config/broad/aou_synthetic/expanded/entity/condition.json
marikomedlock: It should be possible to give the user feedback on the max depth:
- if they overshot, then compute the max hierarchy depth and report it out.
- if they undershot, then I'm not sure we can tell them what the depth is exactly, but we can tell them it's bigger than the value they chose/defaulted to.

This could be a good future improvement.

tjennison-google: Yeah, that would be a good improvement. The silent failure is the really bad problem. The documentation improvements are good for the short term at least.

marikomedlock: Added JIRA ticket for this https://precisionmedicineinitiative.atlassian.net/browse/TNG-94

PR #323, opened 2023-01-17
Update broad + verily underlays.

marikomedlock:
Updated the underlay config files for the broad (cms_synpuf, aou_synthetic) and verily (cms_synpuf, sdd) environments, after running the indexing jobs successfully there. The verily sdd underlay is new -- it's a copy of the VUMC one. Now that we have a Verily SA with permissions on the VUMC `sd_static` dataset, we can index it in Verily's env and deploy it to our dev instance.

Pending are 2 more underlay config file updates: verily (aou_synthetic), vumc (sdd). I will put those in a follow-on PR.


tjennison-google (approved):

melissachang (approved):
This PR adds verily/sdd. We already have a vumc/sdd. Could we choose one and delete the other?

marikomedlock:
> This PR adds verily/sdd. We already have a vumc/sdd. Could we choose one and delete the other?

Unfortunately, I think we need both for now. The `verily/sdd` underlay is the one we'll deploy in the Verily dev instance -- the index dataset lives in `verily-tanagra-dev` GCP project. The `vumc/sdd` underlay is the one VUMC will deploy in their dev instance -- the index dataset lives in `victr-tanagra-test` GCP project. In the future, we should consider making the underlay config files more modular, so that we can e.g. have just one set of the data mapping files (`entity/`, `entitygroup/`, `sql/`, `ui/` sub-dirs) that are referenced from multiple top-level underlay files, possibly pointing to different BQ datasets.

PR #325, opened 2023-01-19
Update verily/aou_synthetic, vumc/sdd underlay config files.

marikomedlock:
- Specified the `maxHierarchyDepth` in the verily/cms_synpuf and verily/aou_synthetic underlays, set to the same values as the broad underlays.
- Renamed the verily/aou_synthetic underlay from `verily_aou_synthetic` -> `aou_synthetic`.
- Re-ran indexing for the verily/aou_synthetic underlay (the verily/cms_synpuf one I re-ran earlier).
- Updated the vumc/sdd underlay to match the verily/sdd files, but did not re-run the indexing there -- will let VUMC know to re-run on their end.


service/src/main/resources/config/verily/aou_synthetic/expanded/aou_synthetic.json
```diff
@@ -1,10 +1,10 @@
 {
-  "name" : "verily_aou_synthetic",
+  "name" : "aou_synthetic",
```
tjennison-google, line 2: I'm a little nervous about this because it would prevent the verily and broad versions from coexisting in the same deployment. Thoughts?

marikomedlock: I think the verily/broad/vumc versions won't ever coexist in the same deployment, mainly because of permissions and devops responsibility. I've indexed the SDD dataset into a verily plato project for our dev instance to read from, so that VUMC can handle their own dataset and deployment as they see fit (e.g. currently the index dataset lives in the same project as the source dataset, where we have read access, but they may decide to move the index dataset to their deployment project in the future).

In the longer term, I think we should decouple the data mapping and UI config files from the top-level underlay file. So basically we'd have one version of all the files in the `entity/`, `entitygroup/`, `sql/`, `ui/` sub-directories for each dataset (sdd, aou, cms). Then each deployment would only have one top-level file that points to data mapping they way, and the actual BQ datasets they live in. Most of this work is just config restructuring, but I think we may also need some kind of versioning of the data mapping files, so that 2 deployments can be on different versions. We're sidestepping that problem now by just every deployment having their own set.

PR #326, opened 2023-01-19
Add genotyping entity, genotyping_person entity group

melissachang:
This is for being able to create cohort based on genotyping platform.

I created `sdstatic_index_011923` and reran indexer.

I added a Platform button when creating cohort:

image

When I click the button, I get an error:

image

All the other domains (Condition, Procedure, Observation, Drug) are associated with an occurrence (eg condition occurrence). I didn't create an occurrence for Platform, because I'm guessing user doesn't care if there are duplicate person-platform pairs. The UI assumes all domains have an occurrence. Changing this was non-trivial, so I'll leave for @tjennison-google .


service/src/main/resources/config/verily/sdd/expanded/entity/platform_occurrence.json
```diff
@@ -0,0 +1,251 @@
+{
```
marikomedlock, line 1: I see expanded files for `platform_occurrence` entity and `platform_person_occurrence` entity group, but no original files. Did you want to go with the `GROUP_ITEMS` or `CRITERIA_OCCURRENCE` entity group type? You can delete the expanded files that you don't need anymore.

service/src/main/resources/config/verily/sdd/original/entity/platform.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "platform",
```
marikomedlock, line 2: [optional] Maybe consider changing the name to "genotyping" or "assay"?

service/src/main/resources/config/verily/sdd/original/entity/platform.json
```diff
@@ -0,0 +1,21 @@
...
+  "idAttribute": "id",
+  "attributes": [
+    { "type": "SIMPLE", "name": "id" },
+    { "type": "SIMPLE", "name": "name" }
+  ],
+
+  "sourceDataMapping": {
```
marikomedlock, line 9: In the SDD UI, this entity has a hierarchy. I think the parent-child relationships live in the `genotype_criteria` table. You can define a `hierarchyMapping` in this source mapping block.

tjennison-google:
> I didn't create an occurrence for Platform, because I'm guessing user doesn't care if there are duplicate person-platform pairs

I'm not quite clear on what this data represents but I think it's referring to the platform on which their genome was sequenced? In that case, can it be sequenced more than once and have that data accessible? If so, then it needs to be represented as an occurrence. If not, it should be configured like the demographic data, which has no occurrences, and it should function correctly regardless of what category it's put it.

service/src/main/resources/config/verily/sdd/original/entity/platform.json
melissachang: Changed to "genotyping" since that's what SD Discover UI has. Reindexed.

service/src/main/resources/config/verily/sdd/original/entity/platform.json
melissachang: In the Wed meeting, I'll ask them if they want hierarchy. Regardless, I'll do in a separate PR.

melissachang (commented):
Can I do tests in a separate PR?

> I'm not quite clear on what this data represents but I think it's referring to the platform on which their genome was sequenced? In that case, can it be sequenced more than once and have that data accessible? If so, then it needs to be represented as an occurrence.

From talking to @marikomedlock , if the user doesn't *care* about duplicates, then we don't need genotyping_occurrence. And I'm guessing they don't care. For example, when creating a cohort, they might care that a person had 2 amputations instead of 1. But for genotyping platform, they just want people that had at least one sample on the platform; they don't care if there are one or 2 samples for a given platform.

marikomedlock:
> Can I do tests in a separate PR?
Yes, sorry meant to say that in the comment.

> > I'm not quite clear on what this data represents but I think it's referring to the platform on which their genome was sequenced? In that case, can it be sequenced more than once and have that data accessible? If so, then it needs to be represented as an occurrence.
> 
> From talking to @marikomedlock , if the user doesn't _care_ about duplicates, then we don't need genotyping_occurrence. And I'm guessing they don't care. For example, when creating a cohort, they might care that a person had 2 amputations instead of 1. But for genotyping platform, they just want people that had at least one sample on the platform; they don't care if there are one or 2 samples for a given platform.
Maybe worth asking VUMC about this. The tables don't have much information about the sequencing event (e.g. a date). It makes me wonder whether they actually do have this information but it's somewhere else (maybe in a table they didn't load?). Even if we don't handle that info today, knowing it exists might incline us towards defining an occurrence.

marikomedlock:
I think there are 2 concerns here:
- How do we define this on the backend so the right queries can get generated for the instances/counts/hints endpoints? We need a relationship between person & genotyping. You can currently do that either with a CRITERIA_OCCURRENCE entity group, which requires defining a genotyping_occurrence entity, or a GROUP_ITEMS entity group, which does not. If there are no duplicate genotyping_occurrences and there really is no additional information about them, I don't know that it makes sense from a backend perspective to define an occurrence entity.
- How does the UI handle this? Currently the UI handles either demographic info (=attributes on person) or domains (=CRITERIA_OCCURRENCE entity groups). If we use GROUP_ITEMS, I assume that would require new UI code to handle the entity filter building.

@tjennison-google does that sound right to you? Maybe worth discussing over GVC on Mon.

> If not, it should be configured like the demographic data, which has no occurrences, and it should function correctly regardless of what category it's put it.

RE demographic data: I think there can be multiple genotyping runs per person, so the cardinality wouldn't be the same as e.g. gender. I guess if there's few enough of them, we could define an attribute of the person that's a list of all the genotyping platforms they've ever been sequenced by. Not sure whether that's the right approach.

tjennison-google:
I have been thinking about this and keep changing my mind on parts.

> I think there are 2 concerns here:
> 
> * How do we define this on the backend so the right queries can get generated for the instances/counts/hints endpoints? We need a relationship between person & genotyping. You can currently do that either with a CRITERIA_OCCURRENCE entity group, which requires defining a genotyping_occurrence entity, or a GROUP_ITEMS entity group, which does not. If there are no duplicate genotyping_occurrences and there really is no additional information about them, I don't know that it makes sense from a backend perspective to define an occurrence entity.

The best thing about using an occurrence is that it'll work like all of the other data we have so far. The only downside I've come up with other than it being overkill is that it may not interact correctly with time related modifiers, though we haven't thought about how to handle those anyways so that doesn't seem too important at this point.

> * How does the UI handle this? Currently the UI handles either demographic info (=attributes on person) or domains (=CRITERIA_OCCURRENCE entity groups). If we use GROUP_ITEMS, I assume that would require new UI code to handle the entity filter building.

There would need to be new UI code to handle this method. However, I think it would just be a matter of supporting classifications on `person` like on occurrences and refactoring some code. I haven't thought through it thoroughly but it's maybe not that bad if we really thought this was the way to go.

> 
> @tjennison-google does that sound right to you? Maybe worth discussing over GVC on Mon.

Yeah, probably worth discussing if you have time. I feel like the biggest question is what makes sense to support as a data type for building a generic system. Would it be valuable to have support for this in other situations? Because these ultimately are occurrences. They happened at specific points in time and we just might not have access to the data. So it's more a question of behaviors than the actual data.

> 
> > If not, it should be configured like the demographic data, which has no occurrences, and it should function correctly regardless of what category it's put it.
> 
> RE demographic data: I think there can be multiple genotyping runs per person, so the cardinality wouldn't be the same as e.g. gender. I guess if there's few enough of them, we could define an attribute of the person that's a list of all the genotyping platforms they've ever been sequenced by. Not sure whether that's the right approach.

I also considered this and I think it makes the most sense conceptually. I'm not sure what this would look like from a query perspective though or if it would have performance implications. From the UI perspective this probably looks the same or close to any other enum.

melissachang (commented):
Discussed offline. We will not add genotyping_occurrence entity. We needed condition_occurrence entity because condition occurrences could be in concept set. For genotyping platform, we're guessing that does not need to be in concept set. (We confirmed with VUMC that neither SNP nor SNP occurrence needs to be in concept set.)

@tjennison-google PTAL

tjennison-google (approved):

PR #329, opened 2023-01-20
Add snp entity, snp_person entity group

melissachang:
For creating a cohort with SNP criteria.


marikomedlock (commented):
Could you add some tests for generated SQL against this new entity+group? That will give us a generated SQL query we can copy/paste into the BQ cloud console to test out what the UI will display. It will also make sure we don't inadvertently change the generated SQL and break the UI without realizing.

ditto for the other PRs that add new config.

melissachang:
Can I do tests in a separate PR? (Thanks for the example test PR btw.)

marikomedlock:
> Can I do tests in a separate PR? (Thanks for the example test PR btw.)

Yes right that's what I meant -- should have noted that in the comment sorry. Np, thanks for tackling all these config changes!

marikomedlock:
I think the same discussion applies here as for genotyping -- whether to starting adding criteria that use the GROUP_ITEMS entity group or just stick to CRITERIA_OCCURRENCE.

melissachang:
> I think the same discussion applies here as for genotyping -- whether to starting adding criteria that use the GROUP_ITEMS entity group or just stick to CRITERIA_OCCURRENCE.

Discussed offline. Will stick with GROUP_ITEMS entity group. @tjennison-google PTAL.

tjennison-google (approved):

PR #334, opened 2023-01-31
Add endpoint for exporting annotations

melissachang:
**Manual testing**

- Create study
- Create cohort [snp = `RS12925749`](https://github.com/DataBiosphere/tanagra/pull/333#issue-1555781947) (356 people)
- Create annotations `A1`, `A2`
- Create review R1 with `size=400` to get all 356 people, review R2 with `size=400`
- Add annotation values:
  ```
  person_id    16804   17681   25259    32638   35089
  A1 R1        V1      V1               V1
  A1 R2        V2              V1       V1
  A2 R1                                         V3
  ```
- CSV file written:
  ```
  person_id,A1,A2
  16804,V2,
  17681,V1,
  32638,V1,
  35089,,V3
  25259,V1,
  ```

**Also:**

- I ended up not needing a request body. I don't need `entityFilter` because there is a `review_instance` -- with entity instance id -- for each entity in review. (I don't think `entityFilter` is needed for `listReviewInstancesAndAnnotations`.) I don't need `includeAttributes` because I'm not returning any concept set info.
- If gcsBucketProjectId or gcsBucketName aren't set in application.yaml, an exception is thrown

### DIFF
--- a/service/src/main/java/bio/terra/tanagra/app/controller/AnnotationsV2ApiController.java
+++ b/service/src/main/java/bio/terra/tanagra/app/controller/AnnotationsV2ApiController.java
@@ -27,7 +27,7 @@ import bio.terra.tanagra.service.accesscontrol.ResourceIdCollection;
 import bio.terra.tanagra.service.artifact.Annotation;
 import bio.terra.tanagra.service.artifact.AnnotationValue;
 import bio.terra.tanagra.service.utils.ToApiConversionUtils;
-import java.util.Collection;
+import com.google.common.collect.Table;
 import java.util.List;
 import java.util.stream.Collectors;
 import org.apache.commons.lang3.RandomStringUtils;
@@ -76,9 +76,10 @@ public class AnnotationsV2ApiController implements AnnotationsV2Api {
   public ResponseEntity<ApiExportFile> exportAnnotationValues(String studyId, String cohortId) {
     accessControlService.throwIfUnauthorized(
         SpringAuthentication.getCurrentUser(), READ, COHORT, new ResourceId(cohortId));
-    Collection<AnnotationValue> values =
+    Table<String, String, String> latestValues =
         annotationService.getAnnotationValuesForLatestReview(studyId, cohortId);
-    String gcsSignedUrl = annotationService.writeAnnotationValuesToGcs(values);
+    String gcsSignedUrl =
+        annotationService.writeAnnotationValuesToGcs(studyId, cohortId, latestValues);
     return ResponseEntity.ok(new ApiExportFile().gcsSignedUrl(gcsSignedUrl));
   }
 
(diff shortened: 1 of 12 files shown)


service/src/main/java/bio/terra/tanagra/app/controller/AnnotationsV2ApiController.java
```diff
@@ -69,6 +72,18 @@ public ResponseEntity<ApiAnnotationV2> createAnnotation(
         toApiObject(annotationService.getAnnotation(studyId, cohortId, newAnnotationId)));
   }
 
+  @Override
+  public ResponseEntity<ApiExportFile> exportAnnotationValues(String studyId, String cohortId) {
+    accessControlService.throwIfUnauthorized(
+        SpringAuthentication.getCurrentUser(), READ, COHORT, new ResourceId(cohortId));
```
tjennison-google, line 78: This permission seem weird because the cohort permission gives you access to all of the review data as a whole (and nothing else about cohorts?) but not to individual reviews. Maybe renaming to `ANNOTATION_EXPORT` or something like that would make it seem more sensible?

service/src/main/java/bio/terra/tanagra/service/AnnotationService.java
```diff
@@ -138,10 +163,119 @@ public AnnotationValue updateAnnotationValue(
...
+        .forEach(
+            entityInstanceId -> {
+              StringBuilder row = new StringBuilder(entityInstanceId);
+              annotations.forEach(
+                  annotation -> {
+                    String value =
+                        values.contains(entityInstanceId, annotation.getDisplayName())
+                            ? values.get(entityInstanceId, annotation.getDisplayName())
+                            : "";
+                    row.append("," + value);
```
tjennison-google, line 264: These need to be escaped because they could contain commas.

service/src/main/java/bio/terra/tanagra/service/AnnotationService.java
```diff
@@ -138,10 +163,119 @@ public AnnotationValue updateAnnotationValue(
...
+                  });
+              fileContents.append(String.format(row + "\n"));
+            });
+
+    String projectId = tanagraExportConfiguration.getGcsBucketProjectId();
+    String bucketName = tanagraExportConfiguration.getGcsBucketName();
+    String fileName = "tanagra_export_annotations_" + System.currentTimeMillis();
+    if (StringUtils.isEmpty(projectId) || StringUtils.isEmpty(bucketName)) {
+      throw new SystemException(
```
tjennison-google, line 273: Would it be helpful to log `fileContents` here so it's not necessary to set up GCS for debugging purposes?

melissachang: Done

service/src/main/java/bio/terra/tanagra/app/controller/AnnotationsV2ApiController.java
melissachang: Discussed offline. Left `READ` since it's cleaner to use `READ` rather than a separate `EXPORT` permission.

Changed to `ANNOTATION`. (Is that what you wanted?) In whole file, changed `ANNOTATION` to check cohort instead of annotation, since we don't plan on enforcing access control at annotation granularity.

service/src/main/java/bio/terra/tanagra/service/AnnotationService.java
melissachang: Discussed offline. Changed to tsv, so people don't have to deal with double quotes. If there's a problem with annotation values having tab character, I can escape later.

service/src/main/java/bio/terra/tanagra/app/controller/AnnotationsV2ApiController.java
tjennison-google: Yeah, this is exactly what I was thinking.

tjennison-google (approved):

PR #335, opened 2023-02-01
Add hierarchy to genotyping entity

melissachang:
If this looks ok, I'll make corresponding changes to vumc config.


service/src/main/resources/config/verily/sdd/expanded/entity/genotyping.json
```diff
@@ -306,7 +330,34 @@
...
+          "tablePointer" : {
+            "rawSql" : "/* Match the platform ids in platform.sql */ SELECT 101 AS id UNION ALL SELECT 102 UNION ALL SELECT 103 "
+          },
+          "fieldPointers" : {
+            "id" : {
+              "column" : "id"
+            }
+          }
+        },
+        "maxHierarchyDepth" : 1
```
tjennison-google, line 358: I would have expected 2 here since there are two levels, but maybe this is more technically the number of passes of the hierarchy algorithm?

tjennison-google (approved):

service/src/main/resources/config/verily/sdd/expanded/entity/genotyping.json
melissachang: Interesting. I think `maxHierarchyDepth=1` makes sense (or at least is one possible interpretation). If it was `maxHierarchyLevels`, that would be 2.

PR #336, opened 2023-02-02
[SDD] Add BioVU attributes to person entity

melissachang:
See `verily-tanagra-dev.sdstatic_index_020123.person`.

Before I merge, I will:

- Update vumc config
- Run `INDEX_ALL`

In source dataset, `person` has `x_sample` column that mostly -- but not always -- matches up with presence of row in `x_biovu_sample_status`. So I did not use `person.x_sample`.

image


service/src/main/resources/config/verily/sdd/expanded/entity/person.json
```diff
@@ -94,29 +94,54 @@
...
-        "count" : 960195
+        "count" : 960199
       }, {
         "enumVal" : {
           "value" : {
             "int64Val" : 38003564
           },
           "display" : "Not Hispanic or Latino"
         },
-        "count" : 2375593
```
tjennison-google, line 113: It seems surprising that these counts changed when you were just adding some new columns. Any idea why?

tjennison-google (approved):

service/src/main/resources/config/verily/sdd/expanded/entity/person.json
melissachang: Good catch! It turns out `x_biovu_sample_status` had duplicate rows for some people. For example, person 4587323 had 11 duplicate rows. (I spot-checked a couple other people; rows were duplicates.)

I changed `person.sql` to take out duplicate rows from `x_biovu_sample_status`. These count changes reverted.

PR #341, opened 2023-02-09
Convert cohort and concept set editing to use backend

tjennison-google:
* The cohort/concept set currently being edited are stored in context to be available where necessary, similar to how they used to be available from redux.
* Adds intermediate routing nodes that inject the appropriate contexts across all cohort/concept set children.
* Removes dependencies on redux except for underlays, which will be similarly removed in a future PR.
* Adds the ability to set the criteria of a concept set during creation.
* Integration tests are changed to use real backends because mocking the necessary backend APIs within the current framework is impractical. A future PR will remove the fakes and potentially augment or replace the backend with Cypress or MSW based tests.

### DIFF
--- a/.github/actions/start-local-server/action.yml
+++ b/.github/actions/start-local-server/action.yml
@@ -1,0 +1,47 @@
+name: "start-local-server"
+description: "Launch a tanagra server"
+outputs:
+  tanagrapid:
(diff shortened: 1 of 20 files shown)


melissachang (approved):
Yay for tests using real backend :)

PR #342, opened 2023-02-13
GHA for keeping configs in sync, for example vumc/sdd and verily/sdd

melissachang:
GHA is a short-term solution. Correct long-term solution is to restructure the configs so that there is a single `entity/`, `entitygroup/`, etc for SDD.

I didn't do long-term solution in this PR because there are 2 complications:

- The sql files refer to BQ dataset, and so are different. To fix this, replace BQ dataset name with a token like `SOURCE_DATASET`.
- `sdd.json` should be split into two parts:
  1. Parts that are different between vumc/sdd, verily/sdd, ie `dataPointers`
  2. Parts that are the same between vumc/sdd, verily/sdd, ie everything besides `dataPointers`

Also, for the configs that were out of sync, this PR makes them in sync.


tjennison-google (approved):
I don't like this but I guess it'll do as a short term solution.

melissachang:
> I don't like this but I guess it'll do as a short term solution.

Yeah, I tried doing the long-term solution, then realized it was more complicated than I thought. So this is a stop-gap. It was getting hard to keep track of all the differences.

PR #343, opened 2023-02-15
Fix hierarchy links sometimes navigating to stale pages

tjennison-google:
This bug could cause criteria to be added or edited in unexpected places because clicking on the view hierarchy button would sometimes navigate to a previously visited page. This was caused by the button being cached by SWR and the setSearchParams call internally storing a reference to stale state via a useLocation call.

* Change entity fetching to store only the hierarchy information and not the button that links to the view.
* Add an option to the TreeGrid to completely replace the content of a cell.
* Change TreeGrid rowCustomization to return a list instead of a Map because the Map constructor couldn't nicely handle non-literal values for merging multiple column customizations.


melissachang (approved):
Is there an integration test that would have caught this bug?

PR #345, opened 2023-02-17
[SDD] Add `age_at_occurrence` columns to occurrence entities

melissachang:
There are 2 ways to do this:

1. In config, manually add age_at_occurrence attribute to occurrence entity. Manually write SQL to subtract date of birth from occurrence start date. See #337 for visit_occurrence.
2. Change Tanagra indexing to automatically compute age_at_occurrence. Add an optional field to entity config: `sourceStartDateColumn`. If this field is set, occurrence index table will have age_at_occurrence column. See [here](https://github.com/DataBiosphere/tanagra/pull/345/files#diff-4a4e8c2263f3588a8bdd9b6bcda5589ac43044c3d6a9167c7ab4dbc6d920b98b) for ingredient_occurrence.

This PR does 2. I think 2 is more in the spirit of Tanagra. Age at occurrence is useful for many datasets (OMOP and not-OMOP). 2 makes it easier to configure age at occurrence ([this](https://github.com/DataBiosphere/tanagra/pull/345/files#diff-4a4e8c2263f3588a8bdd9b6bcda5589ac43044c3d6a9167c7ab4dbc6d920b98b) is simpler than [this](https://github.com/DataBiosphere/tanagra/pull/337/files).)

In this PR:

- If an original occurrence entity config has `sourceStartDateColumn`, `age_of_occurrence` attribute is added to expanded entity config. See Indexer.maybeAddAgeAtOccurrenceAttribute().
- Added new indexing job ComputeAgeAtOccurrence. This job runs if entity has `age_of_occurrence` attribute. This job does the calculation (subtract date of birth from occurrence start date).

Also:
- Revert #337. It's no longer needed with this PR.

Manually tested
- DRY_RUN works
- Inspect BQ, eg `verily-tanagra-dev.sdstatic_index_022723.condition_occurrence`
  - If occurrence config doesn't have `sourceStartDateColumn`, index occurrence table doesn't have 
  `age_of_occurrence` column
  - If occurrence config has `sourceStartDateColumn`:
    - Number rows in index condition_occurrence the same (before and after this PR)
    - Index occurrence `age_at_occurrence` is correct. (Person 1107050, condition occurrence)
    - If person doesn't have source condition_occurrence rows, person doesn't have index condition_occurrence rows (Person 746031)
    - If person has no DOB and source occurrence rows, index occurrence rows are there and age_at_occurrence is blank (Person 4259330, condition_occurrence)
    - If person has DOB and source occurrence rows with no start date, index occurrence row doesn't have age_at_occurrence (Person 120556 has several observation occurrences. Only observation occurrence 92954694 missing start date.)
- Swagger. Called `/v2/underlays/sdd/entities/person/instances` with `age_at_ocurrence` > 60:
  ```
    {
      "includeAttributes": [
        "id"
      ],
      "filter": {
        "filterType": "BOOLEAN_LOGIC",
        "filterUnion": {
          "booleanLogicFilter": {
            "operator": "AND",
            "subfilters" : [
              {
                "filterType": "RELATIONSHIP",
                "filterUnion": {
                  "relationshipFilter": {
                    "entity": "condition",
                    "subfilter": {
                      "filterType": "ATTRIBUTE",
                      "filterUnion": {
                        "attributeFilter": {
                          "attribute": "id",
                          "operator": "EQUALS",
                          "value": {
                            "dataType": "INT64",
                            "valueUnion": {
                              "int64Val": 4002818
                            }
                          }
                        }
                      }
                    }
                  }
                }
              },
              {
                "filterType": "RELATIONSHIP",
                "filterUnion": {
                  "relationshipFilter": {
                    "entity": "condition_occurrence",
                    "subfilter": {
                      "filterType": "ATTRIBUTE",
                      "filterUnion": {
                        "attributeFilter": {
                          "attribute": "age_at_occurrence",
                          "operator": "GREATER_THAN",
                          "value": {
                            "dataType": "INT64",
                            "valueUnion": {
                              "int64Val": 60
                            }
                          }
                        }
                      }
                    }
                  }
                }
              }
            ]
          }
        }
      }
    }
  ```
  Returned sql:
   ```
    SELECT
      p.id AS id
    FROM
      `verily-tanagra-dev.sdstatic_index_022723`.person AS p
    WHERE
      (p.id IN (
        SELECT
          i.id_person
        FROM
          `verily-tanagra-dev.sdstatic_index_022723`.idpairs_condition_person AS i
        WHERE
          i.id_condition IN (
          SELECT
            c.id
          FROM
            `verily-tanagra-dev.sdstatic_index_022723`.condition AS c
          WHERE
            c.id = 4002818))
        AND p.id IN (
        SELECT
          i.id_person
        FROM
          `verily-tanagra-dev.sdstatic_index_022723`.idpairs_condition_occurrence_person AS i
        WHERE
          i.id_condition_occurrence IN (
          SELECT
            c.id
          FROM
            `verily-tanagra-dev.sdstatic_index_022723`.condition_occurrence AS c
          WHERE
            c.age_at_occurrence > 60)))
   ```
   This does not return person 1365473, who does not have any `age_at_ocurrences > 60`. If I change to `age_at_ocurrences > 40`, person 1365473 is returned.


indexer/src/main/java/bio/terra/tanagra/indexing/job/ComputeAgeAtOccurrence.java
```diff
@@ -0,0 +1,409 @@
...
+  public String getName() {
+    return "COMPUTE AGE AT OCCURRENCE (" + getEntity().getName() + ")";
+  }
+
+  @Override
+  public void run(boolean isDryRun) {
+    // Wait for DenormalizeEntityInstances to run.
+    if (!isDryRun) {
+      LOGGER.info("Waiting for DenormalizeEntityInstances to run for {}", getEntity().getName());
+      JavaUtils.retryUntilTrue(
```
tjennison-google, line 71: This is only 75s. If they're triggered at the same time, couldn't it take way longer than that for DenormalizeEntityInstances to complete? Or am I misunderstanding?

tjennison-google (commented):
A bit of a tangent, but having an `age_at_occurrence` attribute is an argument in favor of participant ages being calculated by the backend as well.

indexer/src/main/java/bio/terra/tanagra/indexing/job/ComputeAgeAtOccurrence.java
melissachang: Good point. Just now, I indexed `meaurement_occurrence`, which is one of the biggest tables with 1.4 billion rows and  17G (physical bytes). `ComputeAtAtOccurrence` failed because `DenormalizeEntityInstances` wasn't done in time. `DenormalizeEntityInstances` took 3-4 minutes. I changed this retry for 5 mins.

melissachang (commented):
Say backend computes age.

- On Mar 1, person with age 50 is exported to a notebook
- Person's birthday is Mar 2
- On Mar 3, person is now age 51. But the notebook still has age 50.

`age_at_occurrence` doesn't have this problem because it's computing age for a specific point in time.

tjennison-google:
> Say backend computes age.
> 
> * On Mar 1, person with age 50 is exported to a notebook
> * Person's birthday is Mar 2
> * On Mar 3, person is now age 51. But the notebook still has age 50.
> 
> `age_at_occurrence` doesn't have this problem because it's computing age for a specific point in time.

I hadn't thought about it from the notebook perspective before. If anything, I think that's actually an argument in favor of the backend computing it still because it keeps the exported data consistent with the query that generated it. Otherwise the calculated age changes while the rest of the data is static.

tjennison-google (approved):
Don't forget to fix the failing check.

melissachang:
> > Say backend computes age.
> > 
> > * On Mar 1, person with age 50 is exported to a notebook
> > * Person's birthday is Mar 2
> > * On Mar 3, person is now age 51. But the notebook still has age 50.
> > 
> > `age_at_occurrence` doesn't have this problem because it's computing age for a specific point in time.
> 
> I hadn't thought about it from the notebook perspective before. If anything, I think that's actually an argument in favor of the backend computing it still because it keeps the exported data consistent with the query that generated it. Otherwise the calculated age changes while the rest of the data is static.

Interesting, let's discuss at a standup

PR #350, opened 2023-03-07
Support # of occurrence dates modifier

melissachang:
In OMOP, there might be multiple rows in `condition_occurrence` for one conceptual occurrence. Consider this `condition_occurrence` table:
```
person_id   condition   start_date
100         500         10/1/2020
100         500         10/1/2020
101         500         11/1/2020
101         500         11/2/2020
```

Say I only want people with > 1 occurrence. I want person 101 and not 100. What I really want is people > 1 **occurrence date**.

I added `group_by` fields to RelationshipFilter, to support filtering on # occurrence dates. I made it generic so one can group by other fields if needed.

People with > 1 occurrence date of condition 4002818:

```
{
  "includeAttributes": [
    "id"
  ],
  "filter": {
    "filterType": "RELATIONSHIP",
    "filterUnion": {
      "relationshipFilter": {
        "entity": "condition_occurrence",
        "subfilter": {
          "filterType": "ATTRIBUTE",
          "filterUnion": {
            "attributeFilter": {
              "attribute": "condition",
              "operator": "EQUALS",
              "value": {
                "dataType": "INT64",
                "valueUnion": {
                  "int64Val": 4002818
                }
              }
            }
          }
        },
        "group_by_count_attribute": "start_date",
        "group_by_count_operator": "GREATER_THAN",
        "group_by_count_value": 1
        }
      }
    }
  }
}
```

```
SELECT
  p.id AS id
FROM
  `verily-tanagra-dev.sdstatic_index_022723`.person AS p
WHERE
  p.id IN (
  SELECT
    x.person_id
  FROM (
    SELECT
      c.person_id AS person_id,
      c.start_date AS start_date
    FROM
      `verily-tanagra-dev.sdstatic_index_022723`.condition_occurrence AS c
    WHERE
      c.condition = 4002818
    GROUP BY
      c.person_id,
      c.start_date) AS x
  GROUP BY
    x.person_id
  HAVING
    COUNT(*) > 1)
```
This returns 1502 people, [as expected](https://user-images.githubusercontent.com/10929390/223526622-243cc350-437d-40ea-8b88-6d90b15eee0b.png).

[Person 402 has 2 condition_occurrences](https://user-images.githubusercontent.com/10929390/223526501-11c0e3f5-df81-45f7-bbd8-f86b78a9c48c.png) for condition 4002818 on 2020-05-22. The above query does not return person 402.


service/src/main/resources/api/service_openapi.yaml
```diff
@@ -1466,12 +1465,25 @@ components:
...
         entity:
           type: string
         subfilter:
           $ref: "#/components/schemas/FilterV2"
+        group_by_count_attribute:
+          type: string
+        group_by_count_operator:
+          $ref: "#/components/schemas/BinaryOperatorV2"
+        group_by_count_value:
+          $ref: "#/components/schemas/LiteralV2"
```
tjennison-google, line 1486: This should be an integer rather than a literal, since it's being compared to the count.

melissachang: Done, thanks.

tjennison-google (approved):

PR #356, opened 2023-03-14
Continue basic cohort review support

tjennison-google:
* Load occurrence data for an individual entity in a cohort review.
* Add support for cohort review plugins and add a simple plugin to display a customizable table of occurrence data.
* Use more correct date functions and fix date handling in treegrid.
* Add option to set listData limit.
* Fix loading spinner layout in some instances.
* Fix word wrapping in treegrid fields to break on work boundaries and fix ellipses on non-link cells.
* Split integrtion tests into separate files so they can be run separately since they take a bit to run.


ui/src/app.tsx
```diff
@@ -43,12 +43,43 @@ export default function App() {
...
+            },
+          ];
+
           return {
             name: underlay.name,
             displayName: underlay.displayName ?? underlay.name,
             primaryEntity: underlay.primaryEntity,
             entities: entitiesRes.entities,
-            uiConfiguration: JSON.parse(underlay.uiConfiguration),
+            //uiConfiguration: JSON.parse(underlay.uiConfiguration),
```
melissachang, line 81: Delete?

melissachang (approved):
Yay for separate test files.

What is the rationale for cohort review plugin, is this functionality that not all underlays will want? (An underlay may want a cohort review, but may not want the stuff in the plugin?)

ui/src/app.tsx
tjennison-google: Oops, done. I didn't mean to send this quite yet. I've moved the UI config changes into the underlays and re-expanded them.

tjennison-google:
> 

The plugins are so that deployments can have custom UI for reviewing custom data types, just like for building cohorts. It's the same idea behind as the BioVU filter UI for cohort building applied to cohort review.

PR #358, opened 2023-03-16
Add document criteria plugin

tjennison-google:
* All note text fields are null, so search the titles for now. That's passable for cohort building but we'll definitely need to fix for cohort review.
* Only the SDD underlay has notes, so we'll need to figure out another way to make this usable with synpuf and aou since they're our primary test datasets.
* Update the local dev script to reference the current Verily underlays.


ui/src/criteria/search.tsx
melissachang, line 1: classification.tsx has a bunch of search stuff. How is that different from this file? Maybe clarify with comments, or by renaming things.

melissachang (approved):
Integration test?

ui/src/criteria/search.tsx
tjennison-google: Both of them have searches/queries as part of the UI but they're named based on the type of criteria they create. `classification.tsx` uses the hierarchies and groupings, `search.tsx` uses a text search, `attribute.tsx` uses the value of a single attribute. Do you have particular clarification you're looking for? Or maybe this information as comments would be sufficient?

tjennison-google:
> Integration test?

I didn't add this to the integration tests because it would require running against the Verily underlays at this point which would be inconvenient if we're more typically using the Broad underlays locally. I also can't do the cohort review side of things because the data isn't there. Once we solve those issues I'll loop back to finish documents and add tests.

ui/src/criteria/search.tsx
melissachang: Maybe rename `search.tsx` to `textSearch.ts`, and add comments

melissachang (approved):
Re intregration test - SG. But just curious, don't verily/sdd and vumc/sdd have the same features, since their configs file are in sync? (It's confusing when features get added to one and not the other, since one is used locally and the other on test/dev.)

tjennison-google:
> Maybe rename search.tsx to textSearch.ts, and add comments

Done.

> Re intregration test - SG. But just curious, don't verily/sdd and vumc/sdd have the same features, since their configs file are in sync? (It's confusing when features get added to one and not the other, since one is used locally and the other on test/dev.)

There are actually three separate sets of underlays, designated by where the data lives. The `broad` ones are what we typically use for local development and are also used for GHAs, the `verily` ones are used for dev/test, and the `vumc` ones are used for actual VUMC deployments, though they probably have a fork in their own repo they deploy from. At this point, the similarly named underlays `aou_synthetic`, `cms_synpuf`, and `sdd` should be the same across all environments except for resource locations. I changed both the `verily/sdd` and `vumc/sdd` versions in this PR.

It would be good to have a `broad/sdd` to make testing easier. Mariko and I were actually just talking about that this morning and thinking we should ask if we can at least get permissions for the GH runners to access the sdd data for integration tests.

It would also be good to refactor the underlays so that we can have different resource locations without needing entirely separate copies of them but that's another task entirely.

PR #367, opened 2023-03-30
OpenAPI schema definition: criteria group + collection v3

marikomedlock:
- Add an array of text tags to `CriteriaV2`.
- Defined a new `CriteriaGroupV3` object to represent a primary criteria and associated modifiers.
- Defined a new `CriteriaGroupCollectionV3` object to represent a set of criteria groups. This is essentially equivalent to the previous `CriteriaGroupV2` object.

This PR just includes schema changes. I will handle backend changes in a follow-on PR once we agree on these new objects.


service/src/main/resources/api/service_openapi.yaml
```diff
@@ -1923,6 +1923,66 @@ components:
         - operator
         - excluded
 
+    CriteriaGroupCollectionV3:
```
tjennison-google, line 1926: I don't love this, though I don't hate it either. The double group-collection is a bit awkward. Some other ideas:

- CriteriaBlock/CriteriaGroupBlock
- CriteriaSection/CriteriaGroupSection <-- maybe my favorite?

- EntityFilterGroup
  +- EntityFilter
    +- Criteria

marikomedlock: I'd like to not reuse "EntityFilter" for this object, to avoid confusion with defining a query. This is more of a UI object, at least for now while we're building the filter on the front end. "Section" SGTM.

tjennison-google (approved):

PR #368, opened 2023-03-30
Have GHA index broad/aou_synthetic, broad/cms_synpuf every night

melissachang:
Index will be written to special nightly dataset; index dataset in json is undisturbed. Example: In `expanded/cms_synpuf.json`, index dataset will be temporarily changed from `cmssynpuf_index_011523` to  `cms_synpuf_nightly`. So `cmssynpuf_index_011523` is not disturbed.

Sample run: https://github.com/DataBiosphere/tanagra/actions/runs/4570226139/jobs/8067319893

I made a resuable action `expand-config-and-index`. Here's the workflow that calls the action. Here are the parts VUMC would change to index SDD:

image

Next PR will do Slack integration.


marikomedlock (approved):
This is great, thanks for the reusable action piece in particular.

PR #369, opened 2023-04-03
Make annotation_value PK id -> review + id.

marikomedlock:
- Merged the annotation value create + update endpoints, and forced the annotation value id = entity instance id.
- The update annotation value endpoint now only allows the caller to update the literal value itself. Previously, it also allowed updating the entity instance id, but I think it's less confusing to just delete and create a new one, in that case.
- Changed the primary key of the annotation value table from just the `annotation_value_id` field, to `annotation_value_id + review_id`. This accommodates the caller specifying the `annotation_value_id` without requiring it to be unique across all reviews.


tjennison-google (commented):
We had also discussed having a single CreateOrUpdate endpoint that merges these calls. That would make it easier than having to differentiate between the two if it's not too difficult to change.

tjennison-google (approved):

PR #372, opened 2023-04-05
Add highlighting and categories to docs review plugin

tjennison-google:
* Refactor hint data based selects into new component.
* Adjust select theming so empty selects have the same height as non-empty ones.


marikomedlock (approved):

marikomedlock:
Could you add some screenshots to your UI PRs? Just for my benefit as I try and grok the React code. Not meant to be a burden, just would be much appreciated when easy to include them!

tjennison-google:
> Could you add some screenshots to your UI PRs? Just for my benefit as I try and grok the React code. Not meant to be a burden, just would be much appreciated when easy to include them!

> Will do, not a problem. I think I had intended to do that a while ago and forgot at some point.

PR #374, opened 2023-04-12
Groundwork for modifiers and source hierarchies

tjennison-google:
* Differentiate between criteria occurrences used for filters and those used for outputs to allow for multiple outputs or outputs from alternate hierarchies.
* Separate relationship filters into their own object. This allow on relationship filter to apply to a set of criteria (i.e. modifiers).
* Temporarily insert relationship filters based on each criteria until CriteriaGroups are ready.

### DIFF
--- a/ui/src/data/source.tsx
+++ b/ui/src/data/source.tsx
@@ -31,6 +31,7 @@ import {
   isArrayFilter,
   isAttributeFilter,
   isClassificationFilter,
+  isRelationshipFilter,
   isTextFilter,
   isUnaryFilter,
 } from "./filter";
@@ -587,7 +588,7 @@ export class BackendSource implements Source {
         entityName: this.config.primaryEntity.entity,
         countQueryV2: {
           attributes: groupByAttributes,
-          filter: generateFilter(this, filter, true) ?? undefined,
+          filter: generateFilter(this, filter) ?? undefined,
         },
       })
     );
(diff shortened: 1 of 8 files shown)


ui/src/data/source.tsx
```diff
@@ -1333,101 +1312,98 @@ function generateOccurrenceFilter(
...
 
+function arrayFilterOperator(
+  filter: ArrayFilter
+): tanagra.BooleanLogicFilterV2OperatorEnum {
+  if (!isValid(filter.operator.min) && !isValid(filter.operator.max)) {
+    return tanagra.BooleanLogicFilterV2OperatorEnum.And;
+  }
+  if (filter.operator.min === 1 && !isValid(filter.operator.max)) {
+    return tanagra.BooleanLogicFilterV2OperatorEnum.Or;
+  }
```
marikomedlock, line 1402: Is this function kind of overloading the `filter.operator` property to represent both min/max and or/and? Not objecting, just trying to follow along. And if so, why did you choose to do that instead of add new filter operators?

marikomedlock (approved):

ui/src/data/source.tsx
tjennison-google: I made a generic boolean filter with min and max with the idea we could someday support things like "at least 3 out of 5 are true" and "at most 2 out of 3 are true". AND ( x out of x) and OR (at least 1 out of x) are just special cases of that.

This function also isn't new or changed. The function after it got merged into the function before and the diff made a mess of this one.

PR #375, opened 2023-04-12
Prevent export from occurring before button is clicked

tjennison-google:
* Fix reload button on error messages and optionally show progress when retrying.


ui/src/components/loading.tsx
```diff
@@ -9,21 +9,27 @@ import React, { ReactNode, useEffect, useRef, useState } from "react";
...
   isLoading?: boolean;
+  isMutating?: boolean;
   mutate?: () => void;
 };
 
 type Props = {
   size?: string;
   status?: Status;
   children?: React.ReactNode;
+  showProgressOnMutate?: boolean;
```
marikomedlock, line 20: Does progress here indicate the number of retries (e.g. 4 out of max 10 retries) or the number of calls to the export endpoint (e.g. 2 out of 5 signed urls generated), or something else?

marikomedlock (approved):

ui/src/components/loading.tsx
tjennison-google: No, it's just the generic progress spinner that's shown while loading initially. In other cases, backend updates or retries like changing a cohort are done in the background but for export we explicitly want to show the request is in progress because the user is waiting for it.

PR #391, opened 2023-05-11
Update VUMC admin client

arychj:
* Repoint VUMC admin service at the new VUMC admin client library.
* Update test configuration to point at VUMC tanagra test instance.


service/src/test/java/bio/terra/tanagra/service/VumcAdminServiceTest.java
```diff
@@ -24,12 +25,14 @@ public class VumcAdminServiceTest {
   @Autowired private VumcAdminService vumcAdminService;
 
   @Test
+  @Disabled
```
marikomedlock, line 28: I think these are the only two tests in this class -- perhaps move up the `@Disabled` annotation up to the class level.

service/src/test/java/bio/terra/tanagra/service/VumcAdminServiceTest.java
```diff
@@ -24,12 +25,14 @@ public class VumcAdminServiceTest {
...
   void version() throws ApiException {
     SystemVersion systemVersion = vumcAdminService.version();
     LOGGER.info("version returned success: {}", systemVersion);
   }
 
   @Test
+  @Disabled
```
marikomedlock, line 35: Please add a comment about why these tests are disabled and how to run them locally, e.g.

`@Disabled("VUMC admin service OAuth client id is not checked into this repo. You can run this test locally by setting the vumc-admin properties in application-test.yaml.")`

service/src/test/resources/application-test.yaml
```diff
@@ -14,7 +14,7 @@ tanagra:
     password: dbpwd
 
   vumc-admin:
-    base-path: https://tanagra-dev.api.verily.com
+    base-path: https://tanagra-test.app.vumc.org
     oauth-client-id: 694046000181-baqla782scqfaaga26869orhhd8o7vjo.apps.googleusercontent.com
```
marikomedlock, line 18: The URL and OAuth client id here both refer to a specific deployment of the admin service, so I would expect both to change together.

Given what you said before about not wanting to check the OAuth client id for the VUMC admin service deployment into the core public repo, I think it would make sense to leave these 2 properties as is (i.e. pointing to the Verily admin service deployment) and add a comment above how to set these properties when running locally (e.g. VUMC admin service URL is ... and you can get the OAuth client id from the console or talking to a VUMC data apps eng...?)

marikomedlock (approved):

PR #393, opened 2023-05-16
Switch VUMC primary user identifier back to email address

arychj:
It looks like we're going to have to switch back to using user emails at least temporarily as we are unable to retrieve user profile information from the Google Directory API for users that are not a member of the organization.


service/src/main/java/bio/terra/tanagra/service/accesscontrol/impl/VumcAdminAccessControl.java
```diff
@@ -29,7 +29,7 @@ public String getDescription() {
   public boolean isAuthorized(
       UserId userId, Action action, ResourceType resourceType, @Nullable ResourceId resourceId) {
     return vumcAdminService.isAuthorized(
-        userId.getSubject(),
+        userId.getEmail(),
```
marikomedlock, line 32: nit: Can you add a comment to these two `VumcAdminService` methods describing what the method expects to be passed -- the subject id or the email?

marikomedlock (approved):

PR #394, opened 2023-05-16
Use SQL query for age_at_occurrence and visit_type.

marikomedlock:
- Replaced the table name pointer with a table SQL query file for each of the occurrence entities. The SQL query includes the `age_at_occurrence` and `visit_type` fields, both of which require joins (to `person` and `visit_occurrence`) with other tables and so are not possibly to express with the limited config filters.
- Removed the `ComputeAgeAtOccurrence` Dataflow job. These properties are now computed by the existing `DenormalizeEntityInstances` job.
- Flagged some attributes as do-not-compute-display-hint to make the expanding go a littler faster (e.g. range of `person_id`).

I have made this change for the `verily/sdd` and `vumc/sdd` underlays and re-expanded them. I have also re-indexed the `verily/sdd` underlay.


tjennison-google (approved):
This is a fine approach for now, but it is unfortunate that so many entities have to switch to SQL to handle age at occurrence, which is probably a very common attribute. Longer term it seems like we would want some in-between solution, where either there are more attribute types to encapsulate common calculated fields, or the ability to include custom SQL for individual fields, or something along those lines.

marikomedlock:
> This is a fine approach for now, but it is unfortunate that so many entities have to switch to SQL to handle age at occurrence, which is probably a very common attribute. Longer term it seems like we would want some in-between solution, where either there are more attribute types to encapsulate common calculated fields, or the ability to include custom SQL for individual fields, or something along those lines.

Actually we already have the ability to include custom SQL for individual fields, but it only handles `JOIN`ing with a single field (e.g. `concept_id JOIN to concept` to get the `concept_name). The `age_at_occurrence` is more complicated because it requires a `JOIN` plus a second field (i.e. `person_id JOIN to person` to get the `birth_datetime` and then compare that to the `condition_occurrence_start_datetime`).

Computing this with SQL (either in a raw SQL file or the simple config file) is simpler than doing it with a dedicated Dataflow job, so regardless of where we end up, I think this is a step in the right direction. Longer term, agreed it would be nice to not force a raw SQL file for a current date type calculation. I think this would be straightforward to add, but given our push to beta, this is not the right priority now.

PR #396, opened 2023-05-17
Add verily/sdd_refresh0323 underlay.

marikomedlock:
- Add support for `TIMESTAMP` literals in the API. Some of the BQ columns changed data type from `DATE` -> `TIMESTAMP` in the refresh. Not sure if there's also some UI work to handle this new datatype (@tjennison-google)
- Add a new underlay `verily/sdd_refresh0323`, original and expanded.


tjennison-google (approved):
This won't currently work in the UI because it's returning values in new fields that the UI doesn't know about so it obviously isn't parsing them. The new data types can be trivially mapped onto the existing ones though, `TIMESTAMP -> Date` and `DOUBLE -> number`. However, that loses some type information so I'm not sure what would happen when that data is returned to the backed for queries. This leads to some questions:

1) Javascript doesn't actually have a built in `int64` type, only `number` which is actually a `double`. I've been using that for `int64` attributes since that's what the OpenAPI generator generates which is technically incorrect but works for real world numbers. If I passed back something like a `person_id` as a `double`, that would still work in the generated SQL since they're both numbers?
2) Similarly for `timestamp`, will it work to treat all dates as timestamps, ie. passing back a timestamp to compare against date fields? If not, is it easy for the backend to truncate the time portion of a timestamp is used in a filter on a date field?

Fully propagating the new data types is probably too much work for beta given what else I have to do but could be done later.

marikomedlock:
Per offline discussion about how to handle new datatypes for beta:
- I added cases for `timestamp` and `double` datatypes to the UI literal parsing function.
- We're hoping this is enough for beta, where we only expect to display these new data types. In the future, the UI may also want to filter on attributes with those data types, in which case there is probably additional UI handling needed.

There are some rollup count indexing jobs that have not been run for this sdd refresh underlay, but I'd like to merge this PR and tackle those in a follow-on one. I'll update our `test` deployment with this latest code so others can start looking at it.

tjennison-google:
> * I added cases for `timestamp` and `double` datatypes to the UI literal parsing function.

UI changes look good.

tjennison-google (approved):

tjennison-google (approved):

PR #397, opened 2023-05-17
Fix client library publish GHA.

marikomedlock:
- Use a GitHub personal access token (PAT) when building the client library, so the GHA can pull the VUMC admin service client library from their GitHub package repo (GPR). GPR requires a PAT even for public packages. The PAT is one that I generated with my own GH identity, with only permission to read public packages, and stored as a GH repository secret.
- Ignored the nightly indexing GHA for now.

### DIFF
--- a/.github/actions/expand-config-and-index/action.yaml
+++ b/.github/actions/expand-config-and-index/action.yaml
@@ -37,6 +37,7 @@ runs:
         echo "Running:  $cmd"
         eval ${cmd}
       env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         GOOGLE_APPLICATION_CREDENTIALS: ../${{ inputs.google-application-credentials-relative-path }}
       shell: bash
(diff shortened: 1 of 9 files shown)


marikomedlock:
> Is there no way to create a token for the runner instead?

Not sure what you mean. A token for the entire job instead of setting the env var for a specific step?

tjennison-google:
> > Is there no way to create a token for the runner instead?
> 
> Not sure what you mean. A token for the entire job instead of setting the env var for a specific step?

I mean so it's not your personal token.

marikomedlock:
> > > Is there no way to create a token for the runner instead?
> > 
> > 
> > Not sure what you mean. A token for the entire job instead of setting the env var for a specific step?
> 
> I mean so it's not your personal token.

Figured out how to do this! I read in the docs how to do it, but seems like writing the Maven settings file in addition to setting the runner token confused it. Removed all references to my token.

PR #399, opened 2023-05-18
Store resource id parts separately. Use study id for all child artifacts for VUMC access control.

marikomedlock:
- Changed the `ResourceId` field to store different parts of the ID separately (e.g. review id = study-cohort-review), instead of just a single string.
- Added named `ResourceId` constructors to make it clearer which type of resource the id refers to.
- Added a dedicated `ResourceIdCollcetion.empty` constructor to replace the special  null handling.
- Updated the `VUMCAccessControl` class to just use the study id for all children artifacts (e.g. cohorts).


arychj:
What is the base type of the IDs used for these resources? My understanding was that everything was going to be UUIDs? If that's the case, what is the need to maintain the ancestry as part of the resource id? If not, what is the format and length of the components of the resource id, and what is the expected maximum depth of a resource id (right now it looks like 3)? I need to make sure the field size in the admin app can accommodate the composite keys.

arychj (approved):

marikomedlock:
> Just looked at the PR again, it looks like you're only passing one component to access control, so none of this matters.

Yes, I changed it to only pass either the underlay or study id, so you don't have to parse anything on your side. I think I got the logic right, but please let me know if you see any errors when you're testing with your admin service and happy to debug together.

PR #400, opened 2023-05-18
Use join library for rollup counts indexing job.

marikomedlock:
- Rewrite rollup counts logic to use a JOIN library and subsequent processing. This is an attempt to improve the reliability of the indexing jobs, especially around OOM errors.
- Enable the rollup counts indexing jobs for the `GroupItems` entity groups. This is to provide rollup counts for the `snp` and `genotyping` entities.
- Fix the `indexer` sub-project tests to use JUnit 4, which is required by Apache Beam. All the other sub-projects are using JUnit 5, and when I bumped this version, I didn't realize that the beam tests were being silently skipped.
- Add a generated `id` attribute to the `snp` entity. The generated `id` is an `INT64` which is the only id data type that the indexing jobs handle. The previous `id` attribute (`snp_id` column in BQ source table) is now the `name` and is a `STRING`. This change is so that the compute rollup counts Beam job runs successfully without needing to introduce different type handling -- I plan to do that eventually, but this is just to get beta out the door.
- Updated the `verily/sdd_refresh0323` underlay and re-indexed it into a new BQ dataset.
- Added a `vumc/sdd_refresh0323` underlay that's a copy of the verily version, with just the Verily-specific projects etc. changed in the index dataset.


tjennison-google (approved):

service/src/main/resources/config/verily/sdd_refresh0323/original/ui/top_level.json
```diff
@@ -265,7 +265,7 @@
       "id": "tanagra-snp",
       "title": "SNP variant",
       "category": "Program data",
-      "columns": [{ "key": "id", "width": "100%", "title": "SNP variant" }],
```
marikomedlock, line 268: @tjennison-google Here's the change I made to the UI config so that the new `name` attribute (previously the `id` attribute) is displayed in the list. PTAL.

PR #406, opened 2023-05-19
Add config for Labs and measurements, and modifiers

tjennison-google:
* Values for Labs and measurements need to work a bit differently and will follow separately.


service/src/main/resources/config/verily/sdd/original/ui/top_level.json
```diff
@@ -183,7 +208,8 @@
         { "key": "t_rollup_count", "width": 120, "title": "Roll-up count" }
       ],
       "occurrence": "observation_occurrence",
-      "classification": "observation"
+      "classification": "observation",
+      "modifiers": ["age_at_occurrence", "visit_type", "date_group_by_count"]
```
marikomedlock, line 212: Does the `date_group_by_count` modifier refer to the `measurement_occurrence.date` attribute? And if so, why is it different from the other two with the `group_by_count` suffix?

marikomedlock (approved):

service/src/main/resources/config/verily/sdd/original/ui/top_level.json
tjennison-google: These refer to the `modifierConfigs` defined below. The difference between them is that `condition_occurrence` and `ingredient_occurrence` have `start_date` and `end_date` while the others just have a `date` so they are group by different attributes.

PR #410, opened 2023-05-19
Add support for age, labs and measurements

tjennison-google:
* Change year_of_birth to age throught. Remove chart hacks.
* Refactor slider with range into a shared component.


service/src/main/resources/config/verily/sdd/original/ui/top_level.json
```diff
@@ -324,7 +334,7 @@
       "id": "tanagra-snp",
       "title": "SNP variant",
       "category": "Program data",
-      "columns": [{ "key": "id", "width": "100%", "title": "SNP variant" }],
+      "columns": [{ "key": "name", "width": "100%", "title": "SNP variant" }],
```
marikomedlock, line 337: I haven't updated the `snp` entity in the `sdd` underlays yet, only the `snp_refresh0323` one, so I think this will fail.

ui/src/criteria/classification.tsx
```diff
@@ -440,8 +504,150 @@ function ClassificationEdit(props: ClassificationEditProps) {
...
+          ...data.valueData,
+          attribute,
+          numeric: !!newHintData?.integerHint,
+        };
+
+        if (data.valueData.range.min === 0 && data.valueData.range.max === 0) {
+          data.valueData.range = {
+            id: "",
+            min: newHintData?.integerHint?.min ?? 0,
+            max: newHintData?.integerHint?.max ?? 1000,
```
marikomedlock, line 575: Is this the default integer range when there's no hint data?

marikomedlock (approved):

service/src/main/resources/config/verily/sdd/original/ui/top_level.json
tjennison-google: Right, reverted.

ui/src/criteria/classification.tsx
tjennison-google: The user shouldn't be able to get to this point if there's no hint data but the compiler can't know that so I have to provide fallback values.

PR #412, opened 2023-05-24
Tanagra-Aou-Integration-Updated for AoU changes

chenchals:
The changes include:
- adding mariadb dependency to build.gradle
- creating a copy of liquibase schema change file and adding tag dbms for liqubase to switch based on portgres or mariadb
- updating column data type for mariadb:
  - postgres text -> longtext
  - postgres text[] -> longtext
  - postgres jsonb -> longtext
  - postgres all _id columns from text -> varchar(50)
- including both postgres and mariadb files in  channgelog.xml
- Successfully ran api test for getting person entity from BQ 
image


service/src/main/resources/db/changesets/20230410_schema_reset.yaml
```diff
@@ -441,25 +473,25 @@ databaseChangeLog:
...
               - column:
                   name: selection_data
-                  type: text
+                  type: ${text.type}
                   constraints:
                     nullable: false
               - column:
                   name: ui_config
-                  type: text
+                  type: ${text.type}
```
marikomedlock, line 489: These two fields are serialized json and may be longer than 255 characters. What do you think about using `longtext` for these fields in MySql?

service/src/main/resources/db/changesets/20230530_avoid_postgres_specific_features.yaml
```diff
@@ -19,12 +35,12 @@ databaseChangeLog:
                   remarks: Deleting a study will cascade to delete its properties
               - column:
                   name: key
-                  type: text
+                  type: ${_text.type}
```
marikomedlock, line 38: This column is part of the unique constraint, should it be `id.type` instead?

service/src/main/resources/db/changesets/20230530_avoid_postgres_specific_features.yaml
```diff
@@ -1,7 +1,23 @@
 databaseChangeLog:
+  - property:
+      dbms: postgresql
+      name: _id.type
```
marikomedlock, line 4: Why are the names prefixed with an underscore in this changeset file and not in the previous one? I think it would be easier to follow if they are the same in both places.

service/src/main/resources/db/changesets/20230530_avoid_postgres_specific_features.yaml
```diff
@@ -76,12 +92,12 @@ databaseChangeLog:
                   remarks: Deleting a concept set will cascade to delete the criteria contained in it
               - column:
                   name: key
-                  type: text
+                  type: ${_text.type}
```
marikomedlock, line 95: Same as above. This column is part of the unique constraint, does it need to be `id.type`?

service/src/main/resources/db/changesets/20230530_avoid_postgres_specific_features.yaml
```diff
@@ -104,17 +120,17 @@ databaseChangeLog:
             columns:
               - column:
                   name: enum
-                  type: text
+                  type: ${_text.type}
```
marikomedlock, line 123: Same as above. This column is part of the unique constraint, should it be `id.type`?

marikomedlock (commented):
Thanks for this work! It's great to see it working in the tests.

service/src/main/resources/db/changesets/20230410_schema_reset.yaml
chenchals: Several alternatives: varchar(65xxx), longtext can hold up to 4GB data

service/src/main/resources/db/changesets/20230530_avoid_postgres_specific_features.yaml
chenchals: yes, if it is part of unique constraint, varchar(255) is OK too if key length if >50.

service/src/main/resources/db/changesets/20230530_avoid_postgres_specific_features.yaml
chenchals: no reason, not sure if property names must be unique across changelogs. I will remove the `_` and check

service/src/main/resources/db/changesets/20230530_avoid_postgres_specific_features.yaml
chenchals: yes, if it is part of unique constraint, varchar(255) is OK too if key length if >50.

service/src/main/resources/db/changesets/20230530_avoid_postgres_specific_features.yaml
chenchals: yes, if it is part of unique constraint, varchar(255) is OK too if key length if >50.

service/src/main/resources/db/changesets/20230530_avoid_postgres_specific_features.yaml
chenchals: maybe there is also a global properties file... I will check.

service/src/main/resources/db/changesets/20230410_schema_reset.yaml
marikomedlock: OK either sounds fine: longer varchar (maybe 25k is enough) or longtext. We can always change the datatype in the future if we find we're hitting the limit. I think we'd hit the 255 limit fairly quickly though.

marikomedlock (approved):

PR #415, opened 2023-05-29
Continue UX style overhaul

tjennison-google:
* Align fonts with VDS and update variants used to match mocks.
* Overhaul cohort page to align with mocks. Details of critera and headers still incomplete.
* More rounded corners on controls and paper.
* Replace empty PNG with an SVG version.
* Add saving indicator.


tjennison-google:
image

ui/src/cohortContext.ts
```diff
@@ -62,6 +68,14 @@ export function useNewCohortContext() {
     }
     setState(newState);
     await source.updateCohort(studyId, newState.present);
+
+    setState(
+      produce(newState, (state) => {
+        state.lastSaved = new Date();
```
marikomedlock, line 74: The backend also stores a `lastModified` property of the cohort. I don't think it makes much difference whether you use that one or this UI-generated one for the save state display. Noting in case it's easier to work with.

marikomedlock (approved):

ui/src/cohortContext.ts
tjennison-google: No, good call. That's better since it reflects when changes were actually made.

PR #420, opened 2023-05-31
First pass at export endpoints: list impls, call impl.

marikomedlock:
Draft PR for discussion. Do not merge.

### DIFF
--- a/service/src/main/resources/api/service_openapi.yaml
+++ b/service/src/main/resources/api/service_openapi.yaml
@@ -843,6 +843,50 @@ paths:
         500:
           $ref: "#/components/responses/ServerError"
 
+  "/v2/underlays/{underlayName}/exportImplementations":
+    parameters:
+      - $ref: "#/components/parameters/UnderlayNameV2"
+    get:
+      summary: List the export implementations available to an underlay
+      operationId: listExportImplementations
+      tags: [Export]
+      responses:
+        200:
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ExportImplementationList"
+        404:
+          $ref: "#/components/responses/NotFound"
+        500:
+          $ref: "#/components/responses/ServerError"
+
+  "/v2/studies/{studyId}/export":
+    parameters:
+      - $ref: "#/components/parameters/StudyIdV2"
+    post:
+      summary: Export entity instances and review annotations.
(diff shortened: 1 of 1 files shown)


service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2293,6 +2337,89 @@ components:
...
+                nullable: false
+              query:
+                $ref: "#/components/schemas/QueryV2"
+
+    ExportResult:
+      type: object
+      properties:
+        status:
+          type: string
+          enum: [ "RUNNING", "COMPLETE" ]
```
tjennison-google, line 2383: What does RUNNING mean?

service/src/main/java/bio/terra/tanagra/app/configuration/ExportConfiguration.java
```diff
@@ -0,0 +1,126 @@
...
+          m.getParams().stream().collect(Collectors.joining(",")));
+    }
+  }
+
+  public static class Shared {
+    private String gcsProjectId;
+
+    private List<String> gcsBucketNames;
+
+    public String getGcsProjectId() {
```
tjennison-google, line 62: Are these parameters not specific to individual plugins? They wouldn't be necessary for something like an AWS export, right?

tjennison-google (approved):

service/src/main/resources/api/service_openapi.yaml
marikomedlock: I was thinking we'll have export options that just kick something off (e.g. AoU genomic extraction job).

service/src/main/java/bio/terra/tanagra/app/configuration/ExportConfiguration.java
marikomedlock: Right, they're just shared infrastructure that all plugins can use. Equivalent to duplicating these parameters for each export model. I was thinking of this shared config as useful for terraform outputs, and the per-model config as useful for things like URLs/filename prefixes/etc.

service/src/main/resources/api/service_openapi.yaml
marikomedlock: I was thinking we'll have export options that just kick something off (e.g. AoU genomic extraction job).

tjennison-google (approved):

PR #422, opened 2023-06-02
Restyle tables to match new UX designs

tjennison-google:
* Use buttons for adding criteria instead of clicking on names. Show different text depending on whether the criteria is being added.
* Convert add criteria page to a table and tweak styling.
* Shrink study list text.
* Refactor spacing helper to a shared location.


tjennison-google:
image

PR #425, opened 2023-06-02
Updated with common schema files and seperated postgres data migration

chenchals:
- Updated liquibase schema files
  - copied existing file `202305_avoid...yaml` to  `20230602_schema_extend.yaml`
    - added `columnDataType` to `dropNotNullConstraint`
    - updated all constraint names with a `_2` suffix as there were name-collisions with named constraints in `20230410_sche_reset.yaml` file 
  - separated postgres data migration sqls into a different file with `dbms: postgresql` tag

### DIFF
--- a/service/src/main/resources/db/changesets/20230602_schema_extend.yaml
+++ b/service/src/main/resources/db/changesets/20230602_schema_extend.yaml
@@ -1,9 +1,7 @@
 databaseChangeLog:
   - changeSet:
-      id: avoid_postgres_specific_features
+      id: schema_extend
       author: marikomedlock, chenchalsubraveti
-      # TODO: Remove the jsonb and 2 text[] columns in a follow-on changeset, once we're sure this migration was successful.
-      # study.properties, criteria.tags, annotation_key.enumVals
       changes:
         - createTable:
             tableName: study_property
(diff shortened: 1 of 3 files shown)


service/local-dev/run_server.sh
```diff
@@ -21,7 +21,7 @@ while getopts ":av" arg; do
 done
 
 export TANAGRA_DATABASE_NAME=tanagra_db
-export TANAGRA_DB_INITIALIZE_ON_START=false
+export TANAGRA_DB_INITIALIZE_ON_START=true
```
marikomedlock, line 24: This is helpful for testing, but please leave it set to `false` in the main branch. This is the setting we'll deploy with, so I'd like to keep it as the testing default as well.

service/src/main/resources/db/changesets/20230602_schema_extend.yaml
```diff
@@ -1,9 +1,7 @@
 databaseChangeLog:
   - changeSet:
-      id: avoid_postgres_specific_features
+      id: schema_extend
```
marikomedlock, line 3: This changeset file has already been deployed to our test instance (we deploy nightly), so you can't change the name or any of the schema specification for postgres. If you change it to render differently for mariadb/mysql only, then I think that's okay (though I'd want to test to confirm) but changing the postgres rendering will definitely break things.

marikomedlock (changes requested):

service/local-dev/run_server.sh
chenchals: done!

service/src/main/resources/db/changesets/20230602_schema_extend.yaml
chenchals: I do not follow.

PR #426, opened 2023-06-02
Restyle datasets page

tjennison-google:
* Add link link font to the theme.
* Use larger, black checkboxes to match the mocks.
* Add support for non-text elements in the empty component.

### DIFF
--- a/ui/src/datasets.tsx
+++ b/ui/src/datasets.tsx
@@ -4,14 +4,16 @@ import Dialog from "@mui/material/Dialog";
 import DialogActions from "@mui/material/DialogActions";
 import DialogContent from "@mui/material/DialogContent";
 import DialogTitle from "@mui/material/DialogTitle";
-import Grid from "@mui/material/Grid";
-import IconButton from "@mui/material/IconButton";
 import Link from "@mui/material/Link";
 import Paper from "@mui/material/Paper";
 import Stack from "@mui/material/Stack";
-import Switch from "@mui/material/Switch";
+import Step from "@mui/material/Step";
+import StepConnector from "@mui/material/StepConnector";
+import StepLabel from "@mui/material/StepLabel";
 import Tab from "@mui/material/Tab";
 import Tabs from "@mui/material/Tabs";
+import ToggleButton from "@mui/material/ToggleButton";
+import ToggleButtonGroup from "@mui/material/ToggleButtonGroup";
 import Typography from "@mui/material/Typography";
 import ActionBar from "actionBar";
 import {
@@ -28,6 +30,8 @@ import { findEntity } from "data/configuration";
 import { Filter, makeArrayFilter } from "data/filter";
 import { useSource } from "data/source";
 import { useStudyId, useUnderlay } from "hooks";
+import { GridBox } from "layout/gridBox";
+import GridLayout from "layout/gridLayout";
 import React, {
   Fragment,
   ReactNode,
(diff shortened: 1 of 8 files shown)


tjennison-google:
image

ui/src/datasets.tsx
```diff
@@ -163,204 +171,304 @@ export function Datasets() {
...
-                        !excludedAttributes.get(occurrence.id)?.has(attribute)
-                      }
-                      onChange={() =>
-                        updateExcludedAttributes((selection) => {
-                          if (!selection?.get(occurrence.id)) {
-                            selection?.set(occurrence.id, new Set<string>());
+                  />
+                ) : (
+                  cohorts
+                    .filter((cohort) => cohort.underlayName === underlay.name)
```
marikomedlock, line 223: This is filtering the cohorts in a study to show only the ones for the current underlay, right? Maybe this is a good candidate for an optional filter on the cohort/concept set list endpoints. Similar to the metadata filter I added for studies.

ui/src/datasets.tsx
```diff
@@ -163,204 +171,304 @@ export function Datasets() {
...
+                                <>
+                                  <Link
+                                    variant="link"
+                                    underline="hover"
+                                    component={RouterLink}
+                                    to={absoluteNewConceptSetURL(params)}
+                                  >
+                                    Create a new data feature
+                                  </Link>{" "}
+                                  to define more criteria
```
marikomedlock, line 348: I don't quite follow this text -- create a new concept set to define more criteria?

ui/src/datasets.tsx
```diff
@@ -113,47 +118,50 @@ export function Datasets() {
     editable: boolean,
     conceptSets: { id: string; name: string }[]
   ) => {
-    return conceptSets.map((conceptSet) => (
-      <Stack key={conceptSet.name} direction="row" alignItems="center">
+    // TODO(tjennison): Only show demographics until we have better prepackaged
+    // concept sets.
```
marikomedlock, line 122: By removing the analgesics one from the UI config?

ui/src/datasets.tsx
```diff
@@ -163,204 +171,304 @@ export function Datasets() {
...
+                    variant="outlined"
+                    sx={{
+                      py: 2,
+                      display: "block",
+                      width: "100%",
+                      height: "100%",
+                    }}
+                  >
+                    <GridBox sx={{ px: 1, overflowY: "auto" }}>
+                      {workspaceConceptSets.data?.length === 0 ? (
```
marikomedlock, line 317: Do you call studies workspaces in the ui code?

marikomedlock (approved):

ui/src/datasets.tsx
marikomedlock: Hmm no not really. Is this from the UX recommendations? If so, happy to defer to them.

tjennison-google: Is it just this one or the cohort text too? As I was implementing this I realized that data features were never empty because of "Demographics" and so were never described anywhere. I'll see what Yikai's thoughts are as well.

marikomedlock: Yeah both are a little confusing to me. Dropping the "to define [more] criteria" from both would make more sense to me -- or just a big Create new button or something. The descriptions of cohorts and data features are clear, and I think sufficient.

PR #427, opened 2023-06-05
Align headers, clean up datasets text and tooltips

tjennison-google:
* Convert headers to use standard layout components.
* Add support for header subtitles and rearrange components to use them.
* Add tooltips to datasets page and remove "empty" message from data features list.

### DIFF
--- a/ui/src/datasets.tsx
+++ b/ui/src/datasets.tsx
@@ -1,4 +1,5 @@
 import AddIcon from "@mui/icons-material/Add";
+import InfoIcon from "@mui/icons-material/Info";
 import Button from "@mui/material/Button";
 import Dialog from "@mui/material/Dialog";
 import DialogActions from "@mui/material/DialogActions";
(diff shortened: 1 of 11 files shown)


ui/src/datasets.tsx
```diff
@@ -171,304 +173,275 @@ export function Datasets() {
...
-                    subtitle={
-                      <>
-                        <Link
-                          variant="link"
-                          underline="hover"
-                          onClick={showNewCohort}
-                          sx={{ cursor: "pointer" }}
+                <GridBox sx={{ px: 1, overflowY: "auto" }}>
+                  {cohorts.length === 0 ? (
+                    <Empty
```
marikomedlock, line 214: Do you need one of these `Empty` blocks for the concept sets too? Or maybe you left it out because we expect to always have prepackaged ones?

marikomedlock (approved):

ui/src/datasets.tsx
tjennison-google: Yeah, Demographics will always exist so it's not necessary.

PR #431, opened 2023-06-07
[DT-412] chenchals-AoU-tanagra-create-common schema files

chenchals:
Added changes for common schema files

- updated liquibase schema files
- added changeset file for mitigating postgres duplicate constraint names and column names
   - `20230606_postgres_rename_columns_and_constraints.yaml`
- added new changeset files with `dbms: mariadb,mysql` 
   - `20230606_schema_common_reset.yaml` an edited copy of `20230410_schema_reset.yaml` 
   - `20230606_schema_common_extend.yaml` an edited copy of `20230530_avoid_postgres_specific_features.yaml`
- updated `changelog.xml` file with properties for data-types across different dbms 
- updated column-names with `key` to appropriate: `property_key` or `criteria_key` and `value` to `property_value` or `criteria_value`
- changed the order for `OFFSET x LIMIT x` to `LIMIT x OFFSET x` to be compatible across mariadb and postgresql
- added scripts for creating mariaDB database for tanagra
  - `./service/local-dev/run_mariadb.sh`
  -  `./service/local-dev/local-mariadb-init.sql`
- updated `./service/local-dev/run_server.sh` with `flag -m` for using mariadb 
- updated `docs/CONTRIBUTING.md` with 
  - how to run tanagra service with mariadb or postgres
  - how to run tests with mariadb or postgres by using environmental variable `DBMS`
  - how to start the tanagra service with either postgres or mariadb as backend database
- updated `.github/workflows/service-test.yaml` with env variable `DBMS` set to postgresql


service/src/main/resources/db/changelog.xml
```diff
@@ -1,5 +1,21 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <!-- define global column types for dbms -->
+    <property global="true" dbms="postgresql" name="id.type" value="text"/>
+    <property global="true" dbms="postgresql" name="text.type" value="text"/>
+    <property global="true" dbms="postgresql" name="long.text.type" value="text"/>
+    <property global="true" dbms="postgresql" name="text.array.type" value="text[]"/>
+    <property global="true" dbms="postgresql" name="jsonb.type" value="jsonb"/>
```
marikomedlock, line 8: Do we need a text[] and jsonb type if those won't be used in the common schema?

service/src/main/resources/db/changesets/20230606_postgres_rename_columns_and_constraints.yaml
```diff
@@ -0,0 +1,59 @@
+databaseChangeLog:
+  - changeSet:
+      id: postgres_rename_columns_and_constraints
```
marikomedlock, line 3: Thanks for updating these constraint names.

service/src/main/resources/db/changesets/20230606_postgres_rename_columns_and_constraints.yaml
```diff
@@ -0,0 +1,59 @@
...
+        -  dropForeignKeyConstraint:
+             baseTableName:  criteria_tag
+             constraintName:  fk_crit_cs
+        -  addForeignKeyConstraint:
+             baseColumnNames:  concept_set_id
+             baseTableName:  criteria_tag
+             constraintName:  fk_crit_cs_2
+             onDelete:  CASCADE
+             referencedColumnNames:  id
+             referencedTableName:  concept_set
```
marikomedlock, line 44: I understand that you needed to rename the unique constraints because they included the renamed key->criteria_key column. I didn't expect that you'd also have to rename these FK constraints, since none of the relevant column names changed. Any idea why? Maybe the FK constraints somehow indirectly reference the old unique constraint? Just curious.

service/src/main/resources/db/changelog.xml
```diff
@@ -1,5 +1,21 @@
...
+    <property global="true" dbms="mariadb,mysql" name="long.text.type" value="longtext"/>
+    <property global="true" dbms="mariadb,mysql" name="text.array.type" value="longtext"/>
+    <property global="true" dbms="mariadb,mysql" name="jsonb.type" value="longtext"/>
+    <!-- define changesets postgres-only properties not used-->
     <include file="changesets/20230410_schema_reset.yaml" relativeToChangelogFile="true"/>
     <include file="changesets/20230530_avoid_postgres_specific_features.yaml" relativeToChangelogFile="true"/>
+    <include file="changesets/20230606_postgres_rename_columns_and_constraints.yaml" relativeToChangelogFile="true"/>
+    <!-- define changesets common-schema properties are used but dbms: tag set to mariadb,mysql-->
+    <include file="changesets/20230606_schema_common_reset.yaml" relativeToChangelogFile="true"/>
+    <include file="changesets/20230606_schema_common_extend.yaml" relativeToChangelogFile="true"/>
```
marikomedlock, line 20: What do you think about combining these 2 schema files? Given that we'll only deploy this to a fresh DB (now in mariadb/mysql, hopefully soon in postgres), I don't think there's any need to add the text[], jsonb columns + notnull constraints, only to immediately drop them.

marikomedlock (commented):
Thank you for all this work, especially the scripts to run locally and the documentation. Some comments inline.

service/src/main/resources/db/changesets/20230606_postgres_rename_columns_and_constraints.yaml
marikomedlock: Did you end up needing these after all?

service/src/main/resources/db/changesets/20230608_schema_common_reset.yaml
```diff
@@ -0,0 +1,693 @@
...
+                  type: ${text.type}
+                  constraints:
+                    nullable: false
+              # needed for test cases to run, when removing this columns
+              # fix test cases in StudyServiceTest
+              - column:
+                  name: properties
+                  type: ${jsonb.type}
+                  constraints:
+                    nullable: true
```
marikomedlock, line 55: Ah I think this is my bug -- leftover from when I replaced the properties column with a separate table. I think you can just remove "properties" from [this line](https://github.com/DataBiosphere/tanagra/blob/6e60dacabdab989ea71183f5accc0d2c0da476be/service/src/main/java/bio/terra/tanagra/db/StudyDao.java#L29) to fix, and then you won't need to keep this or the `jsonb.type` around.

marikomedlock (commented):
Looks great, thanks. Two comments inline. Once those are resolved, I'd like to do a quick test deploy before you merge -- given the unexpected error last time, I want to double check this won't break things during user testing.

service/src/main/resources/db/changesets/20230606_postgres_rename_columns_and_constraints.yaml
chenchals: I did not try to rename and check. Will do and resolve, it may not be needed as it deals with only postgres - but have not checked

chenchals: I did not try to rename and check. Will do and resolve, it may not be needed as it deals with only postgres - but have not checked

service/src/main/resources/db/changesets/20230608_schema_common_reset.yaml
chenchals: Remove this initially, and specifically broke I think the study tests as it needed fiields(?) that is why I put it back. Once the specific tests are fixed, we can drop this column. The other text[] columns are removed and did not affect test cases

marikomedlock (approved):
Look great, thanks again!

PR #446, opened 2023-06-19
Also search demographic fields (e.g. race) when adding criteria

tjennison-google:


ui/src/criteria/attribute.tsx
```diff
@@ -288,3 +295,38 @@ function AttributeInline(props: AttributeInlineProps) {
     </Loading>
   );
 }
+
+async function search(
```
marikomedlock, line 299: Can you confirm I've following this logic correctly?
- This function is fetching the hint information for the `person` attribute, and then sorting them by count.
- Previously, you never needed this because a user would select a single attribute and you'd just display all the values alphabetically or whatever, since there are few enough values.
- This PR is adding attribute search to the across-all-criteria search, and now you need the counts so you can slot these into the list with all the other domain matches.

Also, I was initially expecting a corresponding config change e.g. add demographics to list of criteria to search across. Is the idea that if a plugin implements a `search` function, then you'll automatically call that for the across-all-criteria search?

marikomedlock (approved):

ui/src/criteria/attribute.tsx
tjennison-google: That's basically correct. Previously, the hint values would only be shown in the drop down for a criteria and I would just display them in the order the API returned them. I could consider having counts and sorting there as well.

And yes, it just calls the search function on every plugin that implements it. It may make sense to have an option for that eventually but it hasn't been necessary so far.

PR #447, opened 2023-06-20
Add BioVU criteria plugin

tjennison-google:
* Don't force counts for null filters to be 0. In some cases, filters that don't add any restrictions resolve to a null filter (e.g. BioVU with "Any SD Records" selected) and they should match all participants instead of none. This isn't always correct though since an inverted (i.e. "Must not") filter should match 0 participants but the null filters are currently being collapsed together. There are a couple other similar edge cases that also need to be reconsidered but this fixes the most commonly visible case for now.


ui/src/plugins/vumc/biovu.tsx
```diff
@@ -0,0 +1,256 @@
...
+  ANY = "Any",
+  ONE_HUNDRED = "100",
+  FIVE_HUNDRED = "500",
+}
+
+const sampleFilterDescriptions = {
+  [SampleFilter.NONE]: "Any SD Records (not requiring BioVU resources)",
+  [SampleFilter.ANY]: "Any BioVU DNA (no minimum amount or concentration)",
+  [SampleFilter.ONE_HUNDRED]: "At least 100ng with at least ng/µL BioVU DNA",
+  [SampleFilter.FIVE_HUNDRED]: "At least 500ng with at least ng/µL BioVU DNA",
```
marikomedlock, line 33: Doesn't need to be this PR, but could you try and match these strings in a `search` function so this shows up in the across-all-criteria search?

marikomedlock (approved):

ui/src/plugins/vumc/biovu.tsx
tjennison-google: Done, no reason not to support that. It does bring up the consideration that there are no rollup counts for these options. Not particularly important at the moment but worth considering how we might solve that.

PR #453, opened 2023-06-21
Make file download list when exporting generic

tjennison-google:


ui/src/datasets.tsx
```diff
@@ -819,44 +819,58 @@ function ExportDialog(
...
           } else {
-            throw new Error(`Unknown output type ${parts[0]}`);
+            artifactLists.push([artifact]);
           }
         }
 
+        artifactLists.sort((a, b) => a[0].section.localeCompare(b[0].section));
+
+        // TODO(tjennison): Remove fallback for looking up cohort ids once the
+        // backend plugin returns them.
```
marikomedlock, line 852: Does this refer to the backend service or a plugin in the UI?

marikomedlock (approved):

ui/src/datasets.tsx
tjennison-google: The backend service. This is now displaying whatever the export plugin returns so the plugin should be updated to display what we want to see here rather than relying on the UI to do translation. I just added a step that checks if the name is a valid cohort id, in which case I look up the cohort name and display that instead. That's a bit of a hack though  so it would be better to remove it.

tjennison-google: I'm still optionally processing the `:`s to generate category headings, so probably something like:
```
Data:person -->  <person_file_url>
Data:condition_occurrence -->  <person_file_url>
Annotations:Women with diabetes -->  <annotation_file_url>
Annotations:Men with diabetes -->  <annotation_file_url>
```

I'll process that into:
```
Annotations
- Women with diabetes
- Men with diabetes
Data
- person
- condition_occurrence
```

Where each item in the lists is a link to the corresponding URL.

PR #454, opened 2023-06-23
Only fetch the spcified underlay

tjennison-google:
* Only the underlay select page that needs the full list of underlays fetches them all. Other pages where the underlay is specified in the URL fetch that underlay individually.
* Remove redux since its only remaining use was to store the list of underlays. Remove some other bits that were only connected via redux such as the previous storage implementation.


ui/src/router.tsx
```diff
@@ -167,7 +172,8 @@ export function useBaseParams(): BaseParams {
 
 // TODO(tjennison): This is becoming spaghetti. Consider alternative ways to set
 // this up or perhaps alternative libraries.
-const prefix = "tanagra/underlays/:underlayName/studies/:studyId/";
+const underlayPrefix = "tanagra/underlays/:underlayName";
+const prefix = underlayPrefix + "/studies/:studyId/";
```
marikomedlock, line 176: nit: Can you use these in the paths up above?

marikomedlock (approved):

ui/src/router.tsx
tjennison-google: I just reverted that change since I wasn't actually using the split version. It's a bit of a mess since I need several different pieces in different places and I think naming the little fragments just ends up being harder to read.

PR #457, opened 2023-06-27
Render Tanagra inside an iframe

tjennison-google:
* Separate the outer sample app from core Tanagra. Some areas could still use further separation to better delineate the boundaries (e.g. routing). Setting `REACT_APP_ADDITIONAL_ROUTES=none` in a `.env` file will disable it.
* Use the postMessage API to navigate away from the iframe when returning to the study overview page instead of navigating.
* Treat the export page as an optional interstitial page in the routing to indicate whether exit actions should return there or close the iframe.
* Significant changes to integration tests to support the new iframe layout. Cypress also doesn't natively handle iframes.
* Remove the unused SD components.


tjennison-google:
image

ui/src/criteriaHolder.tsx
```diff
@@ -2,25 +2,26 @@ import ActionBar from "actionBar";
...
 import GridLayout from "layout/gridLayout";
 import { useState } from "react";
+import { useNavigate } from "react-router-dom";
 
 export type CriteriaHolderProps = {
   title: string;
   plugin: CriteriaPlugin<object>;
-  doneURL?: string;
   cohort?: boolean;
-  defaultBackURL?: string;
```
marikomedlock, line 11: Are `doneURL` and `defaultBackURL` configurable? Per deployment or per criteria plugin?

ui/src/router.tsx
```diff
@@ -107,18 +113,6 @@ function additionalRoutes() {
...
-    case "sd":
-      return [
-        {
-          path: "sdAdmin",
-          element: <SdAdmin />,
-        },
-        {
-          path: "studies",
-          element: <Studies />,
-        },
```
marikomedlock, line 120: This removes the SD admin pages from the routing, but the code for them still lives in this GH repo, right? Is the plan to move those into the VUMC repo at some point?

ui/src/sampleApp/README.md
```diff
@@ -0,0 +1,9 @@
+# Tanagra Sample App
```
marikomedlock, line 1: Not for this PR, but I think it would be helpful to centralize all our docs (e.g. either move this readme and the top-level ui one to the `docs` directory or add links to these pages in the main readme if you prefer not to move them).

marikomedlock (approved):

ui/src/criteriaHolder.tsx
tjennison-google: The `exitAction` is configurable per deployment via configuration in router.tsx. That is potentially insufficient for some deployments with multiple underlays but I expect all non-trivial deployments to embed Tanagra in an iframe, which case the exit action will always be sending a message to close the iframe.

PR #462, opened 2023-06-30
Support fine-grained access control.

marikomedlock:
Previously we supported only all-or-nothing permissions. Either a user was permitted to read/create/etc. an underlay/study/cohort/etc. or do nothing with it. Now we support fine-grained permissions. So a user could be able to e.g. read a study without being able to modify it.


service/src/main/java/bio/terra/tanagra/app/controller/StudiesV2ApiController.java
```diff
@@ -72,9 +78,12 @@ public ResponseEntity<ApiStudyListV2> listStudies(
       List<String> properties,
       Integer offset,
       Integer limit) {
-    ResourceIdCollection authorizedStudyIds =
-        accessControlService.listResourceIds(
-            SpringAuthentication.getCurrentUser(), STUDY, offset, limit);
+    ResourceCollection authorizedStudyIds =
+        accessControlService.listAuthorizedResources(
+            SpringAuthentication.getCurrentUser(),
+            Permissions.forActions(STUDY, READ),
```
tjennison-google, line 84: Something to consider for the future is that list should potentially be its own permission, at least for some entities. For example, an admin might have permissions to list/create/delete studies but not have access to the contents.

tjennison-google (approved):

service/src/main/java/bio/terra/tanagra/app/controller/StudiesV2ApiController.java
marikomedlock: Like a GCP role, that corresponds to a list of specific permissions? Yes, I could definitely do that as a future enhancement.

marikomedlock: Oh nevermind, I misread your comment. You mean adding a specific `LIST` permission, in addition to the existing `READ` permission. Yes, I could do that too. I was thinking of something similar like a  `READ_METADATA` permission specifically for admins.

PR #464, opened 2023-07-06
Temporarily allow all underlays in VUMC admin access control.

marikomedlock:
This is just a temporary workaround to allow access to all underlays when access control is configured to use the VUMC admin service.

@arychj Let me know if/when this is no longer needed and I can revert back to calling the API.


arychj:
@marikomedlock as long as the core _only_ requests READ for an underlay, this should already be handled in the admin service.

marikomedlock:
From discussion in Slack:
***DO NOT MERGE*** this PR. We think it will no longer be needed, but will hold off closing until VUMC SDD deploys the latest code and confirms everything is working as intended.

PR #477, opened 2023-07-14
Redirect outside of the iframe

tjennison-google:
* VWB rejects requests to show the login page inside an iframe. This is the better behavior anyways since you can otherwise go to VWB after export and be stuck in the iframe with no way to leave it.

### DIFF
--- a/ui/src/datasets.tsx
+++ b/ui/src/datasets.tsx
@@ -47,9 +47,13 @@ import React, {
 } from "react";
 import { useNavigate } from "react-router-dom";
 import {
+  absoluteExportURL,
   cohortURL,
   conceptSetURL,
   newConceptSetURL,
+  redirect,
+  RETURN_URL_PLACEHOLDER,
+  useBaseParams,
   useExitAction,
 } from "router";
 import { StudyName } from "studyName";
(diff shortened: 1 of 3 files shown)


tarekmamdouh:
> * VWB rejects requests to show the login page inside an iframe. This is the better behavior anyways since you can otherwise go to VWB after export and be stuck in the iframe with no way to leave it.

@tjennison-google  So this problem only happens during Export?

tjennison-google:
> > * VWB rejects requests to show the login page inside an iframe. This is the better behavior anyways since you can otherwise go to VWB after export and be stuck in the iframe with no way to leave it.
> 
> @tjennison-google So this problem only happens with Export?

Yes, when Tanagra redirects to the VWB import flow it's currently only redirecting the Tanagra iframe. The model is that Tanagra is now running in an iframe and the containing application (AoU, SDD, etc) will have headers and footers around it. The login page returns a CORS rejection preventing the page from being shown inside an iframe so this change makes the redirect happen outside of the iframe and redirects the entire page instead.

tarekmamdouh (approved):
lgtm

PR #484, opened 2023-07-25
Bump Google OAuth2 library version.

marikomedlock:
Fix dependency conflict that was causing problems decoding bearer token.


chenchals (approved):
LGTM! We both came to same conclusion!

PR #490, opened 2023-08-02
Remove reference to deleted sdd underlay

tjennison-google:


marikomedlock (approved):
Thanks for catching this!

PR #496, opened 2023-08-10
DT-467 setting the study id to an AoU workspaceNamespace. 

freemabd:
This will allow the AoU workbench to save a study with the correct id(workspaceNamespace in AoU land) when creating a workspace in AoU.


service/src/main/java/bio/terra/tanagra/app/controller/StudiesV2ApiController.java
```diff
@@ -44,6 +44,7 @@ public ResponseEntity<ApiStudyV2> createStudy(ApiStudyCreateInfoV2 body) {
         SpringAuthentication.getCurrentUser(), Permissions.forActions(STUDY, CREATE));
     Study.Builder studyToCreate =
         Study.builder()
+            .id(body.getId())
```
freemabd, line 47: If the id is not set in the request body it will set the id to null.

service/src/main/java/bio/terra/tanagra/service/artifact/Study.java
```diff
@@ -179,7 +180,8 @@ public Study build() {
       if (properties == null) {
         properties = new HashMap<>();
       }
-      if (id == null) {
+      //true if the id is empty or null
+      if (StringUtils.isEmpty(id)) {
```
freemabd, line 184: Added an extra check here for empty, just in case someone decides to pass an empty string as id. Not sure how you may want to handle duplicate ids/data integrity violations, but I'll leave that up to your team.

marikomedlock: Thanks for the added check.

I think the duplicate ids case should already be handled [on DB insert](https://github.com/DataBiosphere/tanagra/blob/454cc77099f60f88019435e6da807ae341f78d66/service/src/main/java/bio/terra/tanagra/db/StudyDao.java#L77).

service/src/test/java/bio/terra/tanagra/service/StudyServiceTest.java
```diff
@@ -82,6 +83,16 @@ void createUpdateDelete() throws InterruptedException {
...
+
+    // Create with id.
+    String id = "aou-rw-26297573";
+    Study studyWithId =
+        studyService.createStudy(
+            Study.builder().id(id).displayName(displayName).description(description),
+            createdByEmail);
+    assertNotNull(studyWithId);
+    LOGGER.info("Created study {} at {}", studyWithId.getId(), studyWithId.getCreated());
+    assertEquals(id, studyWithId.getId());
```
marikomedlock, line 95: RE duplicate ids comment above: Maybe a check here would be helpful to make sure we don't break this in the future? e.g. add below

```
    // Create with duplicate id.
    assertThrows(BadRequestException.class, () -> studyService.createStudy(Study.builder().id(id), createdByEmail));

```

marikomedlock (approved):
Thanks for handling this change!

PR #497, opened 2023-08-14
[DT-476] Reindex new synthetic cdr for AoU Test environment

chenchals:
- Created new synthetic CDRs that is `enabled for tanagra` in the AoU CDR config file for test and local environment.
- Moved / renamed all underlay configs from `SR2022Q4R6` to `SR2023Q3R1` and  from SC2022Q4R6` to `SC2023Q3R1` for Registered and Controlled Tiers respectively
- Ran indexing build with `projectId` and `queryProjectId` pointing to `all-of-us-ehr-dev` for both underlays
- Updated `projectId` and `queryProjectId` in RT and CT underlay json files to appropriate values for project inside secure perimeter
- **Note** Indexing builds had issues with running in concurrent mode, so used `SERIAL` for both indexing jobs.


marikomedlock (approved):
LGTM.

What issues did you run into when indexing in parallel mode? Happy to follow-up in Slack.

PR #500, opened 2023-08-14
[DT-476-1] Updated name and display name for underlays

chenchals:
- updated name and displayName properties for underlays

### DIFF
--- a/service/src/main/resources/config/aou/test/SC2023Q3R1/SC2023Q3R1.json
+++ b/service/src/main/resources/config/aou/test/SC2023Q3R1/SC2023Q3R1.json
@@ -1,6 +1,6 @@
 {
-  "name": "SC2022Q4R6",
-  "displayName": " Controlled Tier Vxx",
+  "name": "SC2023Q3R1",
+  "displayName": "Controlled Tier Synthetic Dataset v4 - Tanagra",
   "dataPointers": [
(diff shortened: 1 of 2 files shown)


service/src/main/resources/config/aou/test/SC2023Q3R1/SC2023Q3R1.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "SC2022Q4R6",
-  "displayName": " Controlled Tier Vxx",
+  "name": "SC2023Q3R1",
+  "displayName": "Controlled Tier Synthetic Dataset v4 - Tanagra",
```
marikomedlock, line 3: There's no `displayName` property. It's on my list to add, but haven't gotten there, yet.

If this isn't causing problems when you index, it's fine with me to leave it, but it won't be exposed to the UI.

marikomedlock (approved):

service/src/main/resources/config/aou/test/SC2023Q3R1/SC2023Q3R1.json
chenchals: I see that, since I only see `name` displayed for underlay and I had to fix that. but I will leave it there for-when you implement. Thanks

PR #502, opened 2023-08-24
Allow multiple occurrences in criteria occurrence entity group.

marikomedlock:
- Allowed each `CriteriaOccurrence` entity group to define >=1 occurrence entities. This is to support source hierarchies, which may map to >=1 domain (e.g. `criteria_occurrence` and `procedure_occurrence`).
- Defined the source hierarchies for the SDD underlay: PheWAS, ICD-9-CM, ICD-9-Proc, ICD-10-CM, ICD-10-PCS, CPT-4.
- Re-indexed underlays: `broad/cms_synpuf`, `broad/aou_synthetic`, `verily/cms_synpuf`, `verily/pilot_synthea_2022q3`, `verily/sdd_refresh0323`.

The changed files fall in 4 categories, in case it's useful to review them separately.
- `java` files with the `CriteriaOccurrence` entity group change.
- New config files for source hierarchies, in the verily+vumc sdd underlays.
- Updated config files for existing `CriteriaOccurrence` groups (property `occurrenceEntity` -> `occurrenceEntities`).
- Updated generated SQL files for tests with the new index dataset name (post re-indexing), and some renamed generated index tables.


service/src/main/resources/config/vumc/sdd_refresh0323/ui/top_level.json
```diff
@@ -170,6 +236,172 @@
...
+    {
+      "type": "classification",
+      "id": "tanagra-phewas",
+      "title": "PheWAS",
+      "conceptSet": true,
+      "category": "Source Codes",
+      "columns": [
+        { "key": "name", "width": "100%", "title": "Concept name" },
+        { "key": "code", "width": 120, "title": "Code" },
+        { "key": "label", "width": 120, "title": "Label" },
```
tjennison-google, line 248: What is the difference between `label`, `code`, and `name`? Do we need all three?

tjennison-google (approved):

service/src/main/resources/config/vumc/sdd_refresh0323/ui/top_level.json
marikomedlock: e.g. name='Surgery', code='0234', label='0234 - Surgery'

I think we need `name` and `code` to display in the list/hierarchy views. I'm not sure we need `label`, which is just a concatenation of `name` and `code`. I added `label` because that's what the AoU-CB shows in some places, and I wasn't sure how to do the concatenation in the UI code, so instead I pointed it at a new attribute.

tjennison-google: Seems to me we probably don't want a column for it then but we can look and adjust it later.

PR #503, opened 2023-08-28
[DT-473] Create authorization calls for Tanagra Aou integartion

chenchals:
- call AoU endpoint with studyId and bearer-token to check if user has access to workspace.
- Based on workspace access level of the user implement the 
  - isAuthorized and
  - listAllPermissions
  Methods of the _bio.terra.tanagra.service.accesscontrol.AccessControl_ interface
AoU end-point for the call: **/v1/workspaces/access/{workspaceNamespace=StudyId}**

### DIFF
--- a/service/src/main/java/bio/terra/tanagra/service/accesscontrol/impl/AouAccessControl.java
+++ b/service/src/main/java/bio/terra/tanagra/service/accesscontrol/impl/AouAccessControl.java
@@ -1,0 +1,182 @@
+package bio.terra.tanagra.service.accesscontrol.impl;
+
+import bio.terra.tanagra.exception.SystemException;
+import bio.terra.tanagra.service.accesscontrol.AccessControl;
+import bio.terra.tanagra.service.accesscontrol.Action;
+import bio.terra.tanagra.service.accesscontrol.Permissions;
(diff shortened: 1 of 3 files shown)


service/src/main/java/bio/terra/tanagra/service/accesscontrol/impl/AouAccessControl.java
```diff
@@ -0,0 +1,160 @@
...
+            resource.getStudy(),
+            user.getEmail(),
+            ex);
+        return false;
+      }
+    }
+  }
+
+  @Override
+  public ResourceCollection listAllPermissions(
```
marikomedlock, line 83: This method needs to return a list of the resources that the user can see, along with the permissions they have for each. e.g. For studies, a list of workspace/study id - workspace role/tanagra permissions.

marikomedlock (commented):
I think the `isAuthorized` and `listAllPermissions` methods are not doing quite the right thing. I will put up a branch with an outline of how it could work -- does that sound good? Then you can take a look and we can discuss together.

chenchals:
Closing this PR in lieu of [#506](https://github.com/DataBiosphere/tanagra/pull/506)

PR #506, opened 2023-08-29
AoU Workbench access control implementation class with placeholders.

marikomedlock:
- Added an implementation class of the `AccessControl` interface for the AoU Researcher Workbench. The class defines placeholder methods for actually calling the workbench API, this PR just includes logic for processing possible responses.
- Defined a mock implementation that just uses an in-memory map, instead of actually calling the workbench API, and wrote tests against it.


chenchals (approved):
Thank you! I think most of it is there now. Lets merge this and I will work onit.

PR #507, opened 2023-08-30
API definitions for activity log related endpoints.

marikomedlock:
**This PR does not include any implementation, just API definitions.**

- Added an optional `cohortRevisionId` query parameter to the `/v2/studies/{studyId}/cohorts/{cohortId}` endpoint. The parameter will only be used for `GET` requests. If populated, the returned `CohortV2` object will be for the specified revision. If not populated, the behavior will be the same as now, and the returned `CohortV2` object will be for the latest revision.
- Added a new `isDeleted` flag to the `CohortV2` object to support fetching deleted cohorts.
- Added 2 new endpoints to support auditing requirement for SDD prod:
  - `v2/activityLogEntries`, which lists all entries with optional filtering of returned log entries by user email, resource type, and activity type. This endpoint will be called by the VUMC admin UI.
  - `v2/activityLogEntries/{activityLogEntryId}` endpoint, which returns a single entry. This endpoint will be called by the core UI page that displays details about a single audit entry.


service/src/main/resources/api/service_openapi.yaml
```diff
@@ -974,6 +1004,30 @@ components:
         items:
           type: string
 
+    ActivityLogFilterUserEmail:
+      name: userEmail
+      in: query
+      description: Text search on the activity log entry user email
```
tjennison-google, line 1010: We might want this to be more exact that a text search, or at least to have such an option, otherwise it would be a pain for cases like usernames that are substrings of other usernames. I'm not sure what the best way to represent API-wise but I'm thinking of a version that matches specifically from the beginning of the username, so `def@` doesn't match `abcdef@verily.com`. A regex mode or exact match mode could be other possibilities.

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -1927,6 +1993,10 @@ components:
       items:
         $ref: "#/components/schemas/ConceptSetV2"
 
+    ConceptSetIdV2:
```
tjennison-google, line 1996: We're going to need to audit at the "data feature set" level rather than the individual "concept set" ("data feature") level. Individual concept sets won't be top level objects anymore once we add "data feature sets". Does it make sense to flesh out that API first?

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2386,6 +2456,75 @@ components:
...
+      type: string
+      enum: [ "STUDY", "COHORT", "REVIEW", "CONCEPT_SET" ]
+
+    Resource:
+      type: object
+      properties:
+        type:
+          $ref: "#/components/schemas/ResourceType"
+        studyId:
+          $ref: "#/components/schemas/StudyIdV2"
```
tjennison-google, line 2469: Perhaps these should reference objects for each type rather than ids. For now most would just contain their ids but then you wouldn't need the `additionalInfo` grab bag.

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2386,6 +2456,75 @@ components:
...
+          nullable: true
+        conceptSetId:
+          $ref: "#/components/schemas/ConceptSetIdV2"
+          nullable: true
+      required:
+        - studyId
+
+    ActivityType:
+      type: string
+      enum: [ "CREATE_STUDY", "DELETE_STUDY", "CREATE_COHORT", "DELETE_COHORT", "EXPORT_COHORT", "CREATE_REVIEW", "DELETE_REVIEW", "CREATE_CONCEPT_SET", "DELETE_CONCEPT_SET" ]
```
tjennison-google, line 2488: Thoughts on editing cohorts/"data features sets"? Is it enough to capture at review and export time?

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -1840,6 +1897,8 @@ components:
           description: Timestamp of when the cohort was last modified
           type: string
           format: date-time
+        isDeleted:
```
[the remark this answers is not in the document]
marikomedlock: I don't know if we can delete things -- it depends on whether VUMC needs to keep their audit logs for some amount of time, or even indefinitely. We can ask them in our next meeting. Regardless of what they say, I think we should plan any code changes for post-prod.

service/src/main/resources/api/service_openapi.yaml
marikomedlock: RE top level objects: I'm not sure I'll implement "data feature sets" as actual collections of "concept sets" on the backend. I may just update the definition of a "concept set". Not sure though, since I haven't thought through all the changes, yet.

RE defining data feature sets API first: Since that's not an SDD prod requirement, I want to do the auditing stuff first. So my inclination here is to keep using concept sets for now.

service/src/main/resources/api/service_openapi.yaml
marikomedlock: I'm not sure that would solve the `additionalInfo` bag of optional properties, since that depends on `ActionType`, not `ResourceType` (e.g. the backend would set `additionalInfo.exportType` for the `EXPORT_COHORT` `ActionType`, but not for `CREATE_COHORT`).

Definitely open to suggestions on how to better return information that is not universal for all action types. I considered making them not `required` properties of the top-level `ActivityLogEntry` object, but thought that made it harder to see what was common for all log entries. Open to opinions on that though.

service/src/main/resources/api/service_openapi.yaml
marikomedlock: I asked VUMC before and they said just cohorts at create/delete/review/export. So I guess I could reduce this list of events down to just those 4, but I thought it would be relatively straightforward to add the others once I have all the tables. If it turns into more work than I'm expecting, I'll skip the other ones.

RE cohort edit: This sounds useful to me, but potentially hard to implement because isn't the UI auto-saving every time someone adds a new criteria? Might make the audit log cluttered, unless we had some timer logic (e.g. freeze the cohort revision at most every 1 day).

service/src/main/resources/api/service_openapi.yaml
marikomedlock: Good idea. I'll add an exact match flag.

marikomedlock: done.

service/src/main/resources/api/service_openapi.yaml
marikomedlock: Ah I see. I think one object per `ActionType` is a lot of repetition of properties (e.g. all will have `studyId`). I'd prefer to just define them once.

I could also make `additionalInfo` a generic property map that action types can fill in as necessary. Upside is we don't need to change the API significantly to add a new action type (just a new `ActionType` enum value). Thoughts?

service/src/main/resources/api/service_openapi.yaml
marikomedlock: Per offline discussion -- we now think data features sets will be different enough from concept sets, that they'll probably warrant their own new schemas (i.e. the cohort definition schema isn't reusable enough for this case). So I removed the concept set activity types from this PR, given they're not required for SDD prod. I'll add in ones for data feature sets later post-prod.

service/src/main/resources/api/service_openapi.yaml
marikomedlock: Per offline discussion, we think the current structure of `additionalInfo` will best support the filtering use case in the VUMC admin UI. Leaving as is.

service/src/main/resources/api/service_openapi.yaml
marikomedlock: Per offline discussion, we agreed that supporting any additional audit events can come post-prod, without any major changes.

tjennison-google (approved):

PR #508, opened 2023-08-30
Added source/standard selection when adding criteria

tjennison-google:
Added source/standard tags to criteria as appropriate and rearranged criteria configs to fix overall criteria ordering.


marikomedlock (commented):
Changes lgtm. Can you make the same config changes to the `vumc` underlay too, just so they stay in sync?

tjennison-google:
> Changes lgtm. Can you make the same config changes to the `vumc` underlay too, just so they stay in sync?

Done. Thanks for the reminder.

marikomedlock (approved):

PR #513, opened 2023-09-05
[DT-473] Implement tanagra-AoU authorization call

chenchals:
- Weird for Underlay for `isAuthorized` and `listAllPermissions`
- Tests modified to reflect the weirdness for `AouWorkbenchAccessControlTest.underlay()` and `AouWorkbenchAccessControlTest.study()`
-

### DIFF
--- a/service/src/main/java/bio/terra/tanagra/service/accesscontrol/impl/AouWorkbenchAccessControl.java
+++ b/service/src/main/java/bio/terra/tanagra/service/accesscontrol/impl/AouWorkbenchAccessControl.java
@@ -6,6 +6,12 @@ import java.util.*;
 import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.client.RestTemplate;
 
 /**
  * AoU Researcher Workbench access control plugin implementation that checks permissions on the
@@ -33,10 +39,11 @@ public class AouWorkbenchAccessControl implements AccessControl {
           ResourceType.ANNOTATION_KEY.getActions()); // All annotation key actions.
 
   // Workspace writers have all permissions in a study, except for deleting the study.
+  // Check if a WRITER can update study (workspace)?
   private static final Map<ResourceType, Set<Action>> WORKSPACE_WRITER_PERMISSIONS =
       Map.of(
           ResourceType.STUDY,
-          Set.of(Action.READ, Action.UPDATE, Action.CREATE_COHORT, Action.CREATE_CONCEPT_SET),
+          Set.of(Action.READ, Action.CREATE_COHORT, Action.CREATE_CONCEPT_SET),
           ResourceType.COHORT,
           ResourceType.COHORT.getActions(), // All cohort actions.
           ResourceType.CONCEPT_SET,
@@ -46,7 +53,7 @@ public class AouWorkbenchAccessControl implements AccessControl {
           ResourceType.ANNOTATION_KEY,
           ResourceType.ANNOTATION_KEY.getActions()); // All annotation key actions.
 
-  // Workspace readers cannnot persist any changes to a study, only view the existing state.
+  // Workspace readers cannot persist any changes to a study, only view the existing state.
   private static final Map<ResourceType, Set<Action>> WORKSPACE_READER_PERMISSIONS =
       Map.of(
           ResourceType.STUDY, Set.of(Action.READ),
@@ -54,8 +61,8 @@ public class AouWorkbenchAccessControl implements AccessControl {
           ResourceType.CONCEPT_SET, Set.of(Action.READ),
           ResourceType.REVIEW, Set.of(Action.READ, Action.QUERY_INSTANCES, Action.QUERY_COUNTS),
           ResourceType.ANNOTATION_KEY, Set.of(Action.READ));
+
   private String basePath;
-  private String oauthClientId;
 
   @Override
   public String getDescription() {
@@ -65,21 +72,18 @@ public class AouWorkbenchAccessControl implements AccessControl {
   @Override
   public void initialize(List<String> params, String basePath, String oauthClientId) {
     // Store the basePath and oauthClientId first, so we can use it when calling the workbench API.
-    if (basePath == null || oauthClientId == null) {
-      throw new IllegalArgumentException(
-          "Base URL and OAuth client id are required for Workbench API calls");
+    // check: oauthClientId not required?
+    if (basePath == null) {
+      throw new IllegalArgumentException("Base URL is required for Workbench API calls");
     }
     this.basePath = basePath;
-    this.oauthClientId = oauthClientId;
   }
 
   @Override
   public boolean isAuthorized(UserId user, Permissions permissions, @Nullable ResourceId resource) {
-    if (permissions.getType().equals(ResourceType.ACTIVITY_LOG)) {
-      LOGGER.warn("AoU-RW will not use the activity log, so we should never get here.");
-      return false;
-    } else if (permissions.getType().equals(ResourceType.UNDERLAY)) {
+    if (permissions.getType().equals(ResourceType.UNDERLAY)) {
       // Browsing the cohort builder without saving. Always allow.
+      // for AoU underlays? -> browsing is not allowed. Need underlay specific access (RT and CT)
       return true;
     } else if (Permissions.forActions(ResourceType.STUDY, Action.CREATE).equals(permissions)) {
       // Workbench creates a Tanagra study as part of workspace creation. Always allow.
@@ -88,12 +92,13 @@ public class AouWorkbenchAccessControl implements AccessControl {
       // Check for some permissions on a particular artifact (e.g. study, cohort).
       if (resource == null || resource.getStudy() == null) {
         LOGGER.error(
-            "Study id is required to check authorization of a particular artifact: {}", resource);
+            "AoU Workspace Access Level api Study id is required to check authorization for resource {}",
+            resource);
         return false;
       }
 
       // Call the workbench to get the user's role on the workspace that contains this study.
-      // Expect the Tanagra study id to match the Workbench workspace id.
+      // Expect the Tanagra study id to match the Workbench workspaceNamespace.
       WorkspaceRole role = apiGetWorkspaceAccess(user, resource.getStudy());
       if (role == null) {
         return false;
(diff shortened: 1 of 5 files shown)


service/src/test/java/bio/terra/tanagra/service/accesscontrol/AouWorkbenchAccessControlTest.java
```diff
@@ -39,31 +41,14 @@ protected void cleanup() {
...
     // service.list
-    assertServiceListWithReadPermission(
-        USER_1, ResourceType.UNDERLAY, null, true, cmsSynpufId, aouSyntheticId, sddId);
-    assertServiceListWithReadPermission(
-        USER_2, ResourceType.UNDERLAY, null, true, cmsSynpufId, aouSyntheticId, sddId);
-    assertServiceListWithReadPermission(
-        USER_3, ResourceType.UNDERLAY, null, true, cmsSynpufId, aouSyntheticId, sddId);
-    assertServiceListWithReadPermission(
-        USER_4, ResourceType.UNDERLAY, null, true, cmsSynpufId, aouSyntheticId, sddId);
+    assertResourceTypeNoListPermissions(USER_1, ResourceType.UNDERLAY);
```
marikomedlock, line 49: Is it possible to reuse an existing method here instead? e.g. `assertServiceListWithReadPermission(USER_1, ResourceType.UNDERLAY, /*parent*/ null, /*isAllResources*/ false)`

service/src/test/java/bio/terra/tanagra/service/accesscontrol/AouWorkbenchAccessControlTest.java
```diff
@@ -39,31 +41,14 @@ protected void cleanup() {
...
-    assertHasPermissions(USER_2, aouSyntheticId);
-    assertHasPermissions(USER_2, sddId);
-    assertHasPermissions(USER_3, cmsSynpufId);
-    assertHasPermissions(USER_3, aouSyntheticId);
-    assertHasPermissions(USER_3, sddId);
-    assertHasPermissions(USER_4, cmsSynpufId);
-    assertHasPermissions(USER_4, aouSyntheticId);
-    assertHasPermissions(USER_4, sddId);
-
+    assertUnderlayPermissions(USER_1, aouSyntheticId);
```
marikomedlock, line 45: Is it possible to reuse an existing method here instead? e.g. `assertDoesNotHavePermissions(USER_1, aouSyntheticId)`

service/src/main/java/bio/terra/tanagra/service/accesscontrol/impl/AouWorkbenchAccessControl.java
```diff
@@ -65,18 +72,18 @@ public String getDescription() {
...
     // Store the basePath and oauthClientId first, so we can use it when calling the workbench API.
-    if (basePath == null || oauthClientId == null) {
-      throw new IllegalArgumentException(
-          "Base URL and OAuth client id are required for Workbench API calls");
+    // check: oauthClientId not required?
+    if (basePath == null) {
+      throw new IllegalArgumentException("Base URL is required for Workbench API calls");
     }
     this.basePath = basePath;
-    this.oauthClientId = oauthClientId;
```
marikomedlock, line 73: Could you also please update the `ACCESS_CONTROL.md` doc to remove the oauth client id for the AoU-RW section?

service/src/main/java/bio/terra/tanagra/service/accesscontrol/impl/AouWorkbenchAccessControl.java
```diff
@@ -169,22 +160,42 @@ public Map<ResourceType, Set<Action>> getTanagraPermissions() {
...
-  /** @return a map of the workspace id -> role that the user has on each. */
-  protected Map<String, WorkspaceRole> apiListWorkspacesWithAccess(UserId userId) {
-    LOGGER.debug(
-        "Calling workbench API to list workspaces user {} can see, and their role on each.",
-        userId.getEmail());
-    LOGGER.debug("Workbench API base path: {}, oauth client id: {}", basePath, oauthClientId);
-    // TODO: Call the workbench API to list the workspaces the user has access to, and their role on
-    // each.
-    return Map.of(); // Temporarily, no one has access to any worksapces.
+    String url = basePath + "/" + workspaceId;
```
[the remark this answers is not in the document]
chenchals: good catch! Thought I added it to basePath to restrict any other calls. Thanks

service/src/main/java/bio/terra/tanagra/service/accesscontrol/impl/AouWorkbenchAccessControl.java
```diff
@@ -65,18 +72,18 @@ public String getDescription() {
...
     }
     this.basePath = basePath;
-    this.oauthClientId = oauthClientId;
   }
 
   @Override
   public boolean isAuthorized(UserId user, Permissions permissions, @Nullable ResourceId resource) {
     if (permissions.getType().equals(ResourceType.UNDERLAY)) {
       // Browsing the cohort builder without saving. Always allow.
+      // for AoU underlays? -> browsing is not allowed. Need underlay specific access (RT and CT)
```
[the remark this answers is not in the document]
chenchals: got it

service/src/main/java/bio/terra/tanagra/service/accesscontrol/impl/AouWorkbenchAccessControl.java
```diff
@@ -169,22 +160,42 @@ public Map<ResourceType, Set<Action>> getTanagraPermissions() {
    */
   protected @Nullable WorkspaceRole apiGetWorkspaceAccess(UserId userId, String workspaceId) {
```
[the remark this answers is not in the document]
chenchals: good point

service/src/test/java/bio/terra/tanagra/service/accesscontrol/AouWorkbenchAccessControlTest.java
chenchals: yes that worked for underlays

service/src/test/java/bio/terra/tanagra/service/accesscontrol/AouWorkbenchAccessControlTest.java
chenchals: No it does not work since we return `true` for `isAuthorized` call. The `assertHasPermissions(USER_1,aouSyntheticId)` does not work because the assertion methond assumes that if isAuthorized is true then listAllPermissions must also asert to true see [assertHasPermissions](https://github.com/DataBiosphere/tanagra/blob/90ed11eb51a6413b512367a0c6a3af8c27ca7801/service/src/test/java/bio/terra/tanagra/service/accesscontrol/BaseAccessControlTest.java#L198-L200).  This is the weirdness for underlay

service/src/main/java/bio/terra/tanagra/service/accesscontrol/impl/AouWorkbenchAccessControl.java
chenchals: done

service/src/test/java/bio/terra/tanagra/service/accesscontrol/AouWorkbenchAccessControlTest.java
```diff
@@ -270,4 +254,28 @@ void annotationKey() {
...
+
+  // Specific cases for AoU
+  private void assertUnderlayPermissions(UserId user, ResourceId resource) {
+    assertTrue(impl.isAuthorized(user, Permissions.empty(resource.getType()), resource));
+    assertServiceListWithReadPermission(USER_1, ResourceType.UNDERLAY, null, false);
+  }
+
+  private void assertStudyPermissions(UserId user, ResourceId resource, Action... actions) {
+    assertTrue(impl.isAuthorized(user, Permissions.empty(resource.getType()), resource));
+    assertStudyNoListPermissions(user, resource.getType(), actions);
```
marikomedlock, line 266: Similar to underlays above, is it possible to reuse an existing method here instead? e.g. `assertServiceListWithReadPermission(USER_1, ResourceType.STUDY, /*parent*/ null, /*isAllResources*/ false)`

service/src/test/java/bio/terra/tanagra/service/accesscontrol/AouWorkbenchAccessControlTest.java
marikomedlock: Ah I see -- `isAuthorized` always returns true for a particular underlay, but `listAuthorizedResources` always returns empty when asked to list the underlays. That is confusing. Could you add a comment here to that effect please, so the next person doesn't get tripped up on the same thing?

Ditto for studies, since I think that will be similar (`isAuthorized` sometimes returns true for a particular study, but `listAuthorizedResources` always returns empty when asked to list the studies).

service/src/main/java/bio/terra/tanagra/service/accesscontrol/impl/AouWorkbenchAccessControl.java
```diff
@@ -109,31 +115,10 @@ public boolean isAuthorized(UserId user, Permissions permissions, @Nullable Reso
...
-                WorkspaceRole role = entry.getValue();
-                studyPermissionsMap.put(
-                    ResourceId.forStudy(studyId),
-                    Permissions.forActions(
-                        ResourceType.STUDY, role.getTanagraPermissions().get(ResourceType.STUDY)));
-              });
-      return studyPermissionsMap.isEmpty()
-          ? ResourceCollection.empty(type, parentResource)
-          : ResourceCollection.resourcesDifferentPermissions(studyPermissionsMap);
+    if (ResourceType.UNDERLAY.equals(type)) {
```
marikomedlock, line 118: Per comment below, should this be `UNDERLAY || STUDY`?

service/src/main/java/bio/terra/tanagra/service/accesscontrol/impl/AouWorkbenchAccessControl.java
```diff
@@ -64,22 +70,21 @@ public String getDescription() {
...
   public boolean isAuthorized(UserId user, Permissions permissions, @Nullable ResourceId resource) {
-    if (permissions.getType().equals(ResourceType.ACTIVITY_LOG)) {
-      LOGGER.warn("AoU-RW will not use the activity log, so we should never get here.");
-      return false;
-    } else if (permissions.getType().equals(ResourceType.UNDERLAY)) {
+    if (permissions.getType().equals(ResourceType.UNDERLAY)) {
       // Browsing the cohort builder without saving. Always allow.
+      // For AoU underlays - browsing is not allowed.
+      // If this has to return true consider using "/v1/profile" endpoint
+      // to get user profile for checking RT/CT access
```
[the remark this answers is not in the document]
chenchals: got it. Still at a later date need to chck if user can access CT / RT

service/src/test/java/bio/terra/tanagra/service/accesscontrol/AouWorkbenchAccessControlTest.java
chenchals: Cant use because - Listing for studies is controlled on AoU RW side. so except for list actions other actions are allowed depending on user's workspace access level

service/src/main/java/bio/terra/tanagra/service/accesscontrol/impl/AouWorkbenchAccessControl.java
chenchals: Just UNDERLAY, because  the if logic in else will take care of STUDY, since parentResource for STUDY would be null for listAllPermissions call
```
     if (parentResource == null || parentResource.getStudy() == null) {
        LOGGER.error(
            "Study id is required to list permissions for a child artifact: {}", parentResource);
        return ResourceCollection.empty(type, parentResource);
      }

```

service/src/test/java/bio/terra/tanagra/service/accesscontrol/AouWorkbenchAccessControlTest.java
marikomedlock: I see, OK.

service/src/main/java/bio/terra/tanagra/service/accesscontrol/impl/AouWorkbenchAccessControl.java
marikomedlock: Ah OK -- in that case, can you update the log message there please? i.e. we should never get here ... as the underlays one above.

chenchals: I think it is better to use 
```
if (ResourceType.UNDERLAY.equals(type) || ResourceType.STUDY.equals(type)) {
      // AoU will not call to list Underlays or list studies - this is managed by Workbench
      LOGGER.error("Calls from AoU Workbench should never get here. ");
      return ResourceCollection.empty(type, null); 
```
where it is explici that AoU should never call list underlays or list studies.

marikomedlock (approved):
Thanks for getting this working!

PR #514, opened 2023-09-05
Activity log implementation.

marikomedlock:


service/src/main/java/bio/terra/tanagra/db/ActivityLogDao.java
```diff
@@ -0,0 +1,240 @@
...
+        && activityTypeFilter == null
+        && resourceTypeFilter == null) {
+      return "";
+    }
+
+    // Filter on the userEmail, activityType, and/or resourceType.
+    List<String> whereConditions = new ArrayList<>();
+    if (userEmailFilter != null && !userEmailFilter.isEmpty()) {
+      if (userEmailExactMatch) {
+        whereConditions.add("user_email = :user_email_filter");
```
tjennison-google, line 155: Does this need to be sanitized?

service/src/main/java/bio/terra/tanagra/db/ActivityLogDao.java
```diff
@@ -0,0 +1,240 @@
...
+          "INSERT INTO activity_log_resource (activity_log_id, stable_index, resource_type, study_id, cohort_id, cohort_revision_id, review_id) "
+              + "VALUES (:activity_log_id, :stable_index, :resource_type, :study_id, :cohort_id, :cohort_revision_id, :review_id)";
+      LOGGER.debug("CREATE activity log resource: {}", sql);
+      List<MapSqlParameterSource> resourceParamSets = new ArrayList<>();
+      for (int i = 0; i < activityLog.getResources().size(); i++) {
+        ActivityLogResource resource = activityLog.getResources().get(i);
+        resourceParamSets.add(
+            new MapSqlParameterSource()
+                .addValue("activity_log_id", activityLog.getId())
+                .addValue("stable_index", i)
```
tjennison-google, line 99: Is a stable index actually needed here? Won't the default ordering be by date/time anyway?

service/src/main/java/bio/terra/tanagra/db/ActivityLogDao.java
marikomedlock: No, the `MapSqlParameterSource` (see nearby line `params.addValue`) handles escaping all parameters before executing the prepared statement.

service/src/main/java/bio/terra/tanagra/db/ActivityLogDao.java
marikomedlock: This column is for ordering the list of resources for each activity log entry. You're right that the default ordering for the log entries themselves will always be by the `logged` timestamp, so no need for a stable index on that table.

PR #517, opened 2023-09-06
API updates for the VUMC audit UI.

marikomedlock:
- API updates to support the VUMC audit UI. Discussed in Sept 6 meeting w/product and eng.
  - Add an optional `includeDeleted` flag to the `listStudies` endpoint. The audit UI will set this to true to list all studies that may have associated audit logs.
  - Include the study key-value metadata map in the activity log entry. The audit UI will use this to display the associated IRB number.
  - Add a `primaryEntityCount` to the activity log entry. For the `CREATE_REVIEW` and `EXPORT_COHORT` activity types only, this will be populated with the number of persons in the cohort(s). 
- This PR does not include implementation, just the API definitions. I will do the implementation separately in follow-on PRs.


tjennison-google (approved):

dolbeew (approved):
Thanks

PR #519, opened 2023-09-06
Implement listStudies includeDeleted flag.

marikomedlock:
- Implemented the `listStudies` endpoint new `includeDeleted` flag.
- Added an `isDeleted` flag to the `StudyV2` API object, so the admin UI could display un/deleted studies differently, if desired.


service/src/main/resources/api/service_openapi.yaml
```diff
@@ -1840,6 +1840,9 @@ components:
         lastModified:
           type: string
           format: date-time
+        isDeleted:
+          type: boolean
+          nullable: false
```
marikomedlock, line 1845: One additional API change to support listing deleted studies -- return this flag for each study returned. @tjennison-google @dolbeew

tjennison-google (approved):

PR #521, opened 2023-09-07
Accommodate letters in ICD9-CM codes.

marikomedlock:
- Allowed for letters in the ICD9-CM codes child/parent relationships query. Previously, only numeric codes were allowed.
- Re-indexed just the ICD9-CM entity and group in the existing `verily/sdd` index dataset.
- This fix is in response to a bug report from VUMC while testing out the source hierarchies.

### DIFF
--- a/service/src/main/resources/config/verily/sdd_refresh0323/sql/icd9cm_parentChild.sql
+++ b/service/src/main/resources/config/verily/sdd_refresh0323/sql/icd9cm_parentChild.sql
@@ -23,8 +23,8 @@ SELECT
 FROM `verily-tanagra-dev.aou_static_prep_useast1.prep_concept` p
 LEFT JOIN `victr-tanagra-test.sd_20230328.concept` AS c
     ON c.vocabulary_id = p.vocabulary_id
-    AND UPPER(c.concept_code) >= UPPER(REGEXP_EXTRACT(p.concept_code, r'^([0-9\.]+)-'))
-    AND UPPER(c.concept_code) <= UPPER(REGEXP_EXTRACT(p.concept_code, r'-([0-9\.]+)$'))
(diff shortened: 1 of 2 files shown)


service/src/main/resources/config/verily/sdd_refresh0323/sql/icd9cm_parentChild.sql
```diff
@@ -23,8 +23,8 @@ SELECT
 FROM `verily-tanagra-dev.aou_static_prep_useast1.prep_concept` p
 LEFT JOIN `victr-tanagra-test.sd_20230328.concept` AS c
     ON c.vocabulary_id = p.vocabulary_id
-    AND UPPER(c.concept_code) >= UPPER(REGEXP_EXTRACT(p.concept_code, r'^([0-9\.]+)-'))
-    AND UPPER(c.concept_code) <= UPPER(REGEXP_EXTRACT(p.concept_code, r'-([0-9\.]+)$'))
+    AND UPPER(c.concept_code) >= UPPER(REGEXP_EXTRACT(p.concept_code, r'^([a-zA-Z0-9\.]+)-'))
```
tjennison-google, line 26: Will this also apply to ICD10?

tjennison-google (approved):

service/src/main/resources/config/verily/sdd_refresh0323/sql/icd9cm_parentChild.sql
marikomedlock: No, this change only applies to ICD9-CM. I think the other source hierarchy parent/child relationships are not affected by this bug though:
- CPT4 doesn't parse the code at all.
- ICD9-Proc does parse the code, but it doesn't include any codes with letters, only numbers.
- ICD10-CM does parse the code, but for some reason I remembered to include letters in that one. 🤷🏼‍♀️ 
- ICD10-PCS only looks at the code length to figure out parent/child relationships.

PR #525, opened 2023-09-12
Fix boolean backfill value changelog.

marikomedlock:
- Added a new changelog that uses the `valueBoolean` tag instead of `value`. This tells Liquibase to use the dbms-specific boolean literal.
- Kept the previous changelog, since it has already been deployed in the Verily test env. Tagged it as only applying to postgres, so that AoU deployments won't try to apply it.
- Created a new changelog specifically for MariaDB/MySQL that uses the correct tag. Tested locally that this works with MariaDB when there are existing rows in the `study` table (db upgrades are not handled by our tests - they always start with a clean db).
- Since the AoU deployment had a bad changelog, we'll need to do a partial manual rollback: `alter table study drop column is_deleted;`. Then this new changelog will be applied correctly.
- FYI: We are planning a DB wipe before SDD prod launch, at which time I plan to consolidate all these changelogs.


freemabd (approved):
LGTM. Thanks

PR #532, opened 2023-09-14
[DT-329] Underlays for fitbit for aou

chenchals:
- updated indexing builds for
  - RT SR2023Q3R1
  - CT SC2023Q3R1
Using integer instead of boolean falg for `EXISTS` on t_person table as unable to figure out how to default to boolean-true when for example `Fitbit Activity Summary` is chosen as cohort criteria. With interger, I get a 0-1 range slider for criteria (non exists for boolean).
Below are the functional UI after change. Shown only for CT, RT is similar

_RT Fitbit criteria selection and choosing range slider to be 1-1_
RT-fitbit-selections

_RT Fitbit prepackaged Dataset page_
https://github.com/DataBiosphere/tanagra/assets/4452480/37ef9071-fa9a-440e-90f8-fbafcbe9e49f


ui/src/datasets.tsx
```diff
@@ -126,11 +126,9 @@ export function Datasets() {
     conceptSets: { id: string; name: string }[]
   ) => {
     // TODO(tjennison): Only show demographics until we have better prepackaged
+    // Updated for AoU to show all prepackaged configured in the top_level.json
```
tjennison-google, line 129: We don't want to show the Analgesics prepackaged concept set everywhere so if you want to remove the splice here you'll need to remove Analgesics from any of the other underlays it appears in.

Also, just remove this entire comment now that it's not applicable anymore.

ui/src/datasets.tsx
```diff
@@ -126,11 +126,9 @@ export function Datasets() {
...
     // concept sets.
     const cs = conceptSets;
-    if (!editable) {
-      cs.splice(1, Infinity);
-    }
 
     return cs.map((conceptSet, i) => (
```
tjennison-google, line 133: Remove the previous line and just use `conceptSets.map...` now that the extra variable is no longer needed.

tjennison-google (changes requested):

ui/src/datasets.tsx
chenchals: done

ui/src/datasets.tsx
chenchals: done

tjennison-google (approved):

service/local-dev/run_server.sh
```diff
@@ -53,8 +53,8 @@ elif [[ ${useAouUnderlays} ]]; then
   export TANAGRA_EXPORT_SHARED_GCS_BUCKET_PROJECT_ID=broad-tanagra-dev
   export TANAGRA_EXPORT_SHARED_GCS_BUCKET_NAMES=broad-tanagra-dev-bq-export
   # specify access-control
-  export TANAGRA_ACCESS_CONTROL_BASE_PATH=https://all-of-us-workbench-test.appspot.com
-  export TANAGRA_ACCESS_CONTROL_MODEL=AOU_WORKBENCH
+  export TANAGRA_ACCESS_CONTROL_BASE_PATH=https://api-dot-all-of-us-workbench-test.appspot.com
+  # export TANAGRA_ACCESS_CONTROL_MODEL=AOU_WORKBENCH
```
marikomedlock, line 57: nit: The base path won't be used unless you specify an access control model that uses it (`OPEN_ACCESS` is the default and does not use it). Can you either remove both of these lines, or uncomment both?

service/src/main/resources/config/aou/test/SC2023Q3R1/sql/t_person.sql
```diff
@@ -1,4 +1,41 @@
...
+           WHEN sds.person_id IS NULL THEN 0 ELSE 1 END has_fitbit_sleep_daily_summary,
+       CASE
+           WHEN sl.person_id IS NULL THEN 0 ELSE 1 END has_fitbit_sleep_level,
+       CASE
+           WHEN sl.person_id IS NULL AND sl.person_id IS NULL AND hrs.person_id IS NULL
+            AND si.person_id IS NULL AND sds.person_id IS NULL THEN 0 ELSE 1 END has_fitbit
+FROM `all-of-us-ehr-dev.SC2023Q3R1.person` p
+         LEFT JOIN
+     (SELECT DISTINCT person_id
+      FROM `all-of-us-ehr-dev.SC2023Q3R1.activity_summary`) asum ON (p.person_id = asum.person_id)
```
marikomedlock, line 26: Not for this PR, but I hope we can remove these flags in the future. Will follow-up offline.

service/local-dev/run_server.sh
chenchals: commented both lines and added a comment to uncomment for testing :-)

marikomedlock (approved):

service/src/main/resources/config/aou/test/SC2023Q3R1/sql/t_person.sql
chenchals: That would be good.  More importantly, need to support direct values (boolean is missing, String is missing)/ some SQL like `distinct(column_name) from t_person` for attributeMapping in t_person
```
      "has_fitbit_xxx": {
        "value": { "column": "has_bitbit_xx" },
        "display": { ["true","false"])
or
        "display": {"rawSql??": "SELECT DISTINCT has_fitbit_xx FROM t_person""}
 
```

PR #535, opened 2023-09-18
Add clustering. Skip display fields in count query.

marikomedlock:
This PR contains several performance-related improvements. As part of this PR, I also re-indexed the verily/sdd underlay.

**Clustering**
- Added a new `frequentFilterAttributes` list property to the entity config. Defined it as `['gender','race','age']` for `person` entity. Order matters because I use this to cluster the entity table. For the primary entity, this should be the attributes we expect to use for breakdowns.
- Cluster some index tables.
  - For entity-level display hints table, cluster on the attribute name, enum value columns.
  - For entity table, cluster on the `frequentFilterAttributes` columns defined in the config, or on the text search string column, if text search is enabled.
  - For ancestor-descendant table, cluster on the ancestor column.
  - For parent-child table, cluster on the parent column.

**Count Query**
- Cache entity-level display hints.
- Don't group by display fields in count queries. Only group by the values (e.g. `gender`, not `t_display_gender`) in the SQL query. Then lookup the display fields in the entity-level display hints cache, and populate them manually before returning to the caller.

**Text Search**
- Updated the exact text match to use `REGEXP_CONTAINS` instead of `CONTAINS_SUBSTR`, based on [this list](https://cloud.google.com/bigquery/docs/bi-engine-optimized-sql#strings_and_bytes) of BigQuery optimized built-in string functions.

**General Code Cleanup (not performance-related)**
- Skip writing the relationship id pairs index table for relationships that are a direct foreign-key mapping.
- Moved the indexing jobs definition into a separate class `JobSequencer`. Now the `Indexer` class is just the command line entrypoint.


service/src/main/resources/config/verily/sdd_refresh0323/entity/person.json
```diff
@@ -14,6 +14,7 @@
     { "type": "SIMPLE", "name": "biovu_sample_is_nonshippable", "dataType": "BOOLEAN" },
     { "type": "SIMPLE", "name": "biovu_sample_has_plasma", "dataType": "BOOLEAN" }
   ],
+  "frequentFilterAttributes": [ "gender", "race", "age" ],
```
tjennison-google, line 17: Nothing to do here now but we should clean this up to avoid duplicating what's in the UI config for breakdowns.

tjennison-google (approved):

service/src/main/resources/config/verily/sdd_refresh0323/entity/person.json
marikomedlock: Agreed.

PR #540, opened 2023-09-19
Add basic support for upgrading criteria

tjennison-google:
A future PR will allow plugins to impact the process.


marikomedlock (approved):

marikomedlock:
Discussed offline. We haven't planned out how to handle UI schema upgrades yet, but in the short-term it will have to happen in the UI.

PR #554, opened 2023-09-28
Add support for renaming annotations

tjennison-google:
* Refactor the text input dialog to allow for variable parameters (i.e. the initial value).

### DIFF
--- a/ui/src/sampleApp/studiesList.tsx
+++ b/ui/src/sampleApp/studiesList.tsx
@@ -72,12 +72,7 @@ export function StudiesList() {
     });
   };
 
-  const [newStudyDialog, showNewStudyDialog] = useTextInputDialog({
-    title: "New study",
-    textLabel: "Study name",
-    buttonLabel: "Create",
-    onConfirm: onCreateNewStudy,
-  });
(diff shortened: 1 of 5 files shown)


ui/src/sampleApp/studiesList.tsx
```diff
@@ -151,7 +146,14 @@ export function StudiesList() {
             />
           )}
           <Button
-            onClick={showNewStudyDialog}
+            onClick={() =>
+              showNewStudyDialog({
+                title: "New study",
+                textLabel: "Study name",
+                buttonLabel: "Create",
+                onConfirm: onCreateNewStudy,
```
marikomedlock, line 154: Not for this PR, but would it be possible to open a new study automatically once you create it, instead of bringing you back to the study list page?

marikomedlock (approved):

ui/src/sampleApp/studiesList.tsx
tjennison-google: Yes. I haven't paid much attention to these pages because they won't be part of SD or AoU but that would be a nice, simple improvement.

PR #579, opened 2023-10-10
Add dataflow subnetwork name config property.

marikomedlock:
This property is required for VPC networks set to "custom subnet mode". It is not required for "auto subnet mode".


chenchals (approved):
LGTM! So I just have to give the name right?

PR #580, opened 2023-10-11
Replace sleep with poll for table existence in hints indexing job.

marikomedlock:
Fixed a `todo` in the `ComputeEntityLevelHints` indexing job. Previously, it was just sleeping for 5 sec between creating a table and inserting into it. Replaced this sleep with polling for table existence, every 10 sec for up to 10 min.

### DIFF
--- a/indexer/src/main/java/bio/terra/tanagra/indexing/job/ComputeEntityLevelDisplayHints.java
+++ b/indexer/src/main/java/bio/terra/tanagra/indexing/job/ComputeEntityLevelDisplayHints.java
@@ -1,7 +1,6 @@
 package bio.terra.tanagra.indexing.job;
 
 import bio.terra.tanagra.api.schema.EntityLevelDisplayHints;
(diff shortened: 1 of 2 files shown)


chenchals (approved):
LGTM! Thank you

indexer/src/main/java/bio/terra/tanagra/indexing/job/ComputeEntityLevelDisplayHints.java
```diff
@@ -150,6 +140,9 @@ public void run(boolean isDryRun) {
               }
             });
 
+    // Poll for table existence before we try to insert the hint rows. Maximum 60 x 10 sec = 10 min.
+    pollForTableExistenceOrThrow(getAuxiliaryTable(), 60, Duration.ofSeconds(10));
```
tjennison-google, line 144: Given that we previously expected the table to appear with 5s, waiting 10m seems a bit excessive and just makes failures slower. A minute or two would be plenty.

PR #583, opened 2023-10-16
[DT-320] tanagra aou rt/ct - update underlays for CPT4 source codes

chenchals:
- updated RT/CT with CPT4 pattern for source codes from sdd_refresh0323
- When creating a conceptSet / dataFeature for cpt4 -> The output of export has *all 4* domains that were configured for cpt4 in top_level.json, even though the concept specified in the Data Feature was only in procedure domain. The tabs for other domains are empty.
  - t_procedure_occurrence has data
image

- t_ingredient_occurrence and t_observation_occurrence and t_measurement_domain does not have data - how to restrict to domains that have data? 

  - t_ingredient_occurrence does *not* have data
image


marikomedlock (approved):
LGTM. I think the issue with empty tabs showing up is a TODO to fix in the UI, so not something you misconfigured, but @tjennison-google to confirm.

tjennison-google:
> LGTM. I think the issue with empty tabs showing up is a TODO to fix in the UI, so not something you misconfigured, but @tjennison-google to confirm.

It's not misconfigured but I'm also not convinced this is actually incorrect. The challenge is that there's no way to differentiate between a CPT4 selection that theoretically only matches certain domains and one where there just happens to be no data. Not showing expected tables just because nobody happened to match them feels wrong to me so until we can understand that it doesn't make sense to hide the empty tables.

PR #587, opened 2023-10-17
Add support for feature sets

tjennison-google:
* Add main feature sets UI and link from study overview.
* Fake backend using localstorage until API is ready.
* Add helpers for popovers.
* Refactor functionality from datasets to share with feature sets page.
* Refactor undo/redo toolbar.
* Add arbitrary icon support to checkboxes.
* Add column background styling and header suffixes to treegrid.

TODO:
* Preview using a specified cohort.
* Export page.
* Backend integration.

### DIFF
--- a/ui/src/cohort.ts
+++ b/ui/src/cohort.ts
@@ -1,4 +1,8 @@
-import { ROLLUP_COUNT_ATTRIBUTE, SortDirection } from "data/configuration";
+import {
+  findEntity,
+  ROLLUP_COUNT_ATTRIBUTE,
+  SortDirection,
+} from "data/configuration";
 import {
   Filter,
   FilterType,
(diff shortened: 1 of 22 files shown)


ui/src/cohort.ts
```diff
@@ -213,6 +217,64 @@ export function upgradeCriteria(
...
+  const addFilter = (occurrence: string, filter?: Filter | null) => {
+    if (!occurrences.has(occurrence)) {
+      occurrences.set(occurrence, []);
+    }
+    if (filter) {
+      occurrences.get(occurrence)?.push(filter);
+    }
+  };
+
+  predefinedCriteria
```
marikomedlock, line 249: Does predefined here refer to prepackaged data features? Or what's the difference between predefinedCriteria and userCriteria?

ui/src/hooks.ts
```diff
@@ -145,12 +150,45 @@ export function useConceptSet() {
...
+    throw new PathError(`No valid feature set in current context.`);
+  }
+  return featureSet;
+}
+
+export function useFeatureSet() {
+  return useOptionalFeatureSet(true) as NonNullable<FeatureSet>;
+}
+
+function useOptionalFeatureSetAndCriteria(throwOnUnknown: boolean) {
```
marikomedlock, line 165: Is this for listing the possible data features in the dataset builder page?

marikomedlock (commented):
LGTM. Some questions inline to make sure I'm following how it works.

ui/src/cohort.ts
tjennison-google: Yes, prepackaged. I'm just moving towards clearer naming: predefined vs. user defined.

ui/src/hooks.ts
tjennison-google: No, this is referencing the criteria and feature set it belongs to currently selected via the URL, for use while editing the criteria. The catch is that context values can always potentially be null, since there's no way for the compiler to validate how you compose your components. So this function returns those values optionally in case something needs to check if it's valid, but there's a wrapper that throws so the criteria editing page doesn't have no worry about null values when they're expected to be valid based on routing.

marikomedlock (approved):

PR #596, opened 2023-11-03
Call new isAuthorizedUser endpoint in VUMC admin service plugin.

marikomedlock:
For the VUMC access control plugin, call the new `isAuthorizedUser` admin service endpoint to check if a user has permission to create a study. All other access control checks remain unchanged.


dolbeew (approved):
Thanks!

PR #605, opened 2023-11-14
Check url for token if local env var set

dolbeew:
Add check for a new UI env var `REACT_APP_GET_LOCAL_AUTH_TOKEN` which should only be set for local dev environments. If the env var is set, check the url for the `token` query param and set it as the access token. This is needed for enabling and testing authz for local dev.

This is the same temp solution we've used in the AoU test env, but it will be removed there once we get both apps on the same domain. I have not been able to find a local solution for deploying multiple apps on the same domain/port, so we'll need this workaround for local dev. We are also open to other solutions if you know of any.


tjennison-google (approved):

dolbeew:
I did see nginx as an option but as you said, it would be a significant change for the whole workbench. I may revisit that later though.

PR #606, opened 2023-11-14
[DT-290] config loinc and non-hierarchy

freemabd:
This PR adds the config for the following:

- LOINC measurement hierarchy
- HCPCS non-hierarchy 
- LOINC clinical non-hierarchy


chenchals (approved):
LGTM.

PR #613, opened 2023-11-20
[DT-623-1] Update AoU underlay config after overhaul

chenchals:
- Setting `isComputeDisplayHint": true` for `person.has_*` to enable range selector in UI
- Updated to move entity/entity groups from omop to aouRT (common entity and entity groups)
- updated underlay json files to use `aouRT/` for all `omop/`
- updated to use TANAGRA_UNDERLAY_FILES=aouSR2023Q3R1_local,aouSC2023Q3R1_local for local deploy and test
- testing with local deploy using: `tanagra$ ./service/local-dev/run_server.sh -amt` command-line


underlay/src/main/resources/config/datamapping/aouRT/entity/procedure/entity.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "procedure",
```
freemabd, line 2: Can we name this `procedureSnomed` since it represents the SNOMED hierarchy

underlay/src/main/resources/config/datamapping/aouRT/entitygroup/conditionPerson/entityGroup.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "conditionPerson",
```
freemabd, line 2: Name it `conditionSnomedPerson` since it's the relationship between SNOMED conditions and person

underlay/src/main/resources/config/datamapping/aouRT/entitygroup/procedurePerson/entityGroup.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "procedurePerson",
```
freemabd, line 2: Can we name `procedureSnomedPerson` since it represents the SNOMED hierarchy relationship with person

underlay/src/main/resources/config/service/aouSR2023Q3R1_local.json
```diff
@@ -0,0 +1,27 @@
...
+    },
+    "indexData": {
+      "projectId": "all-of-us-ehr-dev",
+      "datasetId": "SR2023Q3R1",
+      "tablePrefix": "T"
+    },
+    "queryProjectId": "all-of-us-ehr-dev",
+    "dataLocation": "us-central1"
+  },
+  "dataflow": {
```
marikomedlock, line 20: Is this meant to be an `indexer` config file, instead of a `service` one? Service does not need dataflow config properties.

freemabd (commented):
Requested a few name changes.

marikomedlock (commented):
Most of the entity and entity group definitions here look like identical copies of the ones in `omop` (e.g. `observation`, `visitPerson`) -- is that your intent?

chenchals:
> Most of the entity and entity group definitions here look like identical copies of the ones in `omop` (e.g. `observation`, `visitPerson`) -- is that your intent?

yes, except in cases where you have already overridden the entity/entitygroup

freemabd (approved):

PR #615, opened 2023-11-21
[No-Ticket] rename aou service json files

chenchals:
- renamed service json files and moved them under `service/aou` folder
- verified by deploying with `project.rb command` in AoU


freemabd (approved):
Thanks

PR #626, opened 2023-11-30
Add endpoint for showing cohort revisions

tjennison-google:
* This required splitting the APIs that worked within the context of an underlay from the ones that didn't, since the activity log works at a study level.


ui/src/router.tsx
```diff
@@ -134,19 +152,25 @@ function additionalRoutes() {
...
-              element: <StudiesList />,
-            },
-            {
-              path: "underlays/:underlayName/studies/:studyId",
-              element: <StudyOverview />,
-            },
-            {
-              path: "underlays/:underlayName/studies/:studyId/*",
-              element: <TanagraContainer />,
+              element: <UnderlaySourceContextRoot />,
```
marikomedlock, line 159: I'm confused why the underlay root is a child of the study root -- isn't the path underlay/study?

ui/src/cohortReview/newReviewDialog.tsx
```diff
@@ -47,15 +47,17 @@ export type NewReviewDialogProps = {
...
   const countState = useSWRImmutable(
     {
       type: "count",
       cohort: props.cohort,
     },
     async () => {
-      return await source.filterCount(generateCohortFilter(props.cohort));
+      return await underlaySource.filterCount(
+        generateCohortFilter(props.cohort)
+      );
```
marikomedlock, line 60: I see several of these formatting changes. Did you run some command to do this automatically? If so, is there a way to make that happen for every change? For the Java code, there's a Gradle plugin and the GHAs run a check (also via the plugin, not my code) that fails if it hasn't been formatted. Maybe we can do something similar for the UI code? (Not for this PR, just noticed it here)

marikomedlock (commented):
This is the dialog for the audit log entry, correct? Is there a way to see it from from the default UI (i.e. the one in Verily test/dev envs)?

tjennison-google:
> This is the dialog for the audit log entry, correct? Is there a way to see it from from the default UI (i.e. the one in Verily test/dev envs)?

Yes. Unfortunately not but it's the exact same display as the cohort summary on the Export page. At some point it would be worth building a basic sample auditing page that could link to it.

ui/src/cohortReview/newReviewDialog.tsx
marikomedlock: Sounds good. The generated docs + ts file is another pattern -- runs a command then does a `git diff` and fails if anything changed. You could add it there, too.

marikomedlock (approved):

PR #642, opened 2023-12-13
[DT-621] Add non-hierarchy items for procedure and Drug domains

chenchals:
- Procedure:
  - search for “colorectal cancer”  (HCPCS) 
image
  - search for "XR Chest 2 Views" (LOINC) there are no LOINC non-hierarchy in test
- Drug:
  - search for “unclassified” (HCPCS)
image
  - search for “zoster vaccine” (CVX)
image


freemabd (approved):
THanks

PR #643, opened 2023-12-14
[DT-647] Fine tune condition domain

chenchals:
- update for non-hierarchy
- update for clustering when indexing
- update view for 
  - entity table
  - hierarchy-view
 
- contains `item-count` and initial sort on `rollup-count`
image

- hierarchy view sorted on `name`
image

- added non-hierarchy conditions
image

**Note** Concepts with _item-count/rollup-count_ that are _zero_ are not yet filtered/removed from indexing build


freemabd (approved):
Looks good.

PR #645, opened 2023-12-21
[DT-649] Fine tune observation domain

chenchals:
- update for clustering when indexing
- update view for
  - entity table
  - added selection for categorical values

1. contains `item-count` and sorted on item-count. The counts match with AoU (Note: there are also concepts from surveys in tanagra, In AoU we filter out concept_Ids in prep_survey during indexing build). Removed `Roll-up count`
image

1. Selection of concepts with categorical values will show the optional `list of categorical values (unsorted)` to choose from
image


freemabd (approved):
Thanks

PR #650, opened 2024-01-02
Add check for auth token in parent window local storage

dolbeew:
Now that we have the Workbench and Tanagra UIs on the same domain in the AoU test env, we can pass the bearer token through `localStorage`. We will still need to keep the less secure workaround of passing in the iframe url for deploying and testing locally.

### DIFF
--- a/ui/src/apiContext.ts
+++ b/ui/src/apiContext.ts
@@ -319,25 +319,14 @@ class FakeExportAPI {}
 
 class FakeUsersAPI {}
 
-const getAccessToken = () => {
-  let accessToken = "";
-  if (process.env.REACT_APP_GET_LOCAL_AUTH_TOKEN) {
-    // For local dev only, get the bearer token from the iframe url param
-    accessToken =
-      new URLSearchParams(window.location.href.split("?")[1]).get("token") ??
-      "";
-  } else {
-    // Check parent window's localStorage for auth token
-    // Use try/catch block to prevent UI from breaking in case of CORS error
-    try {
-      accessToken =
-        window.parent.localStorage.getItem("tanagraAccessToken") ?? "";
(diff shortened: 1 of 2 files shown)


ui/src/apiContext.ts
```diff
@@ -319,6 +319,26 @@ class FakeExportAPI {}
 
 class FakeUsersAPI {}
 
+const getAccessToken = () => {
```
tjennison-google, line 322: Use "function getAccessToken" for consistency

ui/src/apiContext.ts
```diff
@@ -319,6 +319,26 @@ class FakeExportAPI {}
...
+  let accessToken = "";
+  if (process.env.REACT_APP_GET_LOCAL_AUTH_TOKEN) {
+    // For local dev only, get the bearer token from the iframe url param
+    accessToken =
+      new URLSearchParams(window.location.href.split("?")[1]).get("token") ??
+      "";
+  } else {
+    // Check parent window's localStorage for auth token
+    // Use try/catch block to prevent UI from breaking in case of CORS error
```
tjennison-google, line 331: What is the purpose of this? Isn't the UI already broken if the auth token isn't available?

tjennison-google (approved):

ui/src/apiContext.ts
dolbeew: I added this when we were still getting the domains sorted in AoU test to prevent a blank page on CORS error. Since the Workbench and SD Tanagra UIs will now be on the same domains as the parent window, should be fine to remove it.

ui/src/apiContext.ts
dolbeew: Done

PR #655, opened 2024-01-04
Overhaul UI config

tjennison-google:
This change is not backwards compatible and breaks existing criteria.

* Remove occurrences from the UI config. Change the config to refer directly to entities and entity groups instead.
* Use backend config to replace classifications and primary entity config.
* Update the classification plugin to use the entity group configuration for related occurrences rather than manually specifying them.

### DIFF
--- a/ui/src/cohortReview/reviewHooks.ts
+++ b/ui/src/cohortReview/reviewHooks.ts
@@ -21,10 +21,11 @@ export type ReviewParams = {
 };
 
 export function useReviewParams(): ReviewParams {
+  const underlaySource = useUnderlaySource();
   const underlay = useUnderlay();
   const params = useBaseParams();
 
-  const primaryKey = underlay.uiConfiguration.dataConfig.primaryEntity.key;
+  const primaryKey = underlaySource.primaryEntity().idAttribute;
   const uiConfig = underlay.uiConfiguration.cohortReviewConfig;
   const participantIdAttribute = uiConfig.participantIdAttribute ?? primaryKey;
 
--- a/ui/src/criteria/classification.tsx
+++ b/ui/src/criteria/classification.tsx
@@ -27,7 +27,7 @@ import {
 } from "data/configuration";
 import { Filter, FilterType, makeArrayFilter } from "data/filter";
 import { MergedItem, mergeLists } from "data/mergeLists";
-import { ClassificationNode, UnderlaySource } from "data/source";
+import { EntityGroupData, EntityNode, UnderlaySource } from "data/source";
 import { DataEntry, DataKey } from "data/types";
 import { useUnderlaySource } from "data/underlaySourceContext";
 import { useIsNewCriteria, useUpdateCriteria } from "hooks";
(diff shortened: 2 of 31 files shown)


ui/src/cohortReview/reviewHooks.ts
```diff
@@ -21,10 +21,11 @@ export type ReviewParams = {
 };
 
 export function useReviewParams(): ReviewParams {
+  const underlaySource = useUnderlaySource();
   const underlay = useUnderlay();
   const params = useBaseParams();
 
-  const primaryKey = underlay.uiConfiguration.dataConfig.primaryEntity.key;
+  const primaryKey = underlaySource.lookupEntity("").idAttribute;
```
marikomedlock, line 28: nit: Consider an overloaded function e.g. `underlaySource.primaryEntity()` instead of the special empty string arg.

ui/src/criteria/classification.tsx
```diff
@@ -709,12 +767,49 @@ async function search(
...
+    throw new Error("At least one entity group must be configured.");
   }
-  return value[0];
+  return entityGroups;
 }
+
+const DEFAULT_SORT_ORDER = {
+  attribute: ROLLUP_COUNT_ATTRIBUTE,
+  direction: SortDirection.Desc,
+};
```
marikomedlock, line 815: Not for this PR, but it would be nice to have these defaults defined in the config schema so they show up in documentation. Ditto for the `DEFAULT_LIMIT` above.

ui/src/data/source.tsx
```diff
@@ -161,30 +155,27 @@ export type FeatureSet = {
   output: FeatureSetOutput[];
 };
 
+export type EntityGroupData = {
+  id: string;
+  entityId: string;
+  relatedEntityId?: string;
+  relatedEntityGroupId?: string;
+  occurrenceEntityIds: string[];
```
marikomedlock, line 163: Are these entity/group ids the same as the names in the config file?

ui/src/data/source.tsx
```diff
@@ -349,171 +340,121 @@ export class BackendUnderlaySource implements UnderlaySource {
...
-  }
-
-  lookupClassification(
-    occurrenceID: string,
-    classificationID: string
-  ): Classification {
-    return findByID(
-      classificationID,
-      this.lookupOccurrence(occurrenceID).classifications
+  lookupEntityGroup(entityGroupId: string): EntityGroupData {
```
marikomedlock, line 346: Would it be useful to add an endpoint for getting an entity group, similar to the existing `getEntity`? I think there's similar logic on the backend for lookup up a group from an entity(s).

marikomedlock (approved):

ui/src/cohortReview/reviewHooks.ts
tjennison-google: Done.

ui/src/criteria/classification.tsx
tjennison-google: As long as they would be guaranteed to be defined there. These exist as a fallback for when no value is specified.

ui/src/data/source.tsx
tjennison-google: Yes

PR #658, opened 2024-01-08
Fix sorting bug and misnamed sort fields

tjennison-google:
* Change hierarchy sort order fields using "defaultSort" to "sortOrder".
* Hierarchy sort orders were sometimes being applied to the list view.


chenchals (approved):
Thanks for getting to this!

PR #659, opened 2024-01-08
DT-697-Fine-tuning-Fixed sorting for list view and hierarchy view

chenchals:
- fixed sorting for search results view: No sort order specified -> sorts by rollup count
- fixed sorting for hierarchy view: specifed Specified `name` or `code` by comparing results in AoU


freemabd (approved):
Thanks

PR #661, opened 2024-01-09
Codebase overview readme.

marikomedlock:


docs/CODEBASE_OVERVIEW.md
```diff
@@ -0,0 +1,287 @@
...
+### Data export
+**`service` Gradle sub-project `export` package**
+
+Data export models are configurable per deployment. The various data export implementations, and the interface they
+share live in this package. More information in [this readme](./DATA_EXPORT.md).
+
+## UI
+**`ui` top-level directory**
+
+Web UI code. More information in [this readme](../ui/README.md).
```
tjennison-google, line 199: Would it be better to have a `UI.md` here instead keep all of the docs in the same place?

tjennison-google (approved):

docs/CODEBASE_OVERVIEW.md
marikomedlock: OK with me. I'll move the current `ui/README.md` here then.

marikomedlock: Done. Also added a note there about opening Cypress, which I keep forgetting.

PR #671, opened 2024-01-17
Add all AttributeFilter operators to OpenAPI spec.

marikomedlock:


service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2474,6 +2480,15 @@ components:
     PageMarker:
       type: string
 
+    UnaryOperator:
```
tjennison-google, line 2483: Seems unnecessary to have the operators types (Unary, Nary, Binary) be separate enums when they're all used in the same place as three adjacent variables. Why not just have a single Operator type? Or alternatively split AttributeFilter by operator type.

marikomedlock: Binary operator is used in other places, so didn't want to pollute the set of possible values there. Split into 3 different AttributeFilters works for me.

marikomedlock: OK change of heart after looking at the code. We already combine operators in the boolean logic filter, so I opted to do the same for the attribute filter to keep things consistent.

tjennison-google (approved):

PR #673, opened 2024-01-18
Add reference to the config properties docs in the deployment overview.

marikomedlock:
Also renamed the generated `DEPLOYMENT_CONFIG.md` docs to `APPLICATION_CONFIG.md` to better distinguish between the indexer+service deployment requirements and the service application configuration.


chenchals (approved):
thanks for the update of config props

PR #676, opened 2024-01-19
Add multi-select support to entityGroup criteria

tjennison-google:
* Show other items included by selected items. This is incomplete pending future work to expose all ancestors via the API.
* Fix snack bar showing old group names.


marikomedlock (approved):
Does the multi-select only apply to the hierarchy view, or can you do it in the list view as well?

tjennison-google:
> Does the multi-select only apply to the hierarchy view, or can you do it in the list view as well?

It also applies to the list view, though not the cross-domain search.

PR #679, opened 2024-01-19
DT-705 fixing cohort id on ApiReview

freemabd:
fixing cohort id on ApiReview


freemabd:
This fixed the routing problem that we were seeing. I looked replacing the cohort with cohort revision in the response but the UI is dependent on the cohort currently. I feel like it's probably better to coordinate that change with the UI experts.

marikomedlock (approved):

PR #680, opened 2024-01-20
Adding config for 20230831 index for SD Cohort Builder.

billbridgesvumc:
Adding config for 20230831 index for SD Cohort Builder.

### DIFF
--- a/underlay/src/main/resources/config/indexer/sd20230831_vumc.json
+++ b/underlay/src/main/resources/config/indexer/sd20230831_vumc.json
@@ -1,0 +1,25 @@
+{
+  "underlay": "sd20230831",
+  "bigQuery": {
+    "sourceData": {
+      "projectId": "sd-vumc-tanagra-test",
+      "datasetId": "sd_20230831",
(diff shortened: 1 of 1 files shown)


underlay/src/main/resources/config/indexer/sd20230831_vumc.json
```diff
@@ -0,0 +1,25 @@
+{
+  "underlay": "sd20230831",
```
marikomedlock, line 2: This underlay name needs to match a file in the `config/underlay` directory. You can either:
- Use the existing `sd020230331` underlay. This means that the data mapping is exactly the same (e.g. no table name or column data type changes). This is how I test ran the indexing on the refresh dataset. If we use the same underlay for multiple SD datasets, perhaps it would be better to change the underlay name to just `sd_version1` or something without the exact date in the name.
- Create a new `sd20230831` underlay. This would let us change the data mapping, even if only slightly. We're going to have to do this eventually to update the race field name `race_concept_id` -> `race_source_concept_id`, but doesn't have to be part of this PR.

PR #685, opened 2024-01-24
Fix list parsing in CLI option.

marikomedlock:
Fix list parsing, change from comma-separated option to a space-separated positional parameter.

### DIFF
--- a/docs/generated/indexer-cli/tanagra-clean-entity.adoc
+++ b/docs/generated/indexer-cli/tanagra-clean-entity.adoc
@@ -21,7 +21,8 @@ tanagra-clean-entity - Clean the outputs of all jobs for a single entity, or all
 
 *tanagra clean entity* [*--all*] [*--dry-run*] [*--github-dir*=_<githubDir>_]
                      *--indexer-config*=_<name>_ [*--job-executor*=_<jobExecutor>_]
-                     [*--output-dir*=_<outputDir>_] [*--names*=_<names>_]...
+                     [*--output-dir*=_<outputDir>_] [*--names*=_<names>_[,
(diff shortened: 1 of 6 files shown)


chenchals (changes requested):
A simple adding `split` to the options command seems to fix the issue 
For `EntityNames`
```
public class EntityNames {
  @CommandLine.Option(
      names = "--names",
      split = ",",
      description = "Entity name(s). Comma-separated list if more than one.")
  public List<String> names;
```
Ditto for `EntityGroupNames`
```
public class EntityGroupNames {
  @CommandLine.Option(
      names = "--names",
      split = ",",
      description = "Entity group name(s). Comma-separated list if more than one.")
  public List<String> names;
```

marikomedlock:
> A simple adding `split` to the option command seems to fix the issue For `EntityNames`
> 
> ```
> public class EntityNames {
>   @CommandLine.Option(
>       names = "--names",
>       split = ",",
>       description = "Entity name(s). Comma-separated list if more than one.")
>   public List<String> names;
> ```
> 
> Ditto for `EntityGroupNames`
> 
> ```
> public class EntityGroupNames {
>   @CommandLine.Option(
>       names = "--names",
>       split = ",",
>       description = "Entity group name(s). Comma-separated list if more than one.")
>   public List<String> names;
> ```

Oh that's much better, thanks. I'll give it a try

chenchals (approved):
thanks! what are .adoc files?

marikomedlock:
> thanks! what are .adoc files?

They're like markdown, just a different format for writing documentation. These adoc files are generated by the CLI library we're using for the indexer.

PR #688, opened 2024-01-25
Add PrimaryWithCriteria, OccurrenceForPrimary filters to the OpenAPI spec.

marikomedlock:
The filter implementations + tests were part of earlier PRs. This just exposes the filters to the UI, and in the future, criteria plugins.

### DIFF
--- a/service/src/main/resources/api/service_openapi.yaml
+++ b/service/src/main/resources/api/service_openapi.yaml
@@ -1429,7 +1429,7 @@ components:
         filterType:
           type: string
           enum:
-            ["ATTRIBUTE", "TEXT", "HIERARCHY", "RELATIONSHIP", "BOOLEAN_LOGIC"]
+            ["ATTRIBUTE", "TEXT", "HIERARCHY", "RELATIONSHIP", "BOOLEAN_LOGIC", "PRIMARY_WITH_CRITERIA"]
         filterUnion:
           type: object
           properties:
@@ -1443,6 +1443,8 @@ components:
               $ref: "#/components/schemas/RelationshipFilter"
             booleanLogicFilter:
               $ref: "#/components/schemas/BooleanLogicFilter"
+            primaryWithCriteriaFilter:
+              $ref: "#/components/schemas/PrimaryWithCriteriaFilter"
 
     AttributeFilter:
       type: object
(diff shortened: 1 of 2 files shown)


service/src/main/resources/api/service_openapi.yaml
```diff
@@ -1537,6 +1541,53 @@ components:
...
+        If the group_by_count_operator is set, then the group_by_count_value must also be set.
+        The group_by_count_attributes are optional. If undefined, we group by the id only.
+        When group_by fields are set, we filter on # of related occurrence instances.
+        e.g. filter person instances with >2 related condition_occurrence instances
+      properties:
+        entityGroup:
+          type: string
+        criteriaSubfilter:
+          $ref: "#/components/schemas/Filter"
+        occurrenceSubfiltersAndGroupByAttributes:
```
tjennison-google, line 1573: What is the purpose of this?

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -1537,6 +1541,53 @@ components:
...
+        primarySubfilter:
+          $ref: "#/components/schemas/Filter"
+
+    PrimaryWithCriteriaFilter:
+      type: object
+      required: [entityGroup]
+      description: |
+        Filter the primary entity with a criteria-occurrence entity group (e.g. people where 
+        condition=diabetes and visit_type=inpatient).
+        If the group_by_count_operator is set, then the group_by_count_value must also be set.
```
tjennison-google, line 1564: You should probably use the camelCase names (e.g. `groupByCountOperator`) to match the names in the file.

service/src/main/resources/api/service_openapi.yaml
marikomedlock: It's for modifiers (e.g. visit type, age at occurrence) and group by attributes (e.g. start_date, date).

tjennison-google: Why is in a separate object though? I guess maybe the keys would be for each occurrence entity?

marikomedlock: Right, this allows for different sub-filters and group by attributes per occurrence entity. I'll add a description to make it clear the keys are the occurrence entity names.

tjennison-google (approved):

PR #693, opened 2024-01-26
Prevent duplicate filters in criteria

tjennison-google:
EntityGroup criteria with multiple entity groups that contain the same occurrences would generate the same filter multiple times.

### DIFF
--- a/ui/src/criteria/classification.tsx
+++ b/ui/src/criteria/classification.tsx
@@ -205,11 +205,15 @@ class _ implements CriteriaPlugin<Data> {
   }
 
   filterEntityIds(underlaySource: UnderlaySource) {
-    return configEntityGroups(this.config)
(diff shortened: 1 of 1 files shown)


marikomedlock (approved):
Does the condition criteria plugin contain multiple entity groups? I thought it would just be conditionPerson. Or do you mean that conditionOccurrence is contained in multiple entity groups?

freemabd (approved):
Thanks for fixing this.

marikomedlock:
> Does the condition criteria plugin contain multiple entity groups? I thought it would just be conditionPerson. Or do you mean that conditionOccurrence is contained in multiple entity groups?

Nevermind, just saw your Slack message. Now I understand it's the former -- the AoU underlay specifies multiple entity groups.

PR #708, opened 2024-02-02
Fetch multiple pages of occurrence data in cohort review

tjennison-google:
This now displays all occurrence data in cohort review and fixes an intermittent UI integration test failure.

### DIFF
--- a/ui/src/cohortReview/cohortReview.tsx
+++ b/ui/src/cohortReview/cohortReview.tsx
@@ -144,7 +144,7 @@ export function CohortReview() {
               values: [instance?.data?.[primaryKey]],
             },
             null,
-            10000
+            0
           );
         })
       );
--- a/ui/src/data/source.tsx
+++ b/ui/src/data/source.tsx
@@ -479,20 +479,40 @@ export class BackendUnderlaySource implements UnderlaySource {
       entity.idAttribute
     );
 
-    const res = await parseAPIError(
-      this.underlaysApi.listInstances({
-        entityName: entity.name,
-        underlayName: this.underlay.name,
-        query: this.makeQuery(ra, entityId, cohort, conceptSet, limit),
-      })
-    );
+    let pageMarker: string | undefined;
+    let sql: string | undefined;
+    const data: DataEntry[] = [];
+    while (true) {
+      const res = await parseAPIError(
+        this.underlaysApi.listInstances({
+          entityName: entity.name,
+          underlayName: this.underlay.name,
+          query: this.makeQuery(
+            ra,
+            entityId,
+            cohort,
+            conceptSet,
+            limit,
+            pageMarker
+          ),
+        })
+      );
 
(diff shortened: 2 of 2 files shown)


ui/src/cohortReview/cohortReview.tsx
```diff
@@ -144,7 +144,7 @@ export function CohortReview() {
               values: [instance?.data?.[primaryKey]],
             },
             null,
-            10000
```
marikomedlock, line 147: Was this the previous limit?

ui/src/data/source.tsx
```diff
@@ -479,20 +479,40 @@ export class BackendUnderlaySource implements UnderlaySource {
...
-        underlayName: this.underlay.name,
-        query: this.makeQuery(ra, entityId, cohort, conceptSet, limit),
-      })
-    );
+    let pageMarker: string | undefined;
+    let sql: string | undefined;
+    const data: DataEntry[] = [];
+    while (true) {
+      const res = await parseAPIError(
+        this.underlaysApi.listInstances({
```
marikomedlock, line 487: Does this fetch all the rows, all pages? If so, do you think we should use a limit here instead, so the page doesn't time out if there are e.g. lots of notes? I think you could implement that with a very large page size and a single API call.

ui/src/cohortReview/cohortReview.tsx
tjennison-google: Yes

ui/src/data/source.tsx
tjennison-google: Yes. Doesn't fetching in pages make it less likely to timeout, since each individual fetch is smaller? Is there no backend limit on the page size?

marikomedlock: Does the page display while fetching or wait for all fetches to finish first? I think you're right about fetching in pages being less likely for the API calls to timeout. I was thinking that the display would time out for fetching all the rows, regardless of whether you spread it out across multiple API calls or not.

RE backend limit on page size: There's nothing in Tanagra, and I think the BQ limit is in total size of the data (20MB across multiple rows, or 100MB in a single row), not number of rows per se.

tjennison-google: The timeout is only an issue on individual network requests. The spinner can be displayed for as long as necessary.

I was assuming there would be a page size limit but if not I could do one large fetch instead. It's a little questionable though since I'm surprised 10000 wasn't enough already if there isn't a lower limit we're hitting somewhere.

marikomedlock (approved):

ui/src/data/source.tsx
tjennison-google: Sorry, the approval updated automatically but not your new comment. It will wait until all of the calls return before rendering but it won't fail because the timeout is for each request individually. I should page anyways since the number is arbitrary but we can tweak the page size as needed.

PR #717, opened 2024-02-09
Include PR title in release description.

marikomedlock:
- Set flag to automatically include PR title and link in the description of the GH release that is auto-generated when we merge a PR.
- Put the command I use locally to list the PRs between two release flags into a GHA so we can run it from the repo page.

### DIFF
--- a/.github/tools/diffReleases.sh
+++ b/.github/tools/diffReleases.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+firstRelease=$1
+lastRelease=$2
+
+if [ -z $firstRelease ] || [ -z $lastRelease ];
+then
+  echo "First and last release tags are required (e.g. 0.0.370).";
+  exit 1;
+fi
+
+git log $1..$2 --grep='^bump' --invert-grep --pretty=format:"%h %as %an%x09%s" > diffReleases.txt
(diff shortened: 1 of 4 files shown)


marikomedlock:
> I don't understand how this is meant to be used. Is there somewhere in the GH UI that can run this with parameters?

Yeah you can manually kick off a workflow from the Actions tab. Unfortunately, I can't test that this one works there until at least one version of the GHA yaml file is merged to `main`. But here's an example of one with parameters from another repo.
Screenshot 2024-02-09 at 11 11 00 AM

I don't know if this is going to be useful to other people, but the hope is that it lowers the barrier to diffing releases a little. Release notes don't really help us in that regard I don't think because we generate a release every PR.

tjennison-google:
> > I don't understand how this is meant to be used. Is there somewhere in the GH UI that can run this with parameters?
> 
> Yeah you can manually kick off a workflow from the Actions tab. Unfortunately, I can't test that this one works there until at least one version of the GHA yaml file is merged to `main`. But here's an example of one with parameters from another repo. Screenshot 2024-02-09 at 11 11 00 AM
> 
> I don't know if this is going to be useful to other people, but the hope is that it lowers the barrier to diffing releases a little. Release notes don't really help us in that regard I don't think because we generate a release every PR.

I see it now. It's after selecting an individual workflow. It's maybe not perfect but it's a step in the right direction for sure.

PR #719, opened 2024-02-09
Fix input indentation and gh pull depth in diff releases workflow.

marikomedlock:
We should now be able to run this on demand from the GH UI:
Screenshot 2024-02-09 at 12 35 29 PM

Download the output file:
Screenshot 2024-02-09 at 12 38 53 PM

And open it in a text editor on your local machine:
Screenshot 2024-02-09 at 12 39 52 PM

------

I also updated the test actions to skip when the merger is `broadbot`. This is to skip automated tests on the `bump` merges. We already run all tests on PR merge, so no need to run them twice for each PR (once for the PR merge and once for the bump merge). Hopefully this will reduce competition for the GHA runners so less things get queued for a long time.

### DIFF
--- a/.github/tools/diffReleases.sh
+++ b/.github/tools/diffReleases.sh
@@ -9,8 +9,10 @@ then
   echo "First and last release tags are required (e.g. 0.0.370).";
   exit 1;
 fi
+echo "First release tag: $first_release"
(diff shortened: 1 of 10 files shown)


tjennison-google (approved):
Good idea to skip tests on version bumps.

PR #722, opened 2024-02-13
Use protos to store plugin selection data

tjennison-google:
* Add protos that roughly mirror existing selection data. Decode selection data as protos if they fails to parse as JSON to enable backwards compatability.
* Store encoded protobufs in plugins and convert on demand using existing data structures. This minimizes the code changes as part of this conversion. Future PRs might change this behavior but protos don't work naturally with React state so some kind of workaround will be needed.

### DIFF
--- a/ui/codegen.sh
+++ b/ui/codegen.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+# Generate API OpenAPI code.
+rm -rf src/tanagra-api
+openapi-generator-cli generate -i ../service/src/main/resources/api/service_openapi.yaml -g typescript-fetch -t ./openapi-templates --additional-properties=typescriptThreePlus=true -o src/tanagra-api
+
+# Generate UI OpenAPI code.
+rm -rf src/tanagra-ui
+openapi-generator-cli generate -i ./ui_openapi.yaml -g typescript-fetch -t ./openapi-templates --additional-properties=typescriptThreePlus=true --model-name-prefix UI -o src/tanagra-ui
+
+# Generate plugin proto code.
+rm -rf src/proto
+mkdir src/proto
+npx protoc --proto_path=../underlay/src/main/proto/ --plugin=./node_modules/.bin/protoc-gen-ts_proto --ts_proto_out=./src/proto/ --ts_proto_opt=esModuleInterop=true --ts_proto_opt=outputClientImpl=false ../underlay/src/main/proto/criteriaselector/dataschema/*.proto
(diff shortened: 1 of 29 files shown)


ui/codegen.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
marikomedlock, line 1: This is much easier to read than the long string in the `package.json`. Great improvement.

Not for this PR, but I wonder if we should add the underlay config typescript generation here too.
`./gradlew annotationProcessor:run -Pgenerator=UNDERLAY_CONFIG_TYPESCRIPT`
Currently, we're generating the ts files and checking them into GH, but seems like it would be better to generate at build and gitignore them.

ui/src/cohort.ts
```diff
@@ -337,7 +337,7 @@ export function createCriteria(
 export function getCriteriaPlugin(
   criteria: tanagraUI.UICriteria,
   entity?: string
-): CriteriaPlugin<object> {
+): CriteriaPlugin<string> {
```
marikomedlock, line 340: Is this because you're now passing around the serialized string instead of the config or data objects taht used to be defined in the plugin ts class?

ui/src/criteria/unhintedValue.tsx
```diff
@@ -240,3 +259,60 @@ function UnhintedValueInline(props: UnhintedValueInlineProps) {
     </GridLayout>
   );
 }
+
+function decodeComparisonOperator(
+  operator: dataProto.UnhintedValue_ComparisonOperator
+): tanagraUI.UIComparisonOperator {
+  switch (operator) {
```
marikomedlock, line 266: Any chance there's a `decodeByName` utility function for ts enums? Just curious, I have this same problem on the Java side converting the OpenAPI enums.

ui/codegen.sh
tjennison-google: Or alternatively, maybe all of this code gen should be run from gradle so there's a single place for it. The only use case I can think of for keeping them separate is some small subset of unit tests that use the generated code but also don't need the backend.

ui/src/criteria/unhintedValue.tsx
tjennison-google: That can't really work here because the proto enums are integers, otherwise that is possible, though as the data changes that sometimes breaks down.

ui/src/cohort.ts
tjennison-google: Yes. I'm considering the possibility of switching back to passing around the internal data object and only converting to/from protos right at the edge of API class. That would require some more structural changes though and I wanted to keep that separate.

PR #723, opened 2024-02-13
Add underlay configuration for criteria selectors and prepackaged data features.

marikomedlock:


docs/generated/UNDERLAY_CONFIG.md
```diff
@@ -188,6 +193,136 @@ Name of the field or column name that maps to the occurrence entity id. Required
...
+**required** String
+
+Name of the primary UI display plugin. (e.g. selector for condition, not any of the modifiers).
+
+This plugin name is stored in the application database, so once there are cohorts or data features that use this selector, you can't change the plugin names.
+
+### SZCriteriaSelector.pluginConfig
+**required** String
+
+Serialized configuration of the primary UI display plugin e.g. "{"attribute":"gender"}".
```
tjennison-google, line 247: Might be worth changing things like this to reference serialized protos instead of JSON.

tjennison-google (approved):

docs/generated/UNDERLAY_CONFIG.md
marikomedlock: Are you talking about the example JSON, or the term "serialized configuration", or both?

I was expecting to continue using JSON as the serialization format, to be consistent with the rest of the underlay config, and then deserialize using a protobuf utility (in Java, [JsonFormat](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/util/JsonFormat)).

marikomedlock: Going to merge this pr now, and make any changes in a follow-on pr once we finish discussing.

marikomedlock: For the config files, I lean towards JSON over another human readable format, because that's what the other config files use.

For the `/v2/underlay/{underlayName}` API, the plugin configs are currently also being returned as JSON. I could change that, now that I'll have the schemas -- I don't think the file & wire formats need to match. Although given we're not using protos for the underlay config (yet anyway), we'd be returning an object with properties that have been serialized differently (e.g. entities in JSON, plugin config in proto packed).

For the storage in postgres, the plugin data is currently being stored as it comes over the wire. I could change this too though, now that I'll have the schemas -- I don't think the wire & storage formats need to match. Though I'm not expecting these plugin configs to be huge, so maybe better to prioritize easier debugging at this point.

tjennison-google: > For the config files, I lean towards JSON over another human readable format, because that's what the other config files use.

I agree, JSON makes sense.

> 
> For the `/v2/underlay/{underlayName}` API, the plugin configs are currently also being returned as JSON. I could change that, now that I'll have the schemas -- I don't think the file & wire formats need to match. Although given we're not using protos for the underlay config (yet anyway), we'd be returning an object with properties that have been serialized differently (e.g. entities in JSON, plugin config in proto packed).

To make sure we're on the same page, the selector information with the name etc. is JSON that comes through the SZxxx classes and I agree should definitely stay the way it is. Within that there's the plugin specific proto config which will be serialized as a string, either JSON or packed proto. Are you suggesting we should serialize that as JSON as well so they're the same, even though it'll be serialized into a string? I don't think I feel strongly either way so we could go with JSON everywhere.

> 
> For the storage in postgres, the plugin data is currently being stored as it comes over the wire. I could change this too though, now that I'll have the schemas -- I don't think the wire & storage formats need to match. Though I'm not expecting these plugin configs to be huge, so maybe better to prioritize easier debugging at this point.

I could switch the selection data to store as JSON in that case, but I'll maybe leave that to the last step and drop backwards compatibility with the non-proto data at the same time.

PR #729, opened 2024-02-20
Unpeg grpc library version.

marikomedlock:
Test indexing an entity that includes both BQ and Dataflow jobs, completed successfully:
`tanagra index entity --names condition --indexer-config=aouSR2019q4r4_verily`

I'll plan to re-enable dependency locking in a follow-on PR to prevent this kind of problem in the future.


tjennison-google (approved):

freemabd (approved):
Thanks

PR #734, opened 2024-02-21
Change SD height, weight is_clean from an enum to a flag.

marikomedlock:
As part of this PR, I partially re-indexed the SD 0831 verily underlay to include these two new attributes.

### DIFF
--- a/underlay/src/main/resources/config/datamapping/sd/entity/height/all.sql
+++ b/underlay/src/main/resources/config/datamapping/sd/entity/height/all.sql
@@ -1,8 +1,7 @@
 SELECT
   mo.measurement_id,
   mo.person_id,
-  (SELECT COUNT(*) FROM `${omopDataset}.x_vs_wh` AS xvw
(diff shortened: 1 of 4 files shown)


tjennison-google (commented):
Per offline discussion, the UI will need to add support for boolean fields in the criteria plugin so this probably needs to be an addition rather than changing the existing attribute. In the meantime, if clean+raw gives the correct total we can just consider renaming the options to clarify that.

marikomedlock:
> Per offline discussion, the UI will need to add support for boolean fields in the criteria plugin so this probably needs to be an addition rather than changing the existing attribute. In the meantime, if clean+raw gives the correct total we can just consider renaming the options to clarify that.

We should get the correct totals by selecting both options in the current UI, so this is not as urgent now. Longer term though, I think a boolean flag makes more sense for this data, so I can leave it in and then won't be blocking you when you get to boolean support in the UI criteria plugin.

tjennison-google (approved):
I never noticed the column was called `x_invalid` before. What a strange name for it.

PR #737, opened 2024-02-21
[VDA-1413] Add new underlay sd beta 

chenchals:
- added new underlay for taking care of `GRID` as well as `optimizeGroupByAttributes` 
- update UI
- merged with changes to total-counts
- Merged all code updated underlay
- Refactored underlay and tested


marikomedlock (commented):
Could you edit the existing datamapping/sd/ files (e.g. add person_source_value attribute), instead of cloning into a new datamapping/sd_beta/ sub-directory? Or is the plan to maintain both?

marikomedlock:
> I could have edited the existing sd -> but then that would clobber the existing. Also wanted to make sure we can do some versioning when we get data refresh. Previous SD-Discover app had xx-alpha and xx-beat and between refreshes, the 'dataset' pointer would change to either 'xx=alpha' or 'xx-beta'. I am not sure what is the future state of sd-cohort-builder. But for current troubleshooting, I opted to create a copy of sd and call it sd-beta -> and edited the files in them. I can rename all to plain 'sd' if you think that is a better approach now. i.e users will not need to create fresh study etc...

RE versioning: Agreed that for future data refreshes, we should take a versioning approach. Currently though, we just have the one `0831` dataset right? It's fine with me to rename the directory and underlay to sd_0831 or sd_beta, but I mean to say that I don't think we need two versions of the files, until we have a new refresh. So rename, not clone is what I was hoping for.

> I should be able to rename sd_beta to sd (after deleting sd dir) and retain the datasetId pointer to `sd-vumc-tanagra-test.indexed_sd_20230831_beta` where appropriate right?

Underlay name and BQ dataset name are not connected anywhere except in the service and indexer files, so no need change them together.

chenchals:
> > I could have edited the existing sd -> but then that would clobber the existing. Also wanted to make sure we can do some versioning when we get data refresh. Previous SD-Discover app had xx-alpha and xx-beat and between refreshes, the 'dataset' pointer would change to either 'xx=alpha' or 'xx-beta'. I am not sure what is the future state of sd-cohort-builder. But for current troubleshooting, I opted to create a copy of sd and call it sd-beta -> and edited the files in them. I can rename all to plain 'sd' if you think that is a better approach now. i.e users will not need to create fresh study etc...
> 
> RE versioning: Agreed that for future data refreshes, we should take a versioning approach. Currently though, we just have the one `0831` dataset right? It's fine with me to rename the directory and underlay to sd_0831 or sd_beta, but I mean to say that I don't think we need two versions of the files, until we have a new refresh. So rename, not clone is what I was hoping for.
> 
> > I should be able to rename sd_beta to sd (after deleting sd dir) and retain the datasetId pointer to `sd-vumc-tanagra-test.indexed_sd_20230831_beta` where appropriate right?
> 
> Underlay name and BQ dataset name are not connected anywhere except in the service and indexer files, so no need change them together.

OK! Let me take another swing at it after the meeting. Basically: use `sd` and change references to datasetId in `sd020230831_vumc.json` files in service and indexer dirs from `indexed_sd_20230831` to `indexed_sd_20230831_beta`

underlay/src/main/resources/config/datamapping/sd/entity/weight/entity.json
```diff
@@ -4,8 +4,8 @@
   "attributes": [
     { "name": "id", "dataType": "INT64", "valueFieldName": "measurement_id" },
     { "name": "person_id", "dataType": "INT64" },
+    { "name": "GRID", "dataType": "STRING", "valueFieldName": "person_source_value"},
     { "name": "is_clean", "dataType": "INT64", "displayFieldName": "is_clean_name", "isComputeDisplayHint": true },
-    { "name": "x_invalid_bool", "dataType": "BOOLEAN" },
```
marikomedlock, line 8: Did you mean to remove this?

underlay/src/main/resources/config/datamapping/sd/entity/person/entity.json
```diff
@@ -3,11 +3,12 @@
   "allInstancesSqlFile": "all.sql",
   "attributes": [
     { "name": "id", "dataType": "INT64", "valueFieldName": "person_id" },
+    { "name": "person_id", "dataType": "INT64"},
```
marikomedlock, line 6: This is already covered by the `id` attribute, why add a second one?

underlay/src/main/resources/config/datamapping/sd/entity/height/entity.json
```diff
@@ -4,8 +4,8 @@
   "attributes": [
     { "name": "id", "dataType": "INT64", "valueFieldName": "measurement_id" },
     { "name": "person_id", "dataType": "INT64" },
+    { "name": "GRID", "dataType": "STRING", "valueFieldName": "person_source_value"},
     { "name": "is_clean", "dataType": "INT64", "displayFieldName": "is_clean_name", "isComputeDisplayHint": true },
-    { "name": "x_invalid_bool", "dataType": "BOOLEAN" },
```
marikomedlock, line 8: Did you mean to remove this?

underlay/src/main/resources/config/datamapping/sd/entity/height/all.sql
```diff
@@ -9,7 +10,6 @@ SELECT
     WHEN xvw.x_invalid = 'N' THEN 'Clean'
     ELSE 'Raw'
   END AS is_clean_name,
-  CASE WHEN xvw.x_invalid = 'N' THEN TRUE ELSE FALSE END x_invalid_bool,
```
marikomedlock, line 12: Did you mean to remove this?

underlay/src/main/resources/config/datamapping/sd/entity/weight/all.sql
```diff
@@ -9,7 +10,6 @@ SELECT
     WHEN xvw.x_invalid = 'N' THEN 'Clean'
     ELSE 'Raw'
   END AS is_clean_name,
-  CASE WHEN xvw.x_invalid = 'N' THEN TRUE ELSE FALSE END x_invalid_bool,
```
marikomedlock, line 12: Did you mean to remove this?

marikomedlock (commented):
Thanks for the rename, this is much easier to see what changed. Everything looks good, two questions inline.

I think you may have deleted a recent change I made when you merged (x_invalid_bool attribute in height and weight entities). If that was by accident and you want to put it back, you don't need to re-index everything, just the height/weight entities (other entities and all groups should be fine).

underlay/src/main/resources/config/datamapping/sd/entity/weight/entity.json
chenchals: It possibly was removed from previous sd directory. I checked the ENT_weight in both indexed_sd20230831_beta and indexed_sd20230831 (the previous indexed dataset) and the coldefs match as well as row counts.

underlay/src/main/resources/config/datamapping/sd/entity/person/entity.json
chenchals: This would be explicit for users - since id-in different domains refer to different elements. so yes I needed this.

underlay/src/main/resources/config/datamapping/sd/entity/weight/entity.json
marikomedlock: I just added this added this attribute to height/weight yesterday in PR #734, v0.0.384. It changes the "is measurement clean" attribute from an enum to a boolean, which is more correct based on our conversation at the meeting yesterday. I left the existing enum there as well, so as not to break the UI until the UI switches over to use the boolean. So, it wouldn't have been in the previous index dataset, but I think we should keep it.

marikomedlock (approved):

underlay/src/main/resources/config/datamapping/sd/entity/weight/all.sql
chenchals: It possibly was removed from previous sd directory. I checked the ENT_weight in both indexed_sd20230831_beta and indexed_sd20230831 (the previous indexed dataset) and the coldefs match as well as row counts.  And the counts form the UI in my local matches that seen in sd-discover. I updated the excel doc with my findings

PR #741, opened 2024-02-22
[no ticket] add prod sd config for underlay

chenchals:
- added dev, prod and test configs for service
- No change to indexer -> we copy the indexed dataset to needed datasets in test and prod


underlay/src/main/resources/config/service/sd/sd020230831_test.json
```diff
@@ -11,7 +11,7 @@
     },
     "indexData": {
       "projectId": "sd-vumc-tanagra-test",
-      "datasetId": "indexed_sd_20230831_beta"
+      "datasetId": "indexed_sd_20230831"
```
marikomedlock, line 14: Is this change intended?

marikomedlock (approved):

underlay/src/main/resources/config/service/sd/sd020230831_test.json
chenchals: yes-> we copy

PR #749, opened 2024-02-28
[VDA-1421-1423] Update underlays for ui/occurrence count bugs

chenchals:
- updated underlays for
  - occurrenceCount - `noteOccurrence`
  - universal search results -source codes- `cpt4`, `cpt4Person`
  - universal search results -source codes- `phewas`, `phewasPerson`
- ran indexing
- verified on local deploy

- Occurrence count - Documents (fixed: modifier Occurrence count gave error)
image

- Search results: CPT4 (fixed: did not show code for CPT4)
image

- Search results: pheWAS ( fixed: 1st item did not show name and code)
image

### DIFF
--- a/underlay/src/main/resources/config/datamapping/sd/entity/cpt4/entity.json
+++ b/underlay/src/main/resources/config/datamapping/sd/entity/cpt4/entity.json
@@ -1,25 +1,28 @@
 {
-  "name": "phewas",
+  "name": "cpt4",
   "allInstancesSqlFile": "all.sql",
   "attributes": [
     { "name": "id", "dataType": "INT64" },
+    { "name": "concept_id", "dataType": "INT64" },
     { "name": "name", "dataType": "STRING" },
-    { "name": "label", "dataType": "STRING" },
     { "name": "vocabulary", "dataType": "STRING", "valueFieldName": "type" },
+    { "name": "is_standard", "dataType": "STRING", "isComputeDisplayHint": true },
     { "name": "concept_code", "dataType": "STRING" },
-    { "name": "numeric_code", "dataType": "DOUBLE" }
+    { "name": "label", "dataType": "STRING" }
   ],
   "idAttribute": "id",
   "textSearch": {
-    "attributes": [ "name", "code" ]
+    "attributes": [ "id", "name", "code" ]
   },
   "hierarchies": [
     {
       "childParentIdPairsSqlFile": "childParent.sql",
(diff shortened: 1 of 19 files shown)


underlay/src/main/resources/config/underlay/sd/underlay.json
```diff
@@ -75,7 +75,7 @@
 
     "omop/notePerson",
 
-    "omop/cpt4Person",
+    "sd/cpt4Person",
```
marikomedlock, line 78: Did this change at all? If not, you can keep using the `omop` version, no need to duplicate, even if the entities referenced in the group changed.

underlay/src/main/resources/config/datamapping/sd/entity/cpt4/entity.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "cpt4",
+  "allInstancesSqlFile": "all.sql",
+  "attributes": [
+    { "name": "id", "dataType": "INT64" },
+    { "name": "concept_id", "dataType": "INT64" },
+    { "name": "name", "dataType": "STRING" },
+    { "name": "vocabulary", "dataType": "STRING", "valueFieldName": "type" },
+    { "name": "is_standard", "dataType": "STRING", "isComputeDisplayHint": true },
+    { "name": "concept_code", "dataType": "STRING" },
```
marikomedlock, line 10: Is the only change here renaming the attribute from `code` -> `concept_code`? If so, could you just modify the existing entity in place? SD is the only underlay that uses the `omop/cpt4` definition, aside from the test underlays, so you don't need to worry about breaking other underlays in this case.

marikomedlock (commented):
Thanks for these changes! Two questions inline about updating config files in place instead of copying over -- regardless of what you decide to do, shouldn't require reindexing.

underlay/src/main/resources/config/underlay/sd/underlay.json
chenchals: since I change the cpt4, i preferred to find the cptePerson in the same sd/ tree. Layter when standardizing I am planning to make a copy of **all**  omop -> so that we would not have to update if you do change any omop tables going forward.

marikomedlock: I see, same as you did for the AoU. OK makes sense.

marikomedlock (approved):

PR #751, opened 2024-02-29
Use temporary table for export queries. Parallelize entity export jobs.

marikomedlock:
1. Changed the export from BQ to GCS files to write the query results to a temporary table, and then export the full table. Previously, it used the `EXPORT DATA` [statement](https://cloud.google.com/bigquery/docs/reference/standard-sql/other-statements#export_data_statement) appended to the query to write to GCS without explicitly writing to a table. This should get us to ~1GB maximum size per export file more reliably.

2. Parallelized the export of each file of entity instances. This should reduce the time when exporting data feature sets that span multiple tables. Annotation files are still run serially, plan to parallelize those in a future PR.

-------------

This PR requires some devops changes to existing deployments. In particular, we need a BQ dataset that the service SA can read/write to, and has the default table expiration set very low (e.g. 1 day). This BQ dataset needs to be in the same location as the BQ dataset we're exporting from. If there are >1 underlays, with >1 locations, you can supply a list of available BQ datasets for export and it will pick a compatible one. This is similar to how the GCS buckets used to temporarily host export files work.

I also updated the name of the export project from `gcsProjectId` -> `gcpProjectId` since now it contains GCS buckets and BQ datasets. The new env var names are specified in the generated documentation and example values are in the local server script and application.yml file.

As part of this PR, I created BQ datasets to host these temp tables in the `broad-tanagra-dev` project that we use for GHA testing:
```
bq mk --location=us-central1 --default_table_expiration=86400 broad-tanagra-dev:service_export_uscentral1
bq show --format=prettyjson broad-tanagra-dev:service_export_uscentral1
```

### DIFF
--- a/underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQExecutor.java
+++ b/underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQExecutor.java
@@ -124,7 +124,7 @@ public class BQExecutor {
     String bucketName =
         getCloudStorageService()
             .findBucketForBigQueryExport(exportProjectId, exportBucketNames, datasetLocation);
-    String gcsUrl = String.format("gs://%s/%s", bucketName, tempTableName);
+    String gcsUrl = String.format("gs://%s/%s.gzip", bucketName, tempTableName);
     LOGGER.info("Exporting temporary table to GCS file: {}", gcsUrl);
     Job exportJob = getBigQueryService().exportTableToGcs(tempTableId, gcsUrl, "GZIP", "CSV");
     if (exportJob == null) {
--- a/underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQExecutor.java
+++ b/underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQExecutor.java
@@ -20,6 +20,7 @@ import com.google.common.collect.Iterables;
 import java.io.IOException;
 import java.time.Instant;
 import java.util.List;
+import java.util.Random;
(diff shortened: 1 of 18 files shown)


underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQExecutor.java
```diff
@@ -89,37 +93,49 @@ public SqlQueryResult run(SqlQueryRequest queryRequest) {
...
-          "No (or invalid) filename specified for GCS export ({}), using default: {}",
-          fileNamePrefix,
-          validatedFilename);
-    } else {
-      validatedFilename = fileNamePrefix;
-    }
+    LOGGER.info("Exporting BQ query: {}", queryRequest.getSql());
+
+    // Create a temporary view with the results of the query.
+    final String tempTableName = TEMPORARY_TABLE_BASE_NAME + '_' + Instant.now().toEpochMilli();
```
tjennison-google, line 102: Would it be easy to also include a somewhat unique element of the request in here as well? Or nanoseconds? It's obviously not very likely to have a collision but worth decreasing the odds further if it's easy enough.

tjennison-google (approved):

underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQExecutor.java
marikomedlock: Good idea. Added nanosec and a random int.

PR #757, opened 2024-03-07
Add feature flag for backend filter building.

marikomedlock:
Also added an unauthenticated API endpoint to pass feature flags to the UI. This will allow us to hook up the UI to call new endpoints for cohort breakdowns and test it out behind a feature flag both locally and in the Verily env before making this the default behavior.


tjennison-google (commented):
It's a bit of a pain for me to add state like this that I need to ensure is available to every page. Unless there's a strong reason to pass it from the backend I'd rather have a config option on the UI side, particularly for something transient like this.

marikomedlock:
> It's a bit of a pain for me to add state like this that I need to ensure is available to every page. Unless there's a strong reason to pass it from the backend I'd rather have a config option on the UI side, particularly for something transient like this.

RE config option for UI: Sounds good to me, I'll leave you to do the UI side however is most convenient for you. I'd like to keep the service config property here for the backend side, since I think that will be easier for me, and override it in our test env with an env var.

RE passing feature flags: Even if you won't use this feature flags endpoint for this particular use case, do you want me to leave it for the future?

marikomedlock:
Per offline discussion: I'm going to remove the API, since the UI will handle this temporary feature flag differently, but keep the backend implementation.

tjennison-google (approved):

PR #759, opened 2024-03-08
Add service config property to control multi-threading.

marikomedlock:
`tanagra.feature.maxChildThreads` service config property controls how multi-threading is or isn't used in the application. The environment variable `TANAGRA_FEATURE_MAX_CHILD_THREADS` can also be used to specify this property for a given deployment.

Leave the property unset to allow the application use multi-threading where possible (this is the current behavior). Set the property to 0 or 1 to disable multi-threading and always run things serially. Set the property to some other value to control the maximum number of child threads that are spawned per request.

Currently, the application only uses multi-threading for export requests. Running things serially or limiting the maximum size of the thread pool will significantly affect performance, especially for requests that include many tables. We are adding this config now to help debug some possible AppEngine limitations on multi-threading.


docs/generated/APPLICATION_CONFIG.md
```diff
@@ -302,6 +302,13 @@ When true, artifacts can be created, updated and deleted. Artifacts include stud
 
 *Default value:* `false`
 
+### tanagra.feature.maxChildThreads
+**optional** String
+
+The maximum number of child threads a single request can spawn. When unset, the application will default to using multi-threading. When set to 0 or 1, the application will only ever run things serially.
+
+*Environment variable:* `TANAGRA_FEATURE_MAX_CHILD_THREADS`
+
```
chenchals, line 313: The doc is not very clear: set to 0 -> no multi-hreading set to 1- also no multi threading. What is the difference? What about 2?

marikomedlock: Currently no difference between 0 and 1, because for export they'd be the same thing. But in the future, we could have cases where it would improve performance to run a single thread in the background.

2 = at most 2 child threads will be spawned.

chenchals (approved):
LGTM

PR #763, opened 2024-03-13
Individual file export fixes.

marikomedlock:
- Skip exporting empty files. After creating the temp table, we now check whether there are >0 rows before kicking off the export to GCS job. This addresses the problem of errors unzipping empty files.
Screenshot 2024-03-13 at 9 05 42 AM
- Include csv in compressed file name. The export job takes the compressed file name and strips `.gzip` from the end to get the uncompressed file name. This addresses the problem of the unzipped file not having any file extension.
- Include entity, first cohort name+id, study id, underlay name in file name. This is to make the files easier to recognize after the user downloads them.
Screenshot 2024-03-13 at 11 37 06 AM


service/src/main/java/bio/terra/tanagra/service/export/impl/IndividualFileDownload.java
```diff
@@ -44,10 +45,18 @@ public void initialize(DeploymentConfig deploymentConfig) {
...
+                    + request.getCohorts().get(0).getId())
+            + (request.getCohorts().size() > 1
+                ? "plus" + (request.getCohorts().size() - 1) + "more"
+                : "");
     Map<String, String> entityToGcsUrl =
-        request.writeEntityDataToGcs("tanagra_${entity}_" + Instant.now());
+        request.writeEntityDataToGcs("${entity}_cohort" + cohortRef + "_${random}");
     Map<Cohort, String> cohortToGcsUrl =
-        request.writeAnnotationDataToGcs("tanagra_${cohort}_" + Instant.now());
+        request.writeAnnotationDataToGcs("annotations_cohort${cohort}_${random}");
```
chenchals, line 60: including study_id also would be better, IMHO. Similar to the IpynbFileDownload.java#71 below

chenchals (commented):
LGTM. Should possibly include study_id in the filename

service/src/main/java/bio/terra/tanagra/service/export/impl/IndividualFileDownload.java
tjennison-google: If anything, I think the underlay would actually be the most valuable to add since it seems a lot more likely to export the same data/cohorts on different data releases than entirely different studies

tjennison-google (approved):

service/src/main/java/bio/terra/tanagra/service/export/impl/IndividualFileDownload.java
marikomedlock: Easy enough to include both studyId and underlayName. We may have to pare this down in the future if the names get too long, but we can include everything at first.

I guess this is more of a product question, but I'm skeptical about the utility of the ids (cohort & study). They're not visible anywhere in the UI, so seems unlikely to be useful to the user. And the main purpose of having a readable filename here is so the user can make sense of their downloads folder. Secondary benefit is helping with debugging, but the random suffix would be more reliable for that anyway. We can bring up in the weekly meeting, as a low priority change for now.

marikomedlock: RE ipynb file download: That line is actually writing the study name/id into the ipynb file, not as the filename.

PR #764, opened 2024-03-13
OpenAPI updates to support backend filter building.

marikomedlock:
New and updated endpoint definitions to support backend filter building.
- New cohort counts endpoint to get the group/section counts and cohort breakdown.
- New export preview endpoint to get the list of entity instances in the export preview page and data feature set page.
- Updated export endpoint to include the concept set ids.

Implementation of these endpoints will come in a future PR.


service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2408,6 +2480,25 @@ components:
         round-trip:
           type: string
 
+    ExportPreviewRequest:
+      type: object
+      properties:
+        study:
+          type: string
+          nullable: false
```
tjennison-google, line 2488: I think this is non-nullable by default and instead it should be made non-optional by putting it in the `required:` list. As far as I can tell, OpenAPI differentiates between nullable and optional, though I don't know how that maps to the generated Java code.

tjennison-google (approved):

PR #767, opened 2024-03-14
Implement queryCohortCounts endpoint.

marikomedlock:
- Hooked up the `queryCohortCounts` endpoint to the backend filter building infrastructure.
- Changed from using the `pluginName` field of `Criteria` to the new `selectorOrModifierName` field to fix the config lookup.
- Changed proto deserialization to try JSON format first, and then base64-encoded protobuf binary format. The selector and modifier plugin configs will all be in JSON. The plugin data that comes over the wire and is stored verbatim in the application DB is currently all in base64-encoded protobuf binary format, though we may change that to also use JSON in the future. The plugin data that lives in the config as part of the prepackaged criteria definitions will all be in JSON.
- Added support for instance-level modifiers (e.g. labs & measurements concept-specific enum/range) in the entity group plugin for `CriteriaOccurrence` entity groups only.
- I did some basic testing locally with the Swagger UI for each of the plugin types (gender, age, biovu genetic data, biovu dna, measurements, documents, blood pressure), and modifier types (attribute, group by, instance-level) to make sure the counts returned by this endpoint match at the criteria group, criteria group section, and cohort levels.

### DIFF
--- a/underlay/src/main/java/bio/terra/tanagra/utils/ProtobufUtils.java
+++ b/underlay/src/main/java/bio/terra/tanagra/utils/ProtobufUtils.java
@@ -4,9 +4,22 @@ import bio.terra.tanagra.exception.InvalidConfigException;
 import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.protobuf.Message;
 import com.google.protobuf.util.JsonFormat;
+import java.util.Base64;
 
 public final class ProtobufUtils {
+
   private ProtobufUtils() {}
+
+  @SuppressWarnings({"checkstyle:EmptyCatchBlock", "PMD.EmptyCatchBlock"})
+  public static <T extends Message.Builder> T deserializeFromJsonOrProtoBytes(
+      String serialized, T builder) {
+    try {
+      return deserializeFromJson(serialized, builder);
+    } catch (InvalidConfigException icEx) {
+      // Don't throw an exception, try to deserialize from Protobuf bytes instead.
+    }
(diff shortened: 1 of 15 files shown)


underlay/src/main/java/bio/terra/tanagra/utils/ProtobufUtils.java
```diff
@@ -4,10 +4,23 @@
...
+import java.util.Base64;
 
 public final class ProtobufUtils {
+
   private ProtobufUtils() {}
 
+  @SuppressWarnings({"checkstyle:EmptyCatchBlock", "PMD.EmptyCatchBlock"})
+  public static <T extends Message.Builder> T deserializeFromJsonOrProtoBytes(
+      String serialized, T builder) {
+    try {
```
tjennison-google, line 16: I have similar code in the UI and I've found it annoying when debugging because of all the spurious exceptions so I was thinking about checking if the first character is a "{" (invalid for base 64) rather than trying to deserialize.

tjennison-google (approved):

underlay/src/main/java/bio/terra/tanagra/utils/ProtobufUtils.java
marikomedlock: This should swallow the exceptions, not throw or even log them, so I'm hoping I won't have the same debugging irritation. That's a nice tip though, I think it would make this code a little less ugly. I'll make the change in a separate PR.

PR #770, opened 2024-03-15
Build group by filters using multiple attributes

tjennison-google:
Update existing configs to specify multiple attributes.

### DIFF
--- a/ui/src/criteria/unhintedValue.tsx
+++ b/ui/src/criteria/unhintedValue.tsx
@@ -5,7 +5,7 @@ import OutlinedInput from "@mui/material/OutlinedInput";
 import Select, { SelectChangeEvent } from "@mui/material/Select";
 import Typography from "@mui/material/Typography";
 import { CriteriaPlugin, registerCriteriaPlugin } from "cohort";
-import { FilterType } from "data/filter";
+import { FilterType, makeArrayFilter } from "data/filter";
 import { CommonSelectorConfig } from "data/source";
 import { ComparisonOperator } from "data/types";
 import { useUpdateCriteria } from "hooks";
@@ -105,36 +105,46 @@ class _ implements CriteriaPlugin<string> {
     };
   }
 
-  generateFilter() {
+  generateFilter(occurrenceId: string) {
     const decodedData = decodeData(this.data);
 
     if (this.config.groupByCount) {
       return null;
     }
 
-    return {
-      type: FilterType.Attribute,
-      attribute: this.config.attribute,
-      ranges: [rangeFromData(decodedData)],
-    };
+    const attributes = this.config.attributes[occurrenceId];
+    if (!attributes || !attributes.values.length) {
+      throw new Error(`No attributes found for occurrence: ${occurrenceId}`);
+    }
+
+    return makeArrayFilter(
+      {},
+      attributes.values.map((a) => ({
+        type: FilterType.Attribute,
+        attribute: a,
+        ranges: [rangeFromData(decodedData)],
+      }))
+    );
   }
 
-  groupByCountFilter() {
+  groupByCountFilter(occurrenceId: string) {
     const decodedData = decodeData(this.data);
 
     if (!this.config.groupByCount) {
       return null;
     }
 
+    const attributes = this.config.attributes[occurrenceId];
+
     return {
-      attribute: this.config.attribute,
+      attributes: attributes?.values ?? [],
       operator: decodedData.operator,
       value: decodedData.min,
     };
   }
 
   filterEntityIds() {
-    return [this.config.attribute ?? ""];
+    return Object.keys(this.config.attributes);
   }
 }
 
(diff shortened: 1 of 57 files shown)


ui/src/criteria/unhintedValue.tsx
```diff
@@ -105,36 +105,49 @@ class _ implements CriteriaPlugin<string> {
...
     }
 
-    return {
-      type: FilterType.Attribute,
-      attribute: this.config.attribute,
-      ranges: [rangeFromData(decodedData)],
-    };
+    const attributes = this.config.attributes[occurrenceId];
+    if (!attributes || !attributes.values.length) {
+      throw new Error(`No attributes found for occurrence: ${occurrenceId}`);
```
marikomedlock, line 117: Does this disallow no group by attributes? If so, I think we do want to accommodate that use case (e.g. has at least 2 height measurements) and the backend should already allow that.

underlay/src/main/resources/config/display/sd/criteriaselector/icd10CM/startDateGroupByCount.json
```diff
@@ -1,4 +1,29 @@
 {
   "groupByCount": true,
-  "attribute": "start_date"
-}
\ No newline at end of file
+  "attributes": {
+    "conditionOccurrence": {
+      "values": [
+        "start_date",
+        "condition"
```
marikomedlock, line 7: The JOIN column/attribute for the ICD codes is `source_criteria_id` (which maps to e.g. `condition_source_concept_id`), instead of the domain-named attribute e.g. `condition` (which maps to e.g. `condition_concept_id`) that the standard codes use.

For CPT-4 and PheWAS, you can use the same `source_criteria_id` attribute here. Thinking about this now though, we probably need to do some additional testing for those two because I think it's possible to have multiple `source_concept_id` values per code. I don't know if that actually happens, but the way the config SQL is written it is possible. If so, we'd need to do something further to address that -- maybe adding an attribute for the CPT-4 and PheWAS codes to the occurrence entities and then specifying that here.

underlay/src/main/resources/config/display/sd/criteriaselector/measurement/startDateGroupByCount.json
```diff
@@ -1,4 +1,11 @@
 {
   "groupByCount": true,
-  "attribute": "start_date"
-}
\ No newline at end of file
+  "attributes": {
+    "observationOccurrence": {
+      "values": [
+        "start_date",
+        "observation"
```
marikomedlock, line 7: `measurementOccurrence` and `measurement`?

underlay/src/main/java/bio/terra/tanagra/filterbuilder/impl/core/utils/GroupByCountSchemaUtils.java
```diff
@@ -66,14 +67,16 @@ public static Map<Entity, List<Attribute>> getGroupByAttributesPerOccurrenceEnti
...
-        && !groupByModifierConfig.getAttribute().isEmpty()) {
-      String attributeName = groupByModifierConfig.getAttribute();
+    if (groupByModifierConfig.getAttributesMap() != null) {
       occurrenceEntities.stream()
           .forEach(
               occurrenceEntity ->
                   groupByAttributesPerOccurrenceEntity.put(
-                      occurrenceEntity, List.of(occurrenceEntity.getAttribute(attributeName))));
+                      occurrenceEntity,
+                      groupByModifierConfig.getAttributesMap().get(occurrenceEntity).getValuesList()
```
marikomedlock, line 76: Thanks for updating here too!

marikomedlock (commented):
Among the config files, I only reviewed the SD ones closely.

ui/src/criteria/unhintedValue.tsx
tjennison-google: Yes. Changed to allow empty attribute lists.

underlay/src/main/resources/config/display/sd/criteriaselector/icd10CM/startDateGroupByCount.json
tjennison-google: Changed all of the source entities to `source_criteria_id`. CPT-4 and PheWAS issues to be investigated and fixed separately per offline discussion.

underlay/src/main/resources/config/display/sd/criteriaselector/measurement/startDateGroupByCount.json
tjennison-google: Good catch.

ui/src/criteria/unhintedValue.tsx
marikomedlock: OK -- was this change elsewhere? I don't think this particular section changed at all.

marikomedlock (approved):

ui/src/criteria/unhintedValue.tsx
tjennison-google: Yeah, the group by version is below. This could potentially be removed at some point since we're not using it anymore.

PR #779, opened 2024-03-22
Add file outputs to export API definition.

marikomedlock:
Updated the export endpoint definition:
- Added a possible error message for the entire export request (e.g. "No template ipynb file found.").
- Added a separate `ExportLinkResult` for each file generated by the model. Each link result can have it's own error and/or message. The message can be used for when a file is not generated, on purpose (e.g. "No entity data.").
- The export plugins will control the display in the Export dialog with tags. The tags array for each link result is a list of ancestors (i.e. order matters), and lets each plugin define it's own display hierarchy.
- Fixed a bug with converting the possible values of the status flag.

This PR does NOT include implementation of these changes, just the API definition changes. Implementation to follow in a separate PR once we've got the API we want.


service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2612,15 +2612,43 @@ components:
...
+        displayName:
+          type: string
+        url:
+          type: string
+        entity:
+          type: string
+        cohortId:
+          type: string
+        cohortDisplayName:
+          type: string
```
marikomedlock, line 2646: Is it useful to have both the id & display name here? Or if you already have the cohort definitions handy, I can just delete the display name.

tjennison-google: I do have access to the cohort names but I don't understand how this works compared to the current version. Is it one or the other? Is the idea that entity would be set for the data and cohort would be set for annotations? That seems very specific to the download plugin in any case. Are the sections/tags gone?

marikomedlock: Ah I'd forgotten about the sections/tags. I'll add that field back. And I can set it to "Data"/"Annotations"/"Notebook file" in the export model implementations.

RE entity/cohort properties:
It would be one or the other or neither:
- Entity data export only has entity defined.
- Annotation data export only has cohort defined.
- Notebook file has neither defined.

I guess there are a few options for how we pass out the entity/cohort mapping to the files:
- Treat entity/cohort name as another tag.
- Pass out the entity/cohort name as separate properties.
- Set the entity/cohort name as the `fileDisplayName` instead of the actual file name e.g. `person_cohortFirstCohort_Almt18nxHf_gzI3doyGzR_cmssynpuf_1710864450_1730435588.csv.gzip`.

Thoughts? From the backend perspective, I don't have a strong preference. (I'll put some time on your calendar so we can discuss briefly too.)

marikomedlock: Per offline conversation, we decided on a link result, and the export plugin classes will handle display.

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2612,15 +2612,51 @@ components:
       properties:
         status:
           type: string
-          enum: ["RUNNING", "COMPLETE"]
+          enum: ["COMPLETE", "FAILED"]
```
tjennison-google, line 2615: I like SUCCEEDED rather than COMPLETE to match FAILED.

tjennison-google (approved):

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2612,15 +2612,51 @@ components:
...
         redirectAwayUrl:
           type: string
-          nullable: true
         outputs:
-          description: Key-value map of the export model outputs (e.g. signedUrl=gs://... )
+          description: Key-value map of output data generated by the export model
```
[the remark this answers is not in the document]
marikomedlock: Yes, I'll add a TODO here so we don't forget.

PR #781, opened 2024-03-22
Export error handling tests.

marikomedlock:
- Changed the export result `status` property to an `isSuccessful` boolean. I'd originally used a complete/running status enum, but since we aren't prioritizing implementing long-running jobs (i.e. it will never return status=running), this isn't useful.
- Set a message, but no error, when no file is exported because there is no data. This is stored internally, but not passed out through the API, yet.
  - "Export query returned zero rows. No file generated."
  - "Cohort has no annotation data. No file generated."
- Added an option to the `ipynb` file download to specify a template notebook file with a GCS url. This could be useful in the future, to allow flexibility with the notebook template files, but main reason to add it now is for testing export errors that are not entity or cohort specific.
- Added an icd9cm criteria selector to the cmssynpuf underlay, to allow testing a case where there is one output entity with data and two without.
- Wrote tests for a job-level export error and a no-data message.


service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2610,9 +2610,8 @@ components:
     ExportResult:
       type: object
       properties:
-        status:
-          type: string
-          enum: ["RUNNING", "COMPLETE"]
+        isSuccessful:
```
tjennison-google, line 2613: While not currently needed, changing this to a boolean is potentially limiting in the future. Maybe just change the values to ["SUCCEEDED", "FAILED"]?

tjennison-google (approved):

service/src/main/resources/api/service_openapi.yaml
marikomedlock: OK with me. I'll do this in the other PR where I updated the API -- this change got mixed up between the two.

PR #786, opened 2024-03-26
Optional participant cap per export model.

marikomedlock:
- Added a new per-model configuration property for the maximum number of primary entity instances the model supports. This property is part of the application configuration `tanagra.export.models.numPrimaryEntityCap` for the available export models. Deployments can set this with an environment variable (e.g. `export TANAGRA_EXPORT_MODELS_0_NUM_PRIMARY_ENTITY_CAP=10000`), same as the other export model parameters.
- Passed this cap out to the API. This is not currently used, but in the future could support some informative text in the export flow (e.g. "This export model has a maximum participant cap of 10k.").
- Defined an additional export model for individual file download with the cap set to 10k in the default application configuration. Did the same for the default test configuration, but with the cap set to 5 to make it easier to test.
- Added explicit null checks to suppress the "Invalid [application config value]" messages at startup for undefined, optional properties.

### DIFF
--- a/docs/generated/APPLICATION_CONFIG.md
+++ b/docs/generated/APPLICATION_CONFIG.md
@@ -207,7 +207,7 @@ Configure the export options for each model.
 
 Displayed name of the export model. This is for display only and will be shown in the export dialog when the user initiates an export. Defaults to the display name provided by the export model. It's useful to override the default if you have more than one instance of the same model (e.g. export to workbench parameterized with the dev environment URL, and another parameterized with the test environment URL).
 
-*Environment variable:* `TANAGRA_EXPORT_MODELS_0_DISPLAY_NAME (Note 0 is the list index, so if you have 2 models, you may have 0 and 1 env vars.)`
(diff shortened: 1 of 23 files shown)


tjennison-google (approved):
Alternatively, I could see it being preferable to set the export participant limit per underlay instead. Different data sets might have different rules and having different limits per export method might be worse from a UI perspective. Maybe it would be helpful to get PM input before going ahead with this approach?

The code looks fine though.

marikomedlock:
> Alternatively, I could see it being preferable to set the export participant limit per underlay instead. Different data sets might have different rules and having different limits per export method might be worse from a UI perspective. Maybe it would be helpful to get PM input before going ahead with this approach?
> 
> The code looks fine though.

Good point. I think we may eventually want per-underlay and per-model caps (e.g. SD individual file download 10k cap, SD notebook file no cap, CmsSynpuf individual file download no cap). We previously discussed specifying the list of supported export types per-underlay (e.g. SD individual file download only, eMerge notebook file only) and the implementation of per-underlay caps would be similar.

I'll start a Slack thread to see what other think. Since I think this approach will work for the immediate SD request -- they only have one underlay and one export model anyway -- I'm going to merge and then possibly update with a new PR depending on the feedback we get.

PR #792, opened 2024-04-01
Replace CURRENT_TIMESTAMP with literal to allow paging.

marikomedlock:
- Replaced `CURRENT_TIMESTAMP`, `CURRENT_TIMESTAMP()`, `CURRENT_DATE`, `CURRENT_DATE()` with a literal value. These time-dependent functions prevent BigQuery from using cached query results. This impacts performance when we re-run identical queries and also means we can only get all results in a single request, rather than paging through them.
- Stored the timestamp of when the query was executed in a new property of the `PageMarker`. Then I can reuse this same timestamp for paginated requests, similar to how page tokens work.

### DIFF
--- a/underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQExecutor.java
+++ b/underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQExecutor.java
@@ -5,6 +5,7 @@ import static bio.terra.tanagra.utils.NameUtils.simplifyStringForName;
 import bio.terra.tanagra.api.query.PageMarker;
 import bio.terra.tanagra.api.shared.Literal;
 import bio.terra.tanagra.exception.SystemException;
+import bio.terra.tanagra.query.sql.SqlParams;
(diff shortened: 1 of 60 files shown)


underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQExecutor.java
```diff
@@ -170,6 +174,34 @@ public Pair<String, String> export(
     return Pair.of(getCloudStorageService().createSignedUrl(gcsUrl), fileName);
   }
 
+  public static String replaceFunctionsThatPreventCaching(
```
tjennison-google, line 177: Maybe worth adding this list of supported functions to the docs?

tjennison-google (approved):

underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQExecutor.java
marikomedlock: Done.

PR #798, opened 2024-04-05
Fix date for documents in review underlay config

dolbeew:
In the UI underlay config, the date for the `textSearch` plugin in cohort review (used for Documents) was using `date` when it should be `start_date`, resulting in 'undefined' displayed in the app.
documents_undefined


chenchals (approved):
LGTM. Please update to the latest

PR #799, opened 2024-04-05
Include BioVU DNA selections in criteria title

dolbeew:
Update to the display of the biovu plugin, adding 'Exclude Compromised DNA' and 'Include only samples available for external processing' to the `title` string when selected instead of the `additionalText` array (they are only visible in cohort review when located in `additionalText`)


chenchals (approved):
LGTM. Please update before merge

PR #800, opened 2024-04-05
Update export flow to match mocks

tjennison-google:


ui/src/export.tsx
```diff
@@ -720,178 +730,315 @@ function ExportDialog(
...
+              if (exportState.data?.redirectURL) {
+                redirect(
+                  exportState.data.redirectURL,
+                  absoluteExportURL(params)
+                );
+              }
+            }}
+          >
+            Continue
+          </Button>
```
marikomedlock, line 986: Does this bit give the user a option to Continue even if there were errors?

underlay/src/main/resources/config/underlay/sd/ui.json
```diff
@@ -1,7 +1,4 @@
 {
-  "featureConfig": {
-    "disableExportButton": true
-  },
```
marikomedlock, line 4: 🎉

marikomedlock (approved):

ui/src/export.tsx
tjennison-google: Yes, that's what was in the mocks. Do you think it shouldn't?

marikomedlock: I think it's fine, as long as we also show the errors. Was just making sure I was following.

PR #801, opened 2024-04-05
Fix demographics error for zero participants

tjennison-google:


marikomedlock (approved):
Did you land on an empty chart here instead of the cactus?

PR #809, opened 2024-04-10
Update NodeJS versions in tests

tjennison-google:


marikomedlock (approved):
Is the previous version no longer supported in GHAs?

tjennison-google:
> Is the previous version no longer supported in GHAs?

The previous version doesn't have long term support and failed in at least one instance. It's not clear to me why all of the PRs aren't having an issue but it makes sense to migrate to something stable.

PR #821, opened 2024-04-24
Fr emerge config files

NickofNasa:
config files added for emerge deployments. Underlay value may need to be changed but the rest of the file should be accurate.


underlay/src/main/resources/config/indexer/emerge/emerge20230831.json
```diff
@@ -0,0 +1,25 @@
+{
+  "underlay": "emerge",
```
marikomedlock, line 2: Right, this value needs to point to an [underlay directory](https://github.com/DataBiosphere/tanagra/tree/main/underlay/src/main/resources/config/underlay) that contains 2 files: `underlay.json` and `ui.json`. You could copy from one of the existing underlays as a starting point (e.g. copy the contents of `sd`).

marikomedlock (commented):
Great to see this work getting underway!

PR #829, opened 2024-04-30
Java 17, spring 3.x, library upgrades

dexamundsen:
- Primary goal in this PR was to upgrade Java version to 17
- Some libs required an update as well: upgraded them to recent/stable versions since minimum versions needed for java 17 are difficult to identify (transitive dependencies) and may have unresolved vulnerabilities
- Bumped spring to minimum 3.x stable version, which is recommended to resolve some of the dependencies / vulnerabilities
- Gradle 7.5+ is needed for spring 3, 8.x has better memory management for the daemons. hence bumped gradle from 7.4.1 to 8.7
- Moved some of the duplicated gradle config to support easy maintenance
- Added library versions to root build file to ensure same version is used across subprojects (can be overridden if needed in sub projects)
- Retain client in java 11 (to be handled separately)
- Split the bump-publish-tag workflow job into separate client-publish (java-11) and service-image (java-17) in the same workflow
- Formatting changes are due to upgraded version of spotless
- SQL file changes are only
- resource renamed to resourceObject in openApiSchema: this avoids conflict with openapi keyword (which causes errors in v3)

---
Large change set due to 
- Lockfile updates
- javax -> jakarta namespace migration (spring 2->3)
- sql files regenerated (format changes only)

---
Downstream regression test successful - https://github.com/verily-src/verily-tanagra-backend/actions/runs/9537859915/job/26286400753?pr=396


annotationProcessor/build.gradle
```diff
@@ -39,7 +39,7 @@ dependencyManagement {
         mavenBom org.springframework.boot.gradle.plugin.SpringBootPlugin.BOM_COORDINATES
     }
     dependencies {
-        dependency group: "io.swagger.core.v3", name: "swagger-annotations", version: "2.1.5"
+        dependency group: 'io.swagger.core.v3', name: 'swagger-annotations', version: '2.2.21'
```
dexamundsen, line 42: Only libs upgrades are those required for Java 17 compatibility

indexer/src/main/java/bio/terra/tanagra/indexing/job/bigquery/WriteEntityLevelDisplayHints.java
```diff
@@ -234,13 +233,10 @@ private static boolean isRangeHint(Attribute attribute) {
...
       return false;
     }
-    switch (attribute.getRuntimeDataType()) {
-      case DOUBLE:
-      case INT64:
-        return true;
-      default:
-        return false;
-    }
+    return switch (attribute.getRuntimeDataType()) {
```
dexamundsen, line 236: Enhanced switch statement available from Java 13

service/src/main/java/bio/terra/tanagra/app/controller/ActivityLogApiController.java
```diff
@@ -80,17 +79,16 @@ private ApiActivityLogEntry toApiObject(ActivityLog activityLog) {
...
         .additionalInfo(
             new ApiActivityLogEntryAdditionalInfo()
                 .exportModel(activityLog.getExportModel())
                 .recordsCount(activityLog.getRecordsCount()));
   }
 
-  private ApiResource toApiObject(ActivityLogResource activityLogResource) {
-    ApiResource apiResource =
-        new ApiResource()
+  private ApiResourceObject toApiObject(ActivityLogResource activityLogResource) {
```
dexamundsen, line 89: resource renamed to resourceObject in openApiSchema

service/src/main/java/bio/terra/tanagra/app/authentication/AuthInterceptor.java
```diff
@@ -61,7 +61,7 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
     }
 
     HandlerMethod method = (HandlerMethod) handler;
-    ApiOperation apiOp = AnnotationUtils.findAnnotation(method.getMethod(), ApiOperation.class);
+    Operation apiOp = AnnotationUtils.findAnnotation(method.getMethod(), Operation.class);
```
dexamundsen, line 64: ApiOperation moved to Operation in swagger v3

service/src/main/java/bio/terra/tanagra/app/controller/ExportApiController.java
```diff
@@ -200,8 +200,7 @@ public ResponseEntity<ApiEntityOutputPreviewList> describeExport(
     Underlay underlay = underlayService.getUnderlay(underlayName);
     Map<EntityOutputPreview, String> indexSqlForEntityOutputs = new HashMap<>();
     Map<EntityOutputPreview, String> sourceSqlForEntityOutputs = new HashMap<>();
-    entityOutputPreviews
-        .parallelStream()
+    entityOutputPreviews.parallelStream()
```
dexamundsen, line 203: Formatting changes are due to upgrades spotless

service/src/main/java/bio/terra/tanagra/db/AnnotationDao.java
```diff
@@ -16,7 +16,7 @@
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
-import javax.annotation.Nullable;
+import jakarta.annotation.Nullable;
```
dexamundsen, line 19: javax -> jakarta in spring 3.x

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.5-bin.zip
```
dexamundsen, line 3: spring 3 needs gradle 7.5+

cli/src/main/java/bio/terra/tanagra/cli/utils/Logger.java
```diff
@@ -13,7 +13,7 @@
 import ch.qos.logback.core.rolling.RollingFileAppender;
 import ch.qos.logback.core.rolling.SizeBasedTriggeringPolicy;
 import ch.qos.logback.core.util.FileSize;
-import ch.qos.logback.core.util.StatusPrinter;
+import ch.qos.logback.core.util.StatusPrinter2;
```
dexamundsen, line 16: replaces deprecated StatusPrinter

underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQQueryRunner.java
```diff
@@ -592,4 +593,10 @@ public ExportQueryResult run(ExportQueryRequest exportQueryRequest) {
     return new ExportQueryResult(
         exportFileUrlAndFileName.getRight(), exportFileUrlAndFileName.getLeft());
   }
+
+  // Java17 returns Instant with nanoseconds, BigQuery only supports upto microseconds
+  // Hence, use microseconds precision everywhere
+  public static Instant instantNowMicros() {
```
dexamundsen, line 599: Did not find a shared 'util' class and hence put this function here. It is needed in both underlay and service (see comment)

service/src/main/java/bio/terra/tanagra/db/AnnotationDao.java
```diff
@@ -9,14 +9,14 @@
 import bio.terra.tanagra.exception.SystemException;
 import bio.terra.tanagra.service.artifact.model.AnnotationKey;
 import bio.terra.tanagra.service.artifact.model.AnnotationValue;
+import jakarta.annotation.Nullable;
```
dexamundsen, line 12: javax -> jakarta in spring 3.x

underlay/src/test/resources/sql/BQFilterTest/temporalPrimaryFilterMultipleOutputLastAnyDaysAfter.sql
```diff
@@ -4,93 +4,71 @@
     FROM
         ${ENT_person}      
     WHERE
-        id IN (
```
dexamundsen, line 7: Regenerated sql files

.github/workflows/bump-tag-publish.yaml
```diff
@@ -88,12 +88,11 @@ jobs:
           VERSION_LINE_MATCH: "^gradle.ext.tanagraVersion\\s*=\\s*\".*\""
           VERSION_SUFFIX: SNAPSHOT
 
-      - name: Set up AdoptOpenJDK
-        if: steps.skiptest.outputs.is-bump == 'no'
```
marikomedlock, line 92: What is the reason for removing this condition? This is to avoid an endless GHA chain where each version bump triggers a new version bump -- even if the other steps are skipped, no need to setup a JDK if we won't use it.

buildSrc/src/main/groovy/tanagra.java-conventions.gradle
```diff
@@ -21,29 +21,28 @@ plugins {
...
     google()
 }
 
-sourceCompatibility = JavaVersion.VERSION_11
+sourceCompatibility = JavaVersion.VERSION_17
 
 // See https://docs.gradle.org/current/userguide/dependency_locking.html
 dependencyLocking {
     lockAllConfigurations()
-    lockMode = LockMode.STRICT
```
marikomedlock, line 37: Why the change from strict > lenient?

buildSrc/src/main/groovy/tanagra.test-conventions.gradle
```diff
@@ -31,3 +29,9 @@ def jacocoTestReport = tasks.named('jacocoTestReport') {
 tasks.named('check').configure {
     dependsOn jacocoTestReport
 }
+
+tasks.register('noCloudAccessRequiredTests', Test) {
```
marikomedlock, line 33: Thanks for moving this to a central place. Just checking, this doesn't affect our ability to run all tagged tests with `./gradlew noCloudAccessRequiredTests` right?

client/build.gradle
```diff
@@ -1,30 +1,32 @@
...
 
-group = "bio.terra"
+group = 'bio.terra'
 version = gradle.tanagraVersion
 
-//// Keep the client library at Java 11 for now
-sourceCompatibility = JavaVersion.VERSION_11
-targetCompatibility = JavaVersion.VERSION_11
+sourceCompatibility = JavaVersion.VERSION_17
+targetCompatibility = JavaVersion.VERSION_17
```
marikomedlock, line 29: What are the implications for services that depend on our client library? Does this mean they also need to be on Java 17? I'm not certain, but we may have both a Java 11 (VUMC admin service) and Java 17 (AoU core workbench service) that depend on this client lib.

service/build.gradle
```diff
@@ -50,74 +51,88 @@ repositories {
...
-    implementation group: "org.springframework.security", name: "spring-security-core"
-
-    // Spring version controlled JUnit dependencies.
+    annotationProcessor 'org.springframework.boot:spring-boot-configuration-processor'
+    implementation 'org.springframework.boot:spring-boot-starter-data-jdbc'
+    implementation 'org.springframework.boot:spring-boot-starter-web'
+    implementation 'org.springframework.security:spring-security-core'
+    implementation 'jakarta.validation:jakarta.validation-api'
+    implementation 'jakarta.servlet:jakarta.servlet-api:6.0.0'
+    implementation 'jakarta.ws.rs:jakarta.ws.rs-api:4.0.0'
```
marikomedlock, line 95: Comment says Spring controlled dependencies -- should these have a version specified inline?

service/build.gradle
```diff
@@ -50,74 +51,88 @@ repositories {
...
+
+    // [Application DB] MySQL/MariaDB
+    implementation 'org.mariadb.jdbc:mariadb-java-client:3.4.0'
+    implementation 'com.mysql:mysql-connector-j:8.4.0'
+    implementation 'com.google.cloud.sql:mysql-socket-factory-connector-j-8:1.18.1'
+
+    // GCP libraries. versions are controlled by the BOM specified in buildSrc.
```
marikomedlock, line 102: nit: "Most versions" because the oauth2 library is not?

service/gradle/jib.gradle
```diff
@@ -1,8 +1,9 @@
 jib {
     from {
-        image = "adoptopenjdk:11-jre-hotspot"
+        // Google's distroless images are openjdk, this is the simplest with adoptopenjdk
+        image = 'us.gcr.io/broad-dsp-gcr-public/base/jre:17-debian'
```
marikomedlock, line 4: Curious about the Broad dependency here -- is there a public image that could be better to use instead?

service/src/main/java/bio/terra/tanagra/service/accesscontrol/model/impl/AouWorkbenchAccessControl.java
```diff
@@ -133,10 +133,12 @@ public Set<Action> getStudyPermissions() {
         return null;
       }
     }
+
+    HttpStatus statusCode = (HttpStatus) responseEntity.getStatusCode();
     LOGGER.error(
         "AoU Workspace Access Level api error. Http status code: {}, Message: {}",
-        responseEntity.getStatusCodeValue(),
-        responseEntity.getStatusCode().getReasonPhrase());
+        statusCode,
```
marikomedlock, line 140: This looks like it's logging a different property -- is the change getStatusCodeValue > getStatusCode intentional?

underlay/build.gradle
```diff
@@ -1,50 +1,77 @@
...
-    implementation group: "com.google.cloud", name: "google-cloud-storage"
-    implementation group: "com.google.guava", name: "guava"
+    implementation "com.fasterxml.jackson.core:jackson-core:${gradle.vJackson}"
+    implementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${gradle.vJackson}"
+
+    implementation "org.apache.commons:commons-text:${gradle.vApacheCommonsText}"
+    testFixturesImplementation "org.slf4j:slf4j-api:${gradle.vSlf4jApi}"
+
+    // Spring version controlled dependencies
+    implementation 'org.springframework.retry:spring-retry:2.0.6'
```
marikomedlock, line 50: Comment says Spring controlled version -- should there be a version number here?

underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQQueryRunner.java
marikomedlock: I'm not sure we need to replace `Instant.now()` everywhere with this. In the service, I think we only use Instant in filenames, so this shouldn't matter. And since this is a BQ limitation, is it possible to do this truncation only at the point where we pass it to BQ, rather than farther up the stack?

Maybe a good candidate to discuss over a quick GVC?

marikomedlock (commented):
Thanks for all the Gradle file cleanup! Did you use any automated tooling to do that? If so, it would be great to enable that as part of our PR tests (not for this PR, just for a future ticket).

Haven't made it through all the SQL files yet, but wanted to send along what I have so far.

buildSrc/src/main/groovy/tanagra.test-conventions.gradle
dexamundsen: This file is included as a plugin into all other subprojects - verified that `./gradlew noCloudAccessRequiredTests` works as usual.

.github/workflows/bump-tag-publish.yaml
dexamundsen: oops. this was a copy-paste error. added back

buildSrc/src/main/groovy/tanagra.java-conventions.gradle
dexamundsen: This was due to spotless config errors  - with a strict mode every build command leads to below error 

```
Could not determine the dependencies of task ':underlay:spotlessCheck'.
> Could not create task ':underlay:spotlessGroovyGradleCheck'.
   > Could not create task ':underlay:spotlessGroovyGradle'.
      > Could not resolve all dependencies for configuration ':underlay:spotless-1757186549'.
         > Locking strict mode: Configuration ':underlay:spotless-1757186549' is locked but does not have lock state.
```

to fix it either the lock Mode must be DEFAULT or the '--write-locks ' option must be specified every time
```
 >> ./gradlew underlay:build --write-locks 
```

I looked into this a bit but could not figured the root cause, hence removed the STRICT mode, keeping it consistent with other Verily repo configs downstream

client/build.gradle
dexamundsen: From my understanding - yes, the service should ALSO switch to java17 to use clients built with targetCompatibility 17. 
A java 17 service can use either java 11 or java 17 (example: this repo uses java 11 admin client) as long as the interfaces are compatible. 
However targetCompatibility 17 will require VUMC to switch other services to java 17 (to consume changes committed to this client after java 17 upgrade). to prevent this dependency I will retain the targetCompatibility 11 and create a followup ticket to change it. ([BENCH-3210](https://verily.atlassian.net/browse/BENCH-3210))

more info - https://stefma.medium.com/sourcecompatibility-targetcompatibility-and-jvm-toolchains-in-gradle-explained-d2c17c8cff7c

service/gradle/jib.gradle
dexamundsen: I copied this from Verily repos since they were already vetted by security (atleast broad). 
here is a list of official images - https://github.com/docker-library/docs/blob/master/eclipse-temurin/README.md#supported-tags-and-respective-dockerfile-links
Is there a recommended one for Tanagra

service/build.gradle
dexamundsen: Split them into separate blocks since these need specific versions. Same in other build files

service/build.gradle
dexamundsen: This not controlled in the bom and version needs to be specified, moved to separate block

underlay/build.gradle
dexamundsen: fixed, see other related comment

service/src/main/java/bio/terra/tanagra/service/accesscontrol/model/impl/AouWorkbenchAccessControl.java
dexamundsen: getStatusCodeValue() is deprecated in favor of getStatusCode(). Both return the same effectiv value
getStatusCodeValue --> int value of the enum, which is converted to string in Logger.error. 
getStatusCode -> httpStatus enum, which is converted to string in Logger.error. 

IMO enum to string is cleaner than int to string since the former leverages the native tostring implementation (if any) of the enum.

underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQQueryRunner.java
dexamundsen: Set up a call for monday. I debated where this needs to be put. 
- adding it only at BQ calls: this is the minimum need. Created a util function so that we dont need to maintain / change it in multiple locations when needed
- adding in in filenames / sql etc: agree, not needed per say. However using it everywhere will avoid any breaking changes wrt sql. As of filenames, no impact in functionality

dexamundsen:
> Thanks for all the Gradle file cleanup! Did you use any automated tooling to do that? If so, it would be great to enable that as part of our PR tests (not for this PR, just for a future ticket).

:( manual. however there are tools to automate some of it. sending you details via slack

client/build.gradle
dexamundsen: Moving only target to java 11 did not work as expected. Hence reverted client changes and will be handled separately ([BENCH-3210](https://verily.atlassian.net/browse/BENCH-3210), draft changes https://github.com/DataBiosphere/tanagra/pull/875)
tested local publish

```
tanagra (dexamundsen/build-17) >> ./gradlew client:publishToMavenLocal
...
BUILD SUCCESSFUL in 975ms
16 actionable tasks: 3 executed, 13 up-to-date
tanagra (dexamundsen/build-17) >> ls -l ~/.m2/repository/bio/terra/tanagra-client/
total 8
drwxr-xr-x  7 dexamundsen  primarygroup  224 Jun  8 19:27 0.0.512-SNAPSHOT-LOCAL
-rw-r--r--  1 dexamundsen  primarygroup  381 Jun  8 19:27 maven-metadata-local.xml
```

buildSrc/src/main/groovy/tanagra.java-conventions.gradle
marikomedlock: Is the configuration name randomly generated? i.e. does 1757186549 change or is it consistent. On the off chance it's consistent, you might try just writing locks for that task. If not, removing strict sgtm.

service/build.gradle
```diff
@@ -26,17 +27,17 @@ ext {
...
     }
     maven {
-        url "https://packages.confluent.io/maven/"
+        url 'https://packages.confluent.io/maven/'
     }
     maven {
-        name "github"
-        url "https://maven.pkg.github.com/vanderbilt/tanagra-client-libraries"
+        name 'github'
+        url'https://maven.pkg.github.com/vanderbilt/tanagra-client-libraries'
```
marikomedlock, line 40: Missing space here? `url '`

underlay/build.gradle
```diff
@@ -1,50 +1,78 @@
...
-        url "https://packages.confluent.io/maven/"
+        url 'https://packages.confluent.io/maven/'
+    }
+}
+
+dependencyManagement {
+    imports {
+        mavenBom org.springframework.boot.gradle.plugin.SpringBootPlugin.BOM_COORDINATES
+        mavenBom 'io.opentelemetry:opentelemetry-bom:1.38.0'
+        mavenBom 'io.opentelemetry.instrumentation:opentelemetry-instrumentation-bom-alpha:2.4.0-alpha'
```
marikomedlock, line 31: Are we using open telemetry now?

underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQQueryRunner.java
marikomedlock: OK, yeah let's talk it out. My thinking is that since this is a BQ limitation, we should only force it when passing it to BQ. In the future, we may support other cloud backends (e.g. ElasticSearch) which may have their own limitations (e.g. no micros, only full seconds) and so we don't want to tie our codebase too closely to BQ.

underlay/build.gradle
dexamundsen: No, was required since it is included in terra-common-lib

buildSrc/src/main/groovy/tanagra.java-conventions.gradle
dexamundsen: Manually updating the lock file is not recommended. As discussed leaving it with default lock mode (not strict)

underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQQueryRunner.java
dexamundsen: From discussion moving this to BQ specific code only.

service/build.gradle
dexamundsen: fixed

underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQQueryRunner.java
dexamundsen: Tested (in verily-test) that there is no impact to SQL - 
deployed java 11 - created study, cohort, feature
deployed java 17 - created study, cohort, feature; Updated java11 study (created cohort)
deployed java 11 - Updated java17 study (created feature)

underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQQueryRunner.java
```diff
@@ -53,7 +54,7 @@ public ListQueryResult run(ListQueryRequest listQueryRequest) {
     // Build the SQL query.
     Instant queryInstant =
         listQueryRequest.getPageMarker() == null
-            ? Instant.now()
+            ? instantNowMicros()
```
marikomedlock, line 57: I think it's better to truncate or round to microseconds at the point we pass the timestamp value to BQ. Doing it here isn't really confined to the BQ code, because this `queryInstant` is part of the return value of this function as well. Can you truncate in `BQExecutor.toQueryParameterValue` instead? That would also guarantee that we truncate in any instances where the timestamp is turned into a literal. I will push the example code to a branch.

underlay/build.gradle
```diff
@@ -1,50 +1,78 @@
 plugins {
     id 'java'
     id 'java-test-fixtures'
+    id 'pmd'
+
     id 'tanagra.java-conventions'
+    id 'org.springframework.boot' version '3.2.5'
+    id 'io.spring.dependency-management' version '1.1.5'
```
marikomedlock, line 8: Didn't notice this the first time around, but why do we need a Spring dependency in the `underlay` sub-project? I think we should only need this in the `service` and perhaps `client` sub-projects. If this is required because of the dependency on `terra-common-lib`, and we can't make it work without it, then maybe we should look into removing that dependency. I think (not certain, but will check) that we're only using that library for some common exceptions.

marikomedlock (commented):
I tried running the service locally on my laptop and am seeing a lot of the below exception. The tests pass for me, so that's good, but I'd like to understand what the issue here is. Is this related to the open telemetry dependency added in the `underlay/build.gradle` file?

```
i.o.exporter.internal.http.HttpExporter  : Failed to export logs. The request could not be executed. Full error message: Failed to connect to localhost/[0:0:0:0:0:0:0:1]:4318```

File: underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQQueryRunner.java (hunk 23 above)
Comment (dexamundsen) [reply underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQQueryRunner.java]: Closing this since the change is merge.

**Comment by dexamundsen** [comment]
> I tried running the service locally on my laptop and am seeing a lot of the below exception. The tests pass for me, so that's good, but I'd like to understand what the issue here is. Is this related to the open telemetry dependency added in the `underlay/build.gradle` file?
> 
> ```
> i.o.exporter.internal.http.HttpExporter  : Failed to export logs. The request could not be executed. Full error message: Failed to connect to localhost/[0:0:0:0:0:0:0:1]:4318```
> ```

info on export log exception - https://github.com/open-telemetry/opentelemetry-java-instrumentation/discussions/9861
this dependency is added to service (could not find a way to remove it)

File: .github/workflows/underlay-test.yaml [hunk 25]
```diff
@@ -16,15 +16,15 @@ on:
       - 'docs/**'
   workflow_dispatch:
 jobs:
-  test:
+  underlay-test:
```
marikomedlock, line 19: I'm not certain, but I think these job names may be involved in the GH branch protection rule on the `main` branch. If so, you may get blocked from merging until we update the rule.

.github/workflowsForDownstreamRepo/regression-test-downstream-repo.yaml
```diff
@@ -12,7 +12,7 @@ env:
   DEFAULT_UNDERLAYS: cmssynpuf,aouSR2019q4r4
   TANAGRA_UNDERLAY_FILES: cmssynpuf_verily,aouSR2019q4r4_verily
 jobs:
-  test:
+  regression-test:
```
marikomedlock, line 15: If you do change the job name here, please also change it downstream in `verily-src/verily-tanagra-backend` repo, so we can keep those two consistent.

service/build.gradle
```diff
@@ -130,55 +144,50 @@ protobuf {
...
     }
 }
 
-task regressionTests(type: Test) {
+tasks.register('regressionTests', Test) {
     useJUnitPlatform {
-        includeTags "regression-test"
+        includeTags 'regression-test'
     }
 }
```
marikomedlock, line 161: The regression test tasks doesn't seem to be running any tests. In the GHA run for this PR, I see:
```
The tag 'regression-test' is both included and excluded.  This will result in the tag being excluded, which may not be what was intended.  Please either include or exclude the tag but not both.
```

.github/workflows/bump-tag-publish.yaml
```diff
@@ -44,91 +44,130 @@ on:
...
-        with:
-          path: |
-            ~/.gradle/caches
-            ~/.gradle/wrapper
-          key: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle') }}
-          restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
-      - name: Grant execute permission for gradlew
-        if: steps.skiptest.outputs.is-bump == 'no'
-        run: chmod +x gradlew
+  client-publish:
```
marikomedlock, line 99: Have you been able to run this updated GHA from your branch? Then we can point the SD/AoU teams at a client library that they can test out pre or post merge against our `test` env (running Java 17) if they'd like.

In the past, when working on this action, I think I updated the main branch name hardcoded here from main -> [my branch e.g. java-17] and then opened a PR against my branch and merged it, to check that everything got published/bumped/build/etc correctly.

.github/workflows/underlay-test.yaml
dexamundsen: The status checks are tied to the entire workflow and not the individual jobs. Hence this change will not have any impact on PR checks.
Reason for change - this displays convenient name on the logs screen. 

also tested - in verily-src/terra-cost-billing-service

- verified that branch protection settings have status checks needed
- Create PR which changes job name (https://github.com/verily-src/terra-cost-billing-service/pull/27). Merged the PR. The required status changes are as expected. 
- Create another PR with minor readme changes (https://github.com/verily-src/terra-cost-billing-service/pull/28). Status changes are as expected.

.github/workflowsForDownstreamRepo/regression-test-downstream-repo.yaml
dexamundsen: Done, https://github.com/verily-src/verily-tanagra-backend/pull/396

service/build.gradle
dexamundsen: (yikes). fixed

.github/workflows/underlay-test.yaml
dexamundsen: There may be two stale checks, which i suspect will be fixed when this is merged. 
in the case this is blocker, I can revert the changes

dexamundsen: see https://github.com/DataBiosphere/tanagra/pull/887 -- once the workflows are updated in main, there should be no issues

dexamundsen: reverted

.github/workflows/bump-tag-publish.yaml
dexamundsen: @marikomedlock Tim checked with VUMC - they do not have concerns upgrading the client. Targeting to merge this when you are back from vacation

dexamundsen: also can confirm that client compile succeeds but I cannot test create via the GH workflow. 
local client published and attached here for vumc testing if needed. 
[tanagra-client.zip](https://github.com/user-attachments/files/16022385/tanagra-client.zip)

cc: @tjennison-google

tjennison-google:
> Failed to connect to localhost/[0:0:0:0:0:0:0:1]:4318

Did the spurious exceptions get resolved? If not, maybe "OTEL_JAVAAGENT_ENABLED=false" (from https://opentelemetry.io/docs/zero-code/java/agent/configuration/) would prevent them?

dexamundsen:
> > Failed to connect to localhost/[0:0:0:0:0:0:0:1]:4318
> 
> Did the spurious exceptions get resolved? If not, maybe "OTEL_JAVAAGENT_ENABLED=false" (from https://opentelemetry.io/docs/zero-code/java/agent/configuration/) would prevent them?

Yes, this is resolved. 
To avoid extra configuration (like the env vars) that are otherwise not needed, I excluding the transitive dependency since it was not used

.github/workflows/bump-tag-publish.yaml
marikomedlock: Client compile succeeding is great, but we should test that the publish to Artifactory works.

Can you publish to Artifactory locally? What problem did you hit testing the GHA?

service/build.gradle
```diff
@@ -142,43 +153,40 @@ test {
...
+        systemProperty('REGRESSION_TEST_DIRS', project.findProperty('regressionTestDirs'))
     }
-    if (project.findProperty("regressionTestUnderlays")) {
-        systemProperty("REGRESSION_TEST_UNDERLAYS", project.findProperty("regressionTestUnderlays"))
+    if (project.findProperty('regressionTestUnderlays')) {
+        systemProperty('REGRESSION_TEST_UNDERLAYS', project.findProperty('regressionTestUnderlays'))
     }
-    if (project.findProperty("regressionTestFiles")) {
-        systemProperty("REGRESSION_TEST_FILES", project.findProperty("regressionTestFiles"))
+    if (project.findProperty('regressionTestFile')) {
```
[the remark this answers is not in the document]
dexamundsen: in error, it was only meant to change quotes. fixed

dexamundsen:
@marikomedlock - re: client publish comment (https://github.com/DataBiosphere/tanagra/pull/829#discussion_r1661207878). 
see https://github.com/DataBiosphere/tanagra/pull/829#discussion_r1657942736 . Local publish succeeds, GH workflow fails due to permissions / does not allow publish from non-main branch
Current changes only update Java version. Other changes are no longer needed since vumc does not need older client (see Tim's message on slack?)
-- tagging in a new comment since GH isn't letting me reply to other one.

dexamundsen:
Fixed and tested client release - 
Client - https://broadinstitute.jfrog.io/ui/repos/tree/Properties/libs-snapshot-local/tanagra/tanagra-client/0.0.522-SNAPSHOT
release - https://github.com/DataBiosphere/tanagra/releases/tag/0.0.522
Log - https://github.com/DataBiosphere/tanagra/actions/runs/9748655284/job/26904091974#step:6:630

marikomedlock:
> Fixed and tested client release - Client - https://broadinstitute.jfrog.io/ui/repos/tree/Properties/libs-snapshot-local/tanagra/tanagra-client/0.0.522-SNAPSHOT release - https://github.com/DataBiosphere/tanagra/releases/tag/0.0.522 Log - https://github.com/DataBiosphere/tanagra/actions/runs/9748655284/job/26904091974#step:6:630

Great to see the publish succeeded, but looks like it's going to the wrong path in Artifactory `libs-snapshot-local/tanagra/tanagra-client` instead of `libs-snapshot-local/bio/terra/tanagra-client`. Per conversation at standup, maybe need to set an artifact group property for the updated version of the Artifactory publish Gradle plugin.

dexamundsen:
> > Fixed and tested client release - Client - https://broadinstitute.jfrog.io/ui/repos/tree/Properties/libs-snapshot-local/tanagra/tanagra-client/0.0.522-SNAPSHOT release - https://github.com/DataBiosphere/tanagra/releases/tag/0.0.522 Log - https://github.com/DataBiosphere/tanagra/actions/runs/9748655284/job/26904091974#step:6:630
> 
> Great to see the publish succeeded, but looks like it's going to the wrong path in Artifactory `libs-snapshot-local/tanagra/tanagra-client` instead of `libs-snapshot-local/bio/terra/tanagra-client`. Per conversation at standup, maybe need to set an artifact group property for the updated version of the Artifactory publish Gradle plugin.

Fixed by explicitly adding the group as well - https://broadinstitute.jfrog.io/ui/repos/tree/General/libs-snapshot-local/bio/terra/tanagra-client/0.0.523-SNAPSHOT / https://github.com/DataBiosphere/tanagra/actions/runs/9750085844/job/26908679015

marikomedlock (approved):
Thank you so much for pushing through all these changes!

PR #830, opened 2024-04-30
Update AoU drug hierarchy configuration.

marikomedlock:
Updated the AoU `ingredient` entity definition (`all.sql`, `childParent.sql`) to match the hierarchy built for AoU-CB.
- Built the relationship between the root `ATC-1st` nodes and the ingredient `RxNorm` nodes using the `concept_relationship` table, as diagrammed below.
- Built the relationship between the ingredient `RxNorm` nodes and the dosage nodes using the `concept_ancestor` table (see `childParent.sql` last `SELECT` statement in the `UNION`).

Screenshot 2024-04-25 at 11 29 54 AM

This PR was a joint effort with @freemabd.


freemabd (approved):
Thanks

PR #833, opened 2024-04-30
Don't retry BigQuery 403 total rows too large.

marikomedlock:
Previously, we retried all `403 Forbidden` BigQuery errors.


underlay/src/main/java/bio/terra/tanagra/utils/GoogleBigQuery.java
```diff
@@ -437,17 +437,23 @@ static boolean isRetryable(Exception ex) {
...
     }
-    int statusCode = ((BigQueryException) ex).getCode();
+    BigQueryException bqEx = (BigQueryException) ex;
+    int statusCode = bqEx.getCode();
     LOGGER.error("Caught a BQ error (status code = {}).", statusCode, ex);
 
+    if (statusCode == HttpStatus.SC_FORBIDDEN) {
+      return !bqEx.getMessage()
+          .contains(
+              "Response too large to return. Consider specifying a destination table in your job configuration.");
```
tjennison-google, line 447: Is there not a way to check this other than by text?

tjennison-google (approved):

underlay/src/main/java/bio/terra/tanagra/utils/GoogleBigQuery.java
marikomedlock: Found a custom property on the BQ excpetion I can check instead.

PR #834, opened 2024-04-30
Handle boolean data types in attribute plugin

dolbeew:
Prevent the rendering of the attribute slider and just return `true` as the value for boolean data types. Deployed this branch locally with the workbench and counts match the current cohort builder in the AoU test env.
Screenshot 2024-04-30 at 12 34 31 PM


ui/src/criteria/attribute.tsx
```diff
@@ -231,6 +232,21 @@ function AttributeInline(props: AttributeInlineProps) {
...
+          data.selected = [
+            {
+              name: "",
+              value: true,
+            },
+          ];
+        })
+      );
+    }
+  }, []);
```
dolbeew, line 248: Not sure if this is the best way to set the value but didn't have access to the `dataType` of the plugin in the `generateFilter` function.

Set `name` as an empty string so it doesn't get rendered along with the title.

tjennison-google: It is available, but the linter complains if the parameters are unused. You can just add them back. See `cohort.ts:159` and `criteria/classification.tsx:179`. This cause the criteria to to be changed and resaved.

dolbeew: Nice, good to know. Moved this to `generateFilter`

.gitignore
```diff
@@ -9,6 +9,7 @@ workflow/rendered/
 */bin
 tanagra-indexing.html
 diffReleases.txt
+ui/src/tanagra-underlay/
```
tjennison-google, line 12: Move this to `ui/.gitignore`.

tjennison-google (approved):

.gitignore
dolbeew: Still failed. Just going to fix the lint errors.

ui/src/tanagra-underlay/underlayConfig.ts
```diff
@@ -26,7 +26,7 @@ export enum SZCorePlugin {
   OUTPUT_UNFILTERED = "OUTPUT_UNFILTERED",
   TEXT_SEARCH = "TEXT_SEARCH",
   UNHINTED_VALUE = "UNHINTED_VALUE",
-};
+}
```
tjennison-google, line 29: Sorry, I was telling you to put the fix in the wrong place. It should go in `ui/.eslintignore`. Putting it here won't work because this is generated.

dolbeew: Is this generated? Looks like Mariko made changes here last week: https://github.com/DataBiosphere/tanagra/pull/820/files#diff-bf1a95422b29fc148e786e4b143abb98804528f2181280ab2f825a476fc46ea3

tjennison-google: This file is generated as part of the backend build which is why it's checked in so it doesn't need to be run separately before working on the UI: `./gradlew annotationProcessor:run -Pgenerator=UNDERLAY_CONFIG_TYPESCRIPT`. This is different from the the the UI specific stuff generated by `npm run codegen`, which is what would have made sense to go in .gitignore.

dolbeew: Ah, makes sense. Thanks

PR #835, opened 2024-04-30
Add visualization plugin framework

tjennison-google:
This is pluggable framework for visualizations that currently just replaces the demographics charts. This gives us something to hook up the new queries and configuration to as we build it out.

Also fixes an issue that was causing empty cohorts to unnecessarily refresh the demographic charts. A new random id was being generated every time an empty cohort was fetched causing the UI to constantly think it had changed.

After experiments with other charting libraries didn't pan out, this largely leaves the existing demographic chart code intact but moved into a plugin that's instantiated for each chart.

### DIFF
--- a/ui/src/app.tsx
+++ b/ui/src/app.tsx
@@ -4,6 +4,7 @@ import { enableMapSet } from "immer";
 import "plugins";
 import { RouterProvider } from "react-router-dom";
 import { createAppRouter } from "router";
+import "viz/plugins";
 import theme from "./theme";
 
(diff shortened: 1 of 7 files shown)


marikomedlock (approved):
Does this change the appearance and/or behavior (e.g. tooltips) of the existing charts at all?

tjennison-google:
> Does this change the appearance and/or behavior (e.g. tooltips) of the existing charts at all?

Only some minor details. Updated the PR description to match the current behavior.

PR #836, opened 2024-05-01
Minor cleanup, address warnings

dexamundsen:
No functional or breaking changes. Only containing minor housekeeping changes (address warnings)

- Docs: grammar fixes
- Gradle config: checkstyle, pmd and spotbugs config used in subsequent PR
- Build files: quotes only
- OpenApi: remove disallowed fields
- GitHub workflows: Use V4 actions
- Shell scripts: Use quotes for variables (Best practices)
- cli: address warnings, minor optimizations
- indexer: address warnings, minor optimizations


.github/tools/diffReleases.sh
```diff
@@ -4,7 +4,7 @@ first_release=$1
 last_release=$2
 output_file="diffReleases.txt"
 
-if [ -z $first_release ] || [ -z $last_release ];
+if [ -z "${first_release}" ] || [ -z "${last_release}" ];
```
dexamundsen, line 7: Using quotes prevent word splitting

.github/workflows/bump-tag-publish.yaml
```diff
@@ -65,7 +65,7 @@ jobs:
           echo semver-part=$SEMVER_PART >> $GITHUB_OUTPUT
           echo checkout-branch=$CHECKOUT_BRANCH >> $GITHUB_OUTPUT
       - name: Checkout current code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
```
dexamundsen, line 68: v4 actions are needed to address GH action deprecation warnings for nodejs

annotationProcessor/build.gradle
```diff
@@ -1,8 +1,8 @@
 plugins {
-    id "java"
-    id "tanagra.java-conventions"
-    id "io.spring.dependency-management" version "1.0.11.RELEASE"
-    id "org.springframework.boot" version "2.5.1"
+    id 'java'
```
dexamundsen, line 2: Using single quotes in gradle files for strings where no variable substitution is expected

.github/workflows/bump-tag-publish.yaml
tjennison-google: Thanks. I had noticed these warnings but hadn't gotten around to fixing them yet.

tjennison-google (commented):
I'll let Mariko approve since there are a lot more backend changes.

service/gradle/open-api.gradle
```diff
@@ -1,5 +1,5 @@
-
 // OpenAPI/Swagger Server Generation
+def openapiOutputDir = "${buildDir}/openapi"
```
marikomedlock, line 2: Is this defined in the main `service/build.gradle` file already?

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2082,9 +2075,7 @@ components:
         id:
           $ref: "#/components/schemas/CohortId"
         revisionId:
-          description: ID of the cohort revision, if frozen, null otherwise
```
marikomedlock, line 2085: This seems useful to me, why remove?

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2787,31 +2776,22 @@ components:
       properties:
         type:
           $ref: "#/components/schemas/ResourceType"
-          nullable: false
         studyId:
           $ref: "#/components/schemas/StudyId"
-          nullable: false
         studyDisplayName:
           $ref: "#/components/schemas/StudyDisplayName"
-          nullable: true
```
marikomedlock, line 2796: What is the default for `nullable` -- is it different for required/not properties?

.github/workflowsForDownstreamRepo/regression-test-downstream-repo.yaml
marikomedlock, line 1: This GHA workflow file is used in the downstream `verily-src/tanagra-backend` repo also, as a validation that it works. Would you mind also updating there to keep these in sync please?

marikomedlock (commented):
Thanks for all this cleanup! DId you use any tooling that we could add to our repo to do some of this automatically?

service/gradle/open-api.gradle
dexamundsen: right, had meant to remove the one in service/build.gradle since it was only used in this file. done that now

service/gradle/open-api.gradle
```diff
@@ -10,16 +10,18 @@ swaggerSources {
...
-                "--model-name-prefix",
-                "Api",
-                "-D",
-                "interfaceOnly=true," +
-                "useTags=true," +
-                "dateLibrary=java8"
+                '--model-name-prefix',
+                'Api',
+                '--additional-properties',
+                'errorOnUnknownEnum=true',
```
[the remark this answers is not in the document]
dexamundsen: Yes, copied the values from VWB WSM. 
if errorOnUnknownEnum=true, then the swagger code generation will fail if a enum property field is assigned a string value that is not a known enum

service/src/main/resources/api/service_openapi.yaml
dexamundsen: description field is allowed only for fields (properties) are are defined in place, not for references. 

in this case - revisionId: is reference to another custom schema type ($ref: "#/components/schemas/CohortRevisionId"). The definition of CohortRevisionId internally specifies the type of object and the description.

 in other words, removed it since it was a duplicate. The Swagger UI will still have the description from underlying referenced objects. 
 I can retain them if needed. let me know if so.

service/src/main/resources/api/service_openapi.yaml
dexamundsen: Default value of nullable = true. Same for required / not required. 
Reason for removing here -- nullable applies to primary types and direct object definitions. 
Here again, studyDisplayName refers to a definition  '$ref: "#/components/schemas/StudyDisplayName"'

service/src/main/resources/api/service_openapi.yaml
marikomedlock: I see, okay good to know. I think this comment applies to this property specifically, not necessarily the referenced object everywhere it's used. Perhaps we can refactor this as a comment inline here, rather than using the description field?

service/src/main/resources/api/service_openapi.yaml
marikomedlock: Understood, okay thanks for explaining. I clearly did not grok all the details of nullable/required/description as I was writing this file, so removing these is fine for this PR and I'll add a new ticket to review all these properties for the whole file.

marikomedlock (approved):
Doesn't need to be in this PR, but could you add a section to the DEVELOPER_INFO.md doc about how you generated all these changes please? Even if it's not automateable, I think it's useful info to preserve. Thanks again for all this cleanup work!

service/src/main/resources/api/service_openapi.yaml
dexamundsen: Sure, I will mark them as comments in the same place

dexamundsen: Scratch the previous comment, saw you other comment on reviewing all properties. If there are no concerns we could review the properties and update the comments appropriately at the same time (constraints for example).

service/src/main/resources/api/service_openapi.yaml
dexamundsen: Thanks, Lmk if you want me to create a ticket for properties review. Constraints like min/max, string length etc may need to be reviewed as well

dexamundsen:
> Doesn't need to be in this PR, but could you add a section to the DEVELOPER_INFO.md doc about how you generated all these changes please? Even if it's not automateable, I think it's useful info to preserve. Thanks again for all this cleanup work!

Done, added along with links to style guide

service/src/main/resources/api/service_openapi.yaml
marikomedlock: Constraints are fine but I'd prefer to keep comments like this in that describe behavior. Otherwise we have to go re-discover the behavior later. Reformatting is fine, but removing information I'd like to avoid.

service/src/main/resources/api/service_openapi.yaml
marikomedlock: Created https://verily.atlassian.net/browse/BENCH-2872.
Good idea about constraints, will note that also.

service/src/main/resources/api/service_openapi.yaml
dexamundsen: Noted, thanks

PR #837, opened 2024-05-01
Underlay: address warnings, minor optimizations

dexamundsen:


underlay/src/main/java/bio/terra/tanagra/filterbuilder/SchemaUtils.java
```diff
@@ -20,8 +20,7 @@ public static Literal toLiteral(Value value) {
       case TIMESTAMP_VALUE:
         return Literal.forTimestamp(
             new Timestamp(
-                value.getTimestampValue().getSeconds() * 1000
-                        + value.getTimestampValue().getNanos() * 1
+                value.getTimestampValue().getSeconds() * 1000 + value.getTimestampValue().getNanos()
                     ^ -6));
```
marikomedlock, line 24: This is not due to your changes here, but instead of this modification, I think we should replace this part with [below] for better readability.
```
        return Literal.forTimestamp(
                Timestamp.from(Instant
                        .ofEpochSecond(value.getTimestampValue().getSeconds() , value.getTimestampValue().getNanos())));
```

underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQQueryRunner.java
```diff
@@ -504,7 +516,7 @@ public HintQueryResult run(HintQueryRequest hintQueryRequest) {
                           sqlRowResult.get(attributeColName, DataType.STRING).getStringVal());
               if (attribute.isValueDisplay()
                   || attribute.getRuntimeDataType().equals(DataType.STRING)) {
-                // This is one value/count pair of an enum values hint.
+                // This is one value/count a pair of an enum values hint.
```
marikomedlock, line 519: I don't follow this change. Can you explain?

underlay/src/main/java/bio/terra/tanagra/utils/RetryUtils.java
```diff
@@ -164,10 +164,10 @@ public static <T, E extends Exception> T pollWithRetries(
           // the exception is not retryable: re-throw
           throw ex;
         } else {
-          // keep track of the last retryable exception so we can re-throw it in case of a timeout
+          // keep track of the last retryable exception, so we can re-throw it in case of a timeout
           lastRetryableException = ex;
         }
-        LOGGER.debug("Caught retryable exception: {}", ex);
+        LOGGER.debug("Caught retryable exception {}", ex.getMessage());
```
marikomedlock, line 170: I don't think this does the same thing. We want to log the exception (including cause, etc.) not just the message.

marikomedlock (commented):
Thanks again for all the cleanup! Same question as for the other PR -- anything we can automate here to keep things neat?

dexamundsen:
> Thanks again for all the cleanup! Same question as for the other PR -- anything we can automate here to keep things neat?

Unfortunately I dont think we have plugins to automate it. However IntelliJ tools make it easier to identity and change them en masse.

underlay/src/main/java/bio/terra/tanagra/utils/RetryUtils.java
dexamundsen: It doesn't - This was a warning since debug expected (string, throwable). 
fixing it to log the ex by removing the placeholder

underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQQueryRunner.java
dexamundsen: Just a grammar warning fixed - added the article 'a', no changes to existing comment

underlay/src/main/java/bio/terra/tanagra/filterbuilder/SchemaUtils.java
dexamundsen: done

underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQQueryRunner.java
marikomedlock: Yes, it's a minor thing, but I don't think this edit makes sense. Maybe changing it to "(value, count) pair" would be better wording?

underlay/src/main/java/bio/terra/tanagra/utils/RetryUtils.java
marikomedlock: Oh I see, my mistake. Yes, your fix is what I intended. Thanks for taking care of it!

marikomedlock (approved):

underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQQueryRunner.java
dexamundsen: Oh, apologies got it. I didnt grasp the entire context of the comment. My bad. Reverting it

dexamundsen: Updating it to reflect key,value pair

marikomedlock: Thank you!

PR #842, opened 2024-05-01
Improve adding criteria from a cohort/feature set

tjennison-google:
* Display the number of criteria in the source and their names.
* Disallow adding criteria from an empty source.

### DIFF
--- a/ui/src/addCohort.tsx
+++ b/ui/src/addCohort.tsx
@@ -1,7 +1,9 @@
 import Button from "@mui/material/Button";
 import Link from "@mui/material/Link";
+import Tooltip from "@mui/material/Tooltip";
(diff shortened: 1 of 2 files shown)


marikomedlock (approved):
Is an "empty source" a cohort with no included criteria or a data feature set with no criteria? Does it also include an empty criteria (e.g. gender but you don't pick any options in the dropdown)?

tjennison-google:
> Is an "empty source" a cohort with no included criteria or a data feature set with no criteria? Does it also include an empty criteria (e.g. gender but you don't pick any options in the dropdown)?

Yes. It shows the number of criteria that can actually be added. Gender can't be added to a feature set so it wouldn't count, but an empty condition criteria would.

PR #847, opened 2024-05-02
Enabled backend filter building by default.

marikomedlock:


tjennison-google (approved):
I forgot to send you the line to enable it on the UI. Add `REACT_APP_BACKEND_FILTERS=1` to `ui/.env`. The file doesn't exist yet so you'll have to create it.

PR #850, opened 2024-05-07
First pass at Basics Survey config files.

marikomedlock:
- Added two new entities and one new entity group to support the Basics Survey in the AoU underlays:
  - `surveyBasics` contains one row per node in the Basics Survey hierarchy, which includes answers, questions, topics, and the root node.
  - `surveyOccurrence` contains one row per survey answer per person. This data comes from the `observation` OMOP table but the entity is defined to look more like the `ds_survey` table, which is the index table that AoU uses for survey data concept sets.
  - `surveyBasicsPerson` defines the relationship between the `surveyBasics`, `surveyOccurrence`, and `person` entities.
- These config files have already been indexed in the `all-of-us-ehr-dev` project and can be tested locally.

This PR was a joint effort with @freemabd.


freemabd (approved):
Thanks

PR #852, opened 2024-05-10
Add numeric value answer to AoU Basics survey.

marikomedlock:
- Included the "Select a value" answer in the `surveyBasics` entity hierarchy.
- Added the `value_numeric` attribute to the `surveyOccurrence` entity for the numeric value answers.
- Added an instance-level modifier for the `surveyOccurrence.value_numeric` attribute in the `surveyBasicsPerson` entity group.
- Filtered the `surveyOccurrence` instances to only include those rows for the Basics survey. We can add other surveys as we add support for them, and can validate that their counts are correct.
- Enabled instance-level modifiers in the `surveyBasics` criteria selector and disabled multi-select. The UI does not support multi-select and instance-level modifiers (e.g. Labs & Measurements).

### DIFF
--- a/underlay/src/main/resources/config/display/aouSR2023Q3R2/criteriaselector/surveyBasics/surveyBasics.json
+++ b/underlay/src/main/resources/config/display/aouSR2023Q3R2/criteriaselector/surveyBasics/surveyBasics.json
@@ -57,5 +57,11 @@
       }
     }
   ],
-  "multiSelect": true
+  "multiSelect": false,
+  "valueConfigs": [
+    {
+      "attribute": "value_numeric",
+      "title": "Numeric value"
+    }
+  ]
 }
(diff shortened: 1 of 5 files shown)


underlay/src/main/resources/config/display/aouSR2023Q3R2/criteriaselector/surveyBasics/surveyBasics.json
```diff
@@ -57,5 +57,11 @@
       }
     }
   ],
-  "multiSelect": true
+  "multiSelect": false,
```
freemabd, line 60: Is there a reason you are turning off multi select?

freemabd (approved):
Looks good, just one question about multi select

underlay/src/main/resources/config/display/aouSR2023Q3R2/criteriaselector/surveyBasics/surveyBasics.json
marikomedlock: Yes, because the UI does not currently support multi-select and instance-level modifiers at the same time (e.g. see Labs & Measurements) because the modifier is specified in the cohort-level overview, and there's only one.
Screenshot 2024-05-13 at 11 33 59 AM

PR #859, opened 2024-05-14
Retry VerilyGroups access control initialization.

marikomedlock:
Updated the VerilyGroups access control implementation (only used on the Verily test/dev deployments) to retry the initialization sequence if it fails. We recently saw a failure where the GKE node temporarily lost connection to the Google metadata server, which supplies the application default credentials. This led to the VerilyGroups initialization failing with an invalid credentials error. This fix is to address this failure case.


service/local-dev/run_server.sh
```diff
@@ -88,13 +88,8 @@ if [[ ${disableAuthChecks} ]]; then
...
   echo "Enabling auth checks. bearer-token"
   export TANAGRA_AUTH_DISABLE_CHECKS=false
-  if [[ ${useAouUnderlays} ]]; then
-    export TANAGRA_AUTH_BEARER_TOKEN=true
-    export TANAGRA_AUTH_IAP_GKE_JWT=false
-  elif [[ ${useSdUnderlays} ]]; then
-    export TANAGRA_AUTH_BEARER_TOKEN=false
-    export TANAGRA_AUTH_IAP_GKE_JWT=true
-  fi
+  export TANAGRA_AUTH_BEARER_TOKEN=true
```
tjennison-google, line 91: What is this change for?

tjennison-google (approved):

service/local-dev/run_server.sh
marikomedlock: I think the intent here was to allow disabling auth checks or enabling them with the Google access bearer token configuration. The disable part of the conditional is correct (sets `TANAGRA_AUTH_DISABLE_CHECKS=true`), but this enable part was not always using the bearer token method. For SD underlays, it was acting differently -- expecting an IAP token -- which I don't think really makes sense for a locally running service.

PR #863, opened 2024-05-16
Update counts when backend cohort is updated

tjennison-google:
Since changes are generally made locally then synced to the backend, this created a race condition between updating cohorts and updating counts. After a change was made locally, a request would be made to save it but simultaneously the count UI would see the change a request an update. This worked previously because the filters were generated locally based on the most recent cohort, but now that they're being generated on the backend the backend may still have the previous version of the cohort if the count request arrives first. In theory, the updated cohort returned from saving would trigger an update as well but we specifically prevent updates with no actual changes to avoid spurious refreshes.

This adds tracking of the most recently recently received cohort from the backend. Counts now update based on this version of the cohort to ensure they're always waiting for the backend to updated first.


ui/src/overview.tsx
```diff
@@ -632,7 +666,11 @@ function ParticipantsGroup(props: {
                   "&:hover": { textDecoration: "none" },
                 }}
               >
-                {groupCountState.data?.toLocaleString()}
+                {isValid(groupCountState.data)
+                  ? groupCountState.data < 0
```
marikomedlock, line 670: Are you seeing counts returned that are <0?

marikomedlock (approved):

ui/src/overview.tsx
tjennison-google: No, -1 was just being used as a loading status. Not actually needed here anymore though so I removed it.

PR #865, opened 2024-05-24
[BENCH-2899], [BENCH-2973] Remove v2 prefix from version endpoint, add test

dexamundsen:
All authenticated endpoints start with v2, hence removing v2 prefix from version (in line with status endpoint). 
most monitoring setups use status endpoint, hence this change is unlikely to be a breaking change

----

verified that it runs in regular test suite 
https://github.com/DataBiosphere/tanagra/actions/runs/9275765397/job/25521235115?pr=865#step:8:182

### DIFF
--- a/service/gradle/task-dependencies.gradle
+++ b/service/gradle/task-dependencies.gradle
@@ -18,3 +18,6 @@ for (task in TASKS_ONLY_IF_COMPILE_JAVA) {
 
 // Explicit dep to avoid accidental dep deprecation warning.
 tasks.processResources.dependsOn(generateVersionProperties)
+
+// Status and version endpoint tests added to regression test
+tasks.regressionTests.dependsOn(generateVersionProperties)
--- a/service/gradle/task-dependencies.gradle
+++ b/service/gradle/task-dependencies.gradle
@@ -19,5 +19,5 @@ for (task in TASKS_ONLY_IF_COMPILE_JAVA) {
 // Explicit dep to avoid accidental dep deprecation warning.
 tasks.processResources.dependsOn(generateVersionProperties)
 
-// Status and version endpoint tests added to regression test
-tasks.regressionTests.dependsOn(generateVersionProperties)
+// Status and version endpoint tests added to test
+tasks.test.dependsOn(generateVersionProperties)
--- a/service/src/test/java/bio/terra/tanagra/app/controller/QueryCountRegressionTest.java
+++ b/service/src/test/java/bio/terra/tanagra/app/controller/QueryCountRegressionTest.java
@@ -1,11 +1,10 @@
-package bio.terra.tanagra.regression;
+package bio.terra.tanagra.app.controller;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import bio.terra.tanagra.api.filter.BooleanAndOrFilter;
-import bio.terra.tanagra.app.Main;
 import bio.terra.tanagra.app.configuration.FeatureConfiguration;
 import bio.terra.tanagra.proto.regressiontest.RTCriteria;
 import bio.terra.tanagra.proto.regressiontest.RTDataFeatureSet;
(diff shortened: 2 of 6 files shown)


service/src/test/java/bio/terra/tanagra/app/controller/QueryCountRegressionTest.java
```diff
@@ -41,21 +40,12 @@
...
 import org.junit.jupiter.params.provider.MethodSource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.boot.test.context.SpringBootTest;
-import org.springframework.test.context.ActiveProfiles;
-import org.springframework.test.context.ContextConfiguration;
-import org.springframework.test.context.junit.jupiter.SpringExtension;
 
-@ExtendWith(SpringExtension.class)
```
dexamundsen, line 55: already added to base class

service/src/test/java/bio/terra/tanagra/testing/BaseSpringUnitTest.java
```diff
@@ -18,16 +13,4 @@
 @ContextConfiguration(classes = Main.class)
 @SpringBootTest
 @SuppressWarnings("PMD.TestClassWithoutTestCases")
-public class BaseSpringUnitTest {
-  @MockBean protected RandomNumberGenerator randomNumberGenerator;
```
dexamundsen, line 22: not used

service/src/test/java/bio/terra/tanagra/app/controller/QueryCountRegressionTest.java
```diff
@@ -1,11 +1,10 @@
-package bio.terra.tanagra.regression;
+package bio.terra.tanagra.app.controller;
```
marikomedlock, line 1: This is the regression test runner, not a test of any of the controller classes, so I don't think it should be the same package.

service/gradle/task-dependencies.gradle
```diff
@@ -18,3 +18,6 @@ for (task in TASKS_ONLY_IF_COMPILE_JAVA) {
 
 // Explicit dep to avoid accidental dep deprecation warning.
 tasks.processResources.dependsOn(generateVersionProperties)
+
+// Status and version endpoint tests added to regression test
+tasks.regressionTests.dependsOn(generateVersionProperties)
```
marikomedlock, line 23: Regression tests are a way to check that cohort queries do not change over time for a given dataset ([docs](https://github.com/DataBiosphere/tanagra/blob/main/docs/REGRESSION_TESTING.md)). They should not include any specific endpoint tests.

service/src/test/java/bio/terra/tanagra/app/controller/UnauthenticatedApiControllerTest.java
```diff
@@ -0,0 +1,42 @@
...
+import bio.terra.tanagra.testing.BaseSpringUnitTest;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
+import org.springframework.mock.web.MockHttpServletResponse;
+import org.springframework.test.web.servlet.MockMvc;
+
+@Tag("regression-test")
```
marikomedlock, line 18: Per above, we can run this as part of the regular test suite, not the regression tests.

marikomedlock (commented):
Thanks for the tests!

service/gradle/task-dependencies.gradle
dexamundsen: Was trying to avoid adding another test tag, probably not a good idea. changing

service/src/test/java/bio/terra/tanagra/regression/QueryCountRegressionTest.java
```diff
@@ -41,21 +40,12 @@
...
 import org.junit.jupiter.params.provider.MethodSource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.boot.test.context.SpringBootTest;
-import org.springframework.test.context.ActiveProfiles;
-import org.springframework.test.context.ContextConfiguration;
-import org.springframework.test.context.junit.jupiter.SpringExtension;
 
-@ExtendWith(SpringExtension.class)
```
dexamundsen, line 55: Added to base class

service/src/test/java/bio/terra/tanagra/app/controller/QueryCountRegressionTest.java
dexamundsen: reverted

service/src/test/java/bio/terra/tanagra/app/controller/UnauthenticatedApiControllerTest.java
dexamundsen: changed

marikomedlock (approved):

PR #868, opened 2024-05-29
OpenAPI changes for temporal modifiers.

marikomedlock:
**ON HOLD** pending further UX/product input. SQL engine changes are mostly finished, so not much flexibility for significant logic changes, but UI design could change quite a bit how we want to persist this in the app DB.

Notes from discussion:
- We probably also want to store a flag for whether temporal modifiers are enabled for a particular criteria group section.
- Consider how to handle a temporal modifier toggle. AoU collapses any criteria groups together. We could do the same or just keep the top one. This may change how we store the criteria groups (e.g. use the same criteria groups property for default/temporal first condition, or keep just a single criteria groups property and flag each group somehow).
- Undecided how to handle the group by modifier. It's not supported in AoU or the current Tanagra SQL implementation. We could possibly support it in the future, but visualizations probably higher priority. Possibilities for now include ignoring it on the backend and/or flagging it somehow on the frontend. Whatever we choose, will probably need to flag this in the config somehow that it doesn't support temporal queries.


marikomedlock:
@tjennison-google FYI

PR #879, opened 2024-06-14
Move visualization configuration to config files

tjennison-google:
A future iteration may contain additional options to select where particular visualizations are used with the app but for now all are displayed in the demographic charts area.

Rename "display" config directory to "criteria" to better reflect its contents and better differentiate it from UI config.


underlay/src/main/java/bio/terra/tanagra/underlay/serialization/SZUnderlay.java
```diff
@@ -118,6 +118,23 @@ public class SZUnderlay {
               + "criteria definitions for standing up a new underlay.")
   public Set<String> prepackagedDataFeatures;
 
+  @AnnotatedField(
+      name = "SZUnderlay.visualizations",
+      markdown =
+          "List of paths of all the visualizations.\n\n"
+              + "A visualization contains all of the configuration to display a population level visualization "
```
marikomedlock, line 125: Does "population level" mean "cohort level"? I think the two levels we have are cohort and underlay/dataset.

underlay/src/main/java/bio/terra/tanagra/underlay/serialization/SZVisualization.java
```diff
@@ -0,0 +1,51 @@
...
+  public String name;
+
+  @AnnotatedField(name = "SZVisualization.title", markdown = "Visible title of the visualization.")
+  public String title;
+
+  @AnnotatedField(
+      name = "SZVisualization.config",
+      markdown =
+          "Serialized configuration of the visualization. VizConfig protocol buffer as JSON.")
+  public String config;
```
marikomedlock, line 24: What's the difference between the config/configFile and pluginConfig/pluginConfigFile? The pluginConfig isn't used in the gender & raceAndAge examples.

marikomedlock: Ah okay.

1. Just to make sure I follow, this isn't used for now because you're doing all the query building on the frontend for now, until we get a query API defined and working?
2. Maybe config -> queryConfig or dataSourceConfig and/or pluginConfig -> displayConfig then to clarify?
3. Also, if it's not going to change across plugins, then no need to use protos for the queryConfig right, and could we just leave it as Java like the rest of the shared config?

tjennison-google: > Ah okay.
> 
> 1. Just to make sure I follow, this isn't used for now because you're doing all the query building on the frontend for now, until we get a query API defined and working?

I'm not sure what you mean by this, since the config is being used to build the queries on the frontend. The pluginConfig isn't being used because while I did add the ability to customize the bar colors to the plugin I didn't bother to actually do any customization in the initial definitions.

> 2. Maybe config -> queryConfig or dataSourceConfig and/or pluginConfig -> displayConfig then to clarify?

Unlike the criteria selectors there aren't any backend plugins so I didn't think it was confusing. How about config -> dataConfig? Everything there may not end up in the query and sources are already used as part of the config definition to indicate where the individual variables come from.

> 3. Also, if it's not going to change across plugins, then no need to use protos for the queryConfig right, and could we just leave it as Java like the rest of the shared config?

We didn't fully settle on the API yet, but looking ahead to user defined visualizations I think we'll need to pass the configs through the API and eventually store them, and this seems like a natural way to do that.

marikomedlock: config -> dataConfig sgtm.

I'll put some time on your calendar tomorrow to talk about the visualization config -- probably easier than discussing via PR comments.

marikomedlock (approved):

PR #880, opened 2024-06-17
Truncate timestamp to microseconds before passing to BQ.

marikomedlock:
Truncate at the point we pass the timestamp value to BQ. This would also guarantee that we truncate in any other/future instances where the timestamp is turned into a literal.

### DIFF
--- a/underlay/src/main/java/bio/terra/tanagra/exception/NotFoundException.java
+++ b/underlay/src/main/java/bio/terra/tanagra/exception/NotFoundException.java
@@ -1,2 +1,23 @@
-package bio.terra.tanagra.exception;public class NotFoundException {
+package bio.terra.tanagra.exception;
+
+/** Custom exception class for not found exceptions. */
+public class NotFoundException extends RuntimeException {
+  /**
+   * Constructs an exception with the given message. The cause is set to null.
+   *
+   * @param message description of error that may help with debugging
+   */
+  public NotFoundException(String message) {
+    super(message);
+  }
+
+  /**
+   * Constructs an exception with the given message and cause.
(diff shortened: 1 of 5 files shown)


underlay/src/main/java/bio/terra/tanagra/exception/NotFoundException.java
```diff
@@ -0,0 +1,23 @@
+package bio.terra.tanagra.exception;
+
+/** Custom exception class for not found exceptions. */
+public class NotFoundException extends RuntimeException {
+  /**
+   * Constructs an exception with the given message. The cause is set to null.
+   *
+   * @param message description of error that may help with debugging
+   */
+  public NotFoundException(String message) {
```
dexamundsen, line 10: Are we able to use 
jakarta.ws.rs.NotFoundException / javax.*

underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQExecutor.java
```diff
@@ -193,7 +193,7 @@ public static String replaceFunctionsThatPreventCaching(
       String paramName =
           sqlParams.addParam(
               "currentDate",
-              Literal.forDate(DateTimeFormatter.ofPattern("yyyy-mm-dd").format(queryInstant)));
+              Literal.forDate(DateTimeFormatter.ofPattern("yyyy-MM-dd").format(queryInstant)));
```
dexamundsen, line 196: I was wondering about this one. wasnt sure, so had reverted my change

dexamundsen (approved):

PR #893, opened 2024-06-25
Add support for table visualizations

tjennison-google:
* Add a way to swap visualizations to a table view.
* Fix bar visualization to include the appropriate "core/" prefix.

### DIFF
--- a/ui/src/viz/vizContainer.tsx
+++ b/ui/src/viz/vizContainer.tsx
@@ -1,5 +1,8 @@
+import BarChartIcon from "@mui/icons-material/BarChart";
+import TableViewIcon from "@mui/icons-material/TableView";
 import Typography from "@mui/material/Typography";
 import { generateCohortFilter } from "cohort";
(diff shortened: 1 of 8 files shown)


tjennison-google:
image
image

ui/src/viz/vizContainer.tsx
```diff
@@ -71,6 +80,11 @@ export function VizContainer(props: VizContainerProps) {
             config,
             data,
             plugin: getVizPlugin(config.plugin, config.pluginConfig),
+            tablePlugin: getVizPlugin("core/table", {
```
marikomedlock, line 83: This table visualization will apply to all future visualizations also, not just the existing bar chart -- is that right?

marikomedlock (approved):

ui/src/viz/vizContainer.tsx
tjennison-google: Yeah

PR #896, opened 2024-06-28
do not merge - ui auth draft

dexamundsen:


.github/workflows/generated-files.yaml
```diff
@@ -7,7 +7,7 @@ on:
   push:
     branches: [ main ]
   pull_request:
-    branches: [ '**' ]
+    branches: [ dextest ]
```
dexamundsen, line 10: This (and similar changes in workflows) will be removed.  There are added just to avoid triggering tests when not needed

ui/.env.development
```diff
@@ -1,2 +1,4 @@
 REACT_APP_USE_FAKE_API=n
 DANGEROUSLY_DISABLE_HOST_CHECK=true
+REACT_APP_AUTH0_CLIENT_ID=
```
dexamundsen, line 3: env values will be moved downstream

ui/.env.development
```diff
@@ -1,2 +1,4 @@
 REACT_APP_USE_FAKE_API=n
 DANGEROUSLY_DISABLE_HOST_CHECK=true
+REACT_APP_AUTH0_DOMAIN=verily-terra-dev.us.auth0.com
```
dexamundsen, line 3: env values will be moved downstream

dexamundsen:
created https://github.com/DataBiosphere/tanagra/pull/924

PR #898, opened 2024-07-01
Add lib constraints to address synk warnings

dexamundsen:
Add lib constraints to address synk warnings


annotationProcessor/build.gradle
```diff
@@ -31,10 +36,16 @@ repositories {
 }
 
 dependencies {
+    // added to address synk warnings
+    constraints {
+        implementation 'com.azure:azure-identity:1.12.2'
```
marikomedlock, line 41: Is this due to `terra-common-lib`, and if so, is it possible to exclude it instead of pegging the version here?

dexamundsen: Yes, it is from common-lib. exclusion is a better option, done.

marikomedlock (approved):

PR #901, opened 2024-07-02
Allow configuring export infrastructure per underlay.

marikomedlock:
Previously export infrastructure (temporary BQ datasets, GCS buckets) could only be configured per service deployment, using the service application properties. This PR allows it to be configured per underlay. If export infrastructure is not specified per underlay, it defaults to the per service values. This is to better support coming service deployments that serve multiple underlays (e.g. SD + eMerge, Verily Workbench).

Also moved the logic for writing raw data to GCS behind the `QueryRunner` interface, so we can use the underlay-specific config and in the future, potentially different implementations per underlay.

### DIFF
--- a/docs/generated/UNDERLAY_CONFIG.md
+++ b/docs/generated/UNDERLAY_CONFIG.md
@@ -127,6 +127,32 @@ Pointers to the source and index BigQuery datasets.
 
 Valid locations for BigQuery are listed in the GCP [documentation](https://cloud.google.com/bigquery/docs/locations).
 
+### SZBigQuery.exportBucketNames
+**optional** List [ String ]
+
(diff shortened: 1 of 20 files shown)


docs/generated/UNDERLAY_CONFIG.md
```diff
@@ -127,6 +127,32 @@ Pointers to the source and index BigQuery datasets.
...
 
+### SZBigQuery.exportBucketNames
+**optional** List [ String ]
+
+Comma separated list of all GCS bucket names that all export models can use. Only include the bucket name, not the gs:// prefix. Required if there are any export models that need to write to GCS.
+
+These buckets must live in the [query project](#szbigqueryqueryprojectid) specified above.
+
+You can also specify these export buckets per-deployment, instead of per-underlay, by using
+
```
tjennison-google, line 138: Extra line?

tjennison-google (approved):

PR #908, opened 2024-07-09
Persist temporal criteria group sections in DB.

marikomedlock:
Persist temporal criteria group sections in PostGres/MySQL application DB. Enabled temporal modifiers for the conditions criteria selector on the `cmssynpuf` underlay, for testing.

This PR does not include changes to the filter building, so the counts generated for these temporal criteria group sections are incorrect -- they will be the same as if the operator were `OR/ANY` and use criteria only for the top block of criteria groups. The filter building changes are coming in a follow-on PR.

### DIFF
--- a/service/src/main/resources/db/changelog.xml
+++ b/service/src/main/resources/db/changelog.xml
@@ -25,4 +25,6 @@
     <include file="changesets/20230922_initial_schema.yaml" relativeToChangelogFile="true"/>
     <include file="changesets/20231116_output_attribute.yaml" relativeToChangelogFile="true"/>
     <include file="changesets/20240304_selector_column.yaml" relativeToChangelogFile="true"/>
+    <include file="changesets/20240326_remove_criteria_group_columns.yaml" relativeToChangelogFile="true"/>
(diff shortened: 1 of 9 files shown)


service/src/main/resources/db/changelog.xml
```diff
@@ -25,4 +25,6 @@
     <include file="changesets/20230922_initial_schema.yaml" relativeToChangelogFile="true"/>
     <include file="changesets/20231116_output_attribute.yaml" relativeToChangelogFile="true"/>
     <include file="changesets/20240304_selector_column.yaml" relativeToChangelogFile="true"/>
+    <include file="changesets/20240326_remove_criteria_group_columns.yaml" relativeToChangelogFile="true"/>
```
tjennison-google, line 28: What is this? Is it missing?

tjennison-google (approved):

service/src/main/resources/db/changelog.xml
marikomedlock: Yeah, it was missing from a [previous PR](https://github.com/DataBiosphere/tanagra/pull/788).

PR #912, opened 2024-07-10
Add  remaining default visualizations

tjennison-google:
This includes the remaining top X visualizations and a couple specialized ones for AoU.

AoU requested age at consent for some of the visualizations but it's not currently configured so they'll have to be updated later.


marikomedlock (approved):

dolbeew (approved):
LGTM

PR #914, opened 2024-07-12
Build cohort filter on backend when creating a review.

marikomedlock:
This was causing an error when creating a review from an empty cohort. We no longer fully support frontend filter building, so removing the associated API fields here will come in future tech debt PRs.

### DIFF
--- a/service/src/main/java/bio/terra/tanagra/app/controller/ReviewsApiController.java
+++ b/service/src/main/java/bio/terra/tanagra/app/controller/ReviewsApiController.java
@@ -46,6 +46,7 @@ import bio.terra.tanagra.service.artifact.reviewquery.ReviewInstance;
 import bio.terra.tanagra.service.artifact.reviewquery.ReviewQueryOrderBy;
 import bio.terra.tanagra.service.artifact.reviewquery.ReviewQueryRequest;
 import bio.terra.tanagra.service.artifact.reviewquery.ReviewQueryResult;
+import bio.terra.tanagra.service.filter.*;
 import bio.terra.tanagra.underlay.Underlay;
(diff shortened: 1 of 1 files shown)


tjennison-google (approved):
I have been thinking we should remove the frontend filter building but we've had other priorities.

marikomedlock:
> I have been thinking we should remove the frontend filter building but we've had other priorities.

Yes, agreed. Maybe it makes sense to do this as part of a general review of our OpenAPI spec? I know we've had a longstanding plan to go through and make sure all the required/nullable/etc make sense also. If we can plan out some tickets, I'm game to tackle this over the next ~1-2 months as a background task. Maybe one set/tag of endpoints at a time? It would require backend & frontend joint changes, so probably makes sense for us to tackle endpoints in a joint branch/PR. Thoughts?

tjennison-google:
> > I have been thinking we should remove the frontend filter building but we've had other priorities.
> 
> Yes, agreed. Maybe it makes sense to do this as part of a general review of our OpenAPI spec? I know we've had a longstanding plan to go through and make sure all the required/nullable/etc make sense also. If we can plan out some tickets, I'm game to tackle this over the next ~1-2 months as a background task. Maybe one set/tag of endpoints at a time? It would require backend & frontend joint changes, so probably makes sense for us to tackle endpoints in a joint branch/PR. Thoughts?

Sounds good to me. I can start by removing all the frontend filter generation related code and calls first. Then you'll be free to remove the API and backend parts because they'll be unused. We'll need to collaborate on some of the OpenAPI cleanup but hopefully that won't be too bad. For some changes I can probably make the necessary updates before the API changes are merged (for things like required/nullable) but otherwise we can figure out some kind of joint branch/PR.

PR #918, opened 2024-07-16
Address Java synk security vulnerabilities

dexamundsen:
Fixed issues with spring tomcat-core (bumped spring in terra-common-lib)


chenchals (approved):
lgtm

PR #919, opened 2024-07-18
Minor UI refactor

dexamundsen:
- Add util for process.env
- Minor cleanup of imports
- Add errorHandler.tsx for error related fucntions
- Split route objects into a separate file

---

Reason for change: cleaner and easier when auth code is added


ui/src/errorhandler.tsx
```diff
@@ -0,0 +1,33 @@
+import { Button } from "@mui/material";
```
tjennison-google, line 1: This file should be be "errorPage.tsx" to match the component and standard casing.

ui/src/environment.tsx
```diff
@@ -0,0 +1,18 @@
+// Items here should be added to environment.js.template as well.
```
tjennison-google, line 1: Should this file exist?

ui/src/environment.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+  REACT_APP_USE_EXIT_URL: boolean;
+  REACT_APP_EXIT_URL: string;
+
+  REACT_APP_USE_FAKE_API: string;
+
+  REACT_APP_GET_LOCAL_AUTH_TOKEN: string;
+};
+
+export function getEnvironment(): Environment {
```
tjennison-google, line 16: What is the context for this change? Particularly considering the comment about `environment.js.template`, I'm having trouble evaluating it.

tjennison-google (approved):

marikomedlock (commented):
LGTM, but will let @tjennison-google approve.

ui/src/errorhandler.tsx
dexamundsen: renamed

ui/src/environment.tsx
dexamundsen: Removed comment, had copied from VWB-ui which may not be applicable here

ui/src/environment.tsx
dexamundsen: environment.js.template may not b applicable here. removed

PR #937, opened 2024-07-28
Add support for authn / authz with JWT access tokens

dexamundsen:
Add Auth0 support using env variables. At this time given the limited use of Auth0, setting the client via env variable.
Client secret is optional and used only for confidential clients. 

Proposing to add the code changes here instead of downstream implementation since this is generic code. Only the env values are implementation specific. 

Tested locally - able to get underlays for user based on access token


service/src/main/java/bio/terra/tanagra/app/authentication/AuthInterceptor.java
```diff
@@ -32,10 +33,21 @@ public class AuthInterceptor implements HandlerInterceptor {
...
+  private Auth0Client auth0Client;
 
   @Autowired
   public AuthInterceptor(AuthenticationConfiguration authenticationConfiguration) {
     this.authenticationConfiguration = authenticationConfiguration;
+
+    if (authenticationConfiguration.isAuth0Jwt()) {
+      this.auth0Client =
+          (authenticationConfiguration.isAuth0Jwt())
+              ? new Auth0Client(
```
tjennison-google, line 45: No need for the ternary inside the if.

service/src/main/java/bio/terra/tanagra/app/authentication/AuthInterceptor.java
```diff
@@ -82,20 +93,28 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
                 jwt,
                 authenticationConfiguration.getGcpProjectNumber(),
                 authenticationConfiguration.getGkeBackendServiceId());
+
```
tjennison-google, line 96: Is this a common style in VWB? I haven't seen it before.

service/src/main/java/bio/terra/tanagra/app/authentication/AuthInterceptor.java
```diff
@@ -32,10 +33,21 @@ public class AuthInterceptor implements HandlerInterceptor {
   private static final String OPENAPI_TAG_AUTH_NOT_REQUIRED = "Unauthenticated";
 
   private final AuthenticationConfiguration authenticationConfiguration;
+  private Auth0Client auth0Client;
```
tjennison-google, line 36: final?

tjennison-google (commented):
I'm not very familiar with how VWB is set up, but should we be following the proxy model that it uses? It also wasn't clear to me from that discussion how the flow works for local dev so maybe this is it?

service/src/main/java/bio/terra/tanagra/app/authentication/AuthInterceptor.java
dexamundsen: silly me, missed  to remove one

service/src/main/java/bio/terra/tanagra/app/authentication/AuthInterceptor.java
dexamundsen: only added to improve readability, new line not enforced here. We use spotlessJava to enforce general code style.

dexamundsen:
> I'm not very familiar with how VWB is set up, but should we be following the proxy model that it uses? It also wasn't clear to me from that discussion how the flow works for local dev so maybe this is it?

the backend already has a proxy that autheticates the jwt in request, but it does not provide the decoded entries to the service. 
VWB services always get user details from sam. it is a discussion for when Mariko is back as to whether we want to call auth0 (more flexible, short term) or sam (needs interface and product direction). 
existing code that handles bearer token only handles jwt from google, not the ones from auth0

tjennison-google (approved):

service/src/main/java/bio/terra/tanagra/service/authentication/JWTAccessTokenUtils.java
```diff
@@ -0,0 +1,112 @@
...
+      return kf.generatePublic(keySpec);
+    } catch (NoSuchAlgorithmException e) {
+      throw new InvalidCredentialsException(
+          "Public key generation failed, algorithm not be found: " + algorithm, e);
+    } catch (InvalidKeySpecException e) {
+      throw new InvalidCredentialsException("Public key generation failed, invalid key", e);
+    }
+  }
+
+  public static PublicKey readPublicKeyFromFile(String publicKeyPemFilePath, String algorithm)
```
tjennison-google, line 67: private?

service/src/main/java/bio/terra/tanagra/service/authentication/JWTAccessTokenUtils.java
```diff
@@ -0,0 +1,112 @@
...
+    reader.close();
+    return content;
+  }
+
+  private static PublicKey getPublicKey(byte[] keyBytes, String algorithm) {
+    /*
+     if the file contains certificate, public key must be extracted from it
+     # cat <cert.file> | openssl x509 -pubkey -noout > pub.file
+
+     CertificateFactory cf = CertificateFactory.getInstance("X.509");
```
tjennison-google, line 51: Just delete this code if it's not being used. Don't the above instructions remove the need for it anyways? Or support both if you think it's needed.

tjennison-google (approved):

service/src/main/java/bio/terra/tanagra/service/authentication/JWTAccessTokenUtils.java
dexamundsen: Changed

service/src/main/java/bio/terra/tanagra/service/authentication/JWTAccessTokenUtils.java
dexamundsen: This is information is found easily online, removed. 
No need to support both

service/build.gradle
```diff
@@ -92,6 +92,10 @@ dependencies {
     implementation 'org.springframework.boot:spring-boot-starter-web'
     implementation 'org.springframework.security:spring-security-core'
 
+    // JWT dependencies (used to parse access token)
```
marikomedlock, line 95: nit: "used to parse token from Auth0"?

service/src/main/java/bio/terra/tanagra/app/configuration/AuthenticationConfiguration.java
```diff
@@ -15,10 +15,10 @@
     name = "Authentication",
     markdown =
         "Configure the authentication model.\n\n"
-            + "There are four separate flags that control which model is used: "
-            + "`tanagra.auth.disableChecks`, `tanagra.auth.iapGkeJwt`, `tanagra.auth.iapAppEngineJwt`, `tanagra.auth.bearerToken`. "
+            + "There are five separate flags that control which model is used: "
+            + "`tanagra.auth.disableChecks`, `tanagra.auth.iapGkeJwt`, `tanagra.auth.iapAppEngineJwt`, `tanagra.auth.bearerToken`, `tanagra.auth.auth0Jwt`. "
```
marikomedlock, line 19: `tanagra.auth.auth0Jwt` seems like a good name for the flag. It doesn't match the field names below though. Can you please update the flags below to `tanagra.auth.auth0Jwt.___` to match here?

service/src/main/java/bio/terra/tanagra/app/authentication/AuthInterceptor.java
```diff
@@ -32,10 +34,19 @@ public class AuthInterceptor implements HandlerInterceptor {
   private static final String OPENAPI_TAG_AUTH_NOT_REQUIRED = "Unauthenticated";
 
   private final AuthenticationConfiguration authenticationConfiguration;
+  private final JwtAccessTokenUtils jwtAccessTokenUtils;
```
marikomedlock, line 37: Instead of adding a new property to this class that's only used for one auth configuration, can you follow the existing pattern of calling a static method in the token decoder utility class? If you want to do some one-time setup (e.g. reading the key file), you can create a singleton instance and store that in the utility class.

service/src/main/java/bio/terra/tanagra/service/authentication/JwtAccessTokenUtils.java
```diff
@@ -0,0 +1,95 @@
...
+import org.bouncycastle.util.io.pem.PemObject;
+import org.bouncycastle.util.io.pem.PemReader;
+import org.springframework.util.Assert;
+
+public class JwtAccessTokenUtils {
+  private final JWTVerifier jwtVerifier;
+  private final ObjectMapper mapper = new ObjectMapper();
+  private final Base64.Decoder decoder = Base64.getUrlDecoder();
+
+  private static byte[] parsePEMFile(String pemFileName) throws IOException {
```
marikomedlock, line 35: nit: Consider moving the private methods to the bottom of the file, or after the public methods they support? I think it's a little easier to read the class when the constructor & public methods are at the top, since that's how someone would get into this class to begin with.

service/src/main/java/bio/terra/tanagra/service/authentication/JwtAccessTokenUtils.java
```diff
@@ -0,0 +1,95 @@
...
+import java.security.spec.InvalidKeySpecException;
+import java.security.spec.X509EncodedKeySpec;
+import java.util.Base64;
+import java.util.Map;
+import org.apache.commons.lang3.StringUtils;
+import org.bouncycastle.util.io.pem.PemObject;
+import org.bouncycastle.util.io.pem.PemReader;
+import org.springframework.util.Assert;
+
+public class JwtAccessTokenUtils {
```
marikomedlock, line 30: We support JWTs from IAP also, so I think putting Auth0 in the classname somewhere would be helpful e.g. `Auth0JwtTokenUtils`. Also, I think JWT already includes "token" in the acronym.

service/src/main/java/bio/terra/tanagra/service/authentication/JwtAccessTokenUtils.java
```diff
@@ -0,0 +1,95 @@
...
+            /* RSAPrivateKey= */ null);
+    jwtVerifier = JWT.require(algorithm).withIssuer(issuer).build();
+  }
+
+  public UserId getUserIdFromToken(String accessToken) {
+    Assert.isTrue(StringUtils.isNotBlank(accessToken), "user accessToken empty");
+
+    try {
+      DecodedJWT verifiedJWT = jwtVerifier.verify(accessToken);
+      String payloadJSON = new String(decoder.decode(verifiedJWT.getPayload()));
```
marikomedlock, line 78: We verify the JWT when Tanagra is behind IAP, but from the discussion in Slack, it's not clear to me that we should be doing the verifying here or whether the proxy in front will do that. Maybe better for a discussion over GVC?

dexamundsen: From discussion -
downstream VWB deployment is behind proxy, which takes care of authn. Services call sam to decode token.
verification was added for non-proxy deployments, and hence will be made optional here.

service/src/main/java/bio/terra/tanagra/service/authentication/JwtAccessTokenUtils.java
dexamundsen: This was an artifact of me going back and forth on names. Will rename as discussed to `unauth-jwt`

service/src/main/java/bio/terra/tanagra/service/authentication/JwtAccessTokenUtils.java
dexamundsen: done

service/build.gradle
dexamundsen: Updated, avoiding `for auth0` since this can be used for other token issuers as well. there are other third party libraries as well, though not widely used

service/src/main/java/bio/terra/tanagra/app/authentication/AuthInterceptor.java
dexamundsen: done

service/src/main/java/bio/terra/tanagra/app/configuration/AuthenticationConfiguration.java
dexamundsen: updated to `tanagra.auth.unverifiedJwt`

service/src/main/java/bio/terra/tanagra/app/configuration/AuthenticationConfiguration.java
```diff
@@ -96,6 +105,25 @@ public class AuthenticationConfiguration {
       environmentVariable = "TANAGRA_AUTH_GKE_BACKEND_SERVICE_ID")
   private String gkeBackendServiceId;
 
+  @AnnotatedField(
+      name = "tanagra.auth.unverifiedJwt.issuer",
+      markdown = "The issuer of JWT used for its verification. ",
+      environmentVariable = "TANAGRA_AUTH_UNVERIFIED_JWT_ISSUER")
+  private String unverifiedJwtIssuer;
```
marikomedlock, line 112: It's confusing to me to have configuration for verifying a JWT when the authentication method is called "unverifiedJwt". Perhaps you can treat this as a different authentication method (e.g. "auth0Jwt" or "auth0VerifiedJwt")?

dexamundsen: Autho library, but it can work for other issuers too. If we want to tailor it to only auth0, we can name it TANAGRA_AUTH_AUTH0_JWT
IapJwtUtils could be modified to be generic for all JWT. trying that if those can be renamed and refactored

dexamundsen: Renamed to TANAGRA_AUTH_JWT for generic JWT, refactored IapTokenUtils to share the JWT decoding

service/src/main/java/bio/terra/tanagra/service/authentication/JwtUtils.java
```diff
@@ -0,0 +1,125 @@
...
+      return makeUserId(
+          payloadMap.getOrDefault("sub", null), payloadMap.getOrDefault("email", null), jwt);
+
+    } catch (IllegalArgumentException | JsonProcessingException | IndexOutOfBoundsException e) {
+      throw new InvalidCredentialsException("Error decoding user info from JWT access token", e);
+    }
+  }
+
+  private static UserId makeUserId(String sub, String email, String token) {
+    if (StringUtils.isEmpty(sub) || StringUtils.isEmpty(email)) {
```
marikomedlock, line 93: I think there should be a null check here also.

marikomedlock (approved):
LGTM -- one comment inline about adding a null check.

service/src/main/java/bio/terra/tanagra/service/authentication/JwtUtils.java
dexamundsen: StringUtils.isEmpty includes null check. Did not think token needs a null check, because userId can be created and used w/o it

marikomedlock: Ah gotcha, I meant for the email & subject, not the token.

PR #947, opened 2024-08-06
Handle non-string claim values in JWT

dexamundsen:
Issue - claim values (such as aud) may be array-of-strings. Current code only handles strings, this may fail deser even if the value is not needed
Fix - change deser return value to Map<string, object>, and explicitly cast the required claims


service/src/main/java/bio/terra/tanagra/service/authentication/JwtUtils.java
```diff
@@ -80,9 +80,11 @@ public static UserId verifyJwtAndGetUserid(
   public static UserId getUserIdFromJwt(String jwt) {
     try {
       String payloadJSON = new String(DECODER.decode(jwt.split("\\.")[1]));
-      Map<String, String> payloadMap = MAPPER.readValue(payloadJSON, new TypeReference<>() {});
+      Map<String, Object> payloadMap = MAPPER.readValue(payloadJSON, new TypeReference<>() {});
       return makeUserId(
-          payloadMap.getOrDefault("sub", null), payloadMap.getOrDefault("email", null), jwt);
+          (String) payloadMap.getOrDefault("sub", null),
+          (String) payloadMap.getOrDefault("email", null),
```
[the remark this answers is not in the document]
dexamundsen: In short, yes. 
The return value of MAPPER.readValue() is actually Map<String, object>. 
good idea to check the type though sub and email should always be string in jwt (iiuc)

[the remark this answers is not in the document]
dexamundsen: Added

service/src/main/java/bio/terra/tanagra/service/authentication/JwtUtils.java
```diff
@@ -71,6 +74,18 @@ public static UserId verifyJwtAndGetUserid(
     }
   }
 
+  // Check and cast the object
+  @SuppressWarnings("unchecked")
+  private static <T> T castObject(String objName, @Nullable Object obj, Class<T> clazz) {
+    if ((obj != null) && !clazz.equals(obj.getClass())) {
+      throw new IllegalArgumentException(
```
marikomedlock, line 81: Since this is just throwing an exception, instead of suppressing and using null, I think you can just catch a `ClassCastException` in the catch block, no need for this helper method. e.g.
```
      return makeUserId(
         (String) payloadMap.getOrDefault(CLAIM_SUB, null),
          (String) payloadMap.getOrDefault(CLAIM_EMAIL, null),
          jwt);

    } catch (ClassCastException e) {
      LOGGER.error("Subject id or email claims are not strings as expected: {}, {}", payloadMap.getOrDefault(CLAIM_SUB, null), payloadMap.getOrDefault(CLAIM_EMAIL, null)));
      throw new InvalidCredentialsException(...)
```

dexamundsen: Forgot about ClassCastException, its cleaner.

dexamundsen: Skipped an explicit log here - since the other catch both will still log the exception cause.

marikomedlock (approved):

PR #951, opened 2024-08-07
[OOB] - Update tanagra admin client after java 17

chenchals:
- Updated tanagra admin client to 0.0.98-SNAPSHOT
- Updated gradle.lockfile (s)
- Updated pointers for sd and emerge underlays


service/local-dev/run_server.sh
```diff
@@ -69,19 +69,20 @@ elif [[ ${useAouUnderlays} ]]; then
...
 elif [[ ${useEmergeUnderlays} ]]; then
   echo "Using eMerge underlay."
-  export TANAGRA_UNDERLAY_FILES=emerge/emerge_20240603_local,sd/sd020230831_local
+  export TANAGRA_UNDERLAY_FILES=emerge/emerge_local,sd/sd_local
   export TANAGRA_EXPORT_SHARED_GCP_PROJECT_ID=vumc-emerge-dev
   export TANAGRA_EXPORT_SHARED_GCS_BUCKET_NAMES=vumc-emerge-dev-exports
   # uncomment both lines below for emerge access-control model
-  # export TANAGRA_ACCESS_CONTROL_BASE_PATH=https://vumc-emerge-dev.victrvumc.org
+  # export TANAGRA_ACCESS_CONTROL_BASE_PATH=https://vumc-cohort-builder-dev.victrvumc.org
+  # export TANAGRA_ACCESS_CONTROL_BASE_PATH=https://vumc-cohort-builder-dev.victrvumc.org
```
marikomedlock, line 85: Duplicate line?

service/build.gradle
```diff
@@ -129,7 +129,7 @@ dependencies {
     implementation "org.glassfish.jersey.media:jersey-media-multipart:${gradle.vJersey2}"
     implementation "org.glassfish.jersey.media:jersey-media-json-jackson:${gradle.vJersey2}"
 
-    implementation 'org.vumc.vda.tanagra:admin-client:0.0.85-SNAPSHOT'
+    implementation 'org.vumc.vda.tanagra:admin-client:0.0.98-SNAPSHOT'
```
dexamundsen, line 132: re: https://github.com/DataBiosphere/tanagra/blob/main/service/build.gradle#L125
Jersey libraries were kept at v2 for java 11 client. These can be updated to v3. Common versions set and defined here - https://github.com/DataBiosphere/tanagra/blob/main/build.gradle

service/local-dev/run_server.sh
chenchals: aaah.. will fix it next round...

service/build.gradle
chenchals: I tried change to v3- but other dependencies broke for tanagara-admin hence will do it later.

PR #952, opened 2024-08-07
Use Jersey3 libs in service

dexamundsen:
@chenchals Only did the changes (since they were small) but it may need testing on the vumc deployment


dexamundsen:
@chenchals does the client generate code in javax (old, for spring 2.x) or jakarta (new, for spring 3.x) namespace. This can set like this - https://github.com/DataBiosphere/tanagra/blob/main/client/gradle/open-api.gradle#L34

PR #957, opened 2024-08-16
Check for token expiry in auth0

dexamundsen:
issue - authcontext.expired was set to false as long as the authentication created a valid token in the local cache, even if it was expired. hence the relogin (re-authorize) flow was not called in UI, but when the token was validated in backend proxy it would fail

fix - there is no out of the box way in auth0-react to validate if the token is expired. Check if the token is close to expiration (60 seconds) and set a state.tokenExpired to true. This state in turn will reset the authcontext.expired, thus triggering the auth flow again. As long as the access token is valid and not expired, the user does not need to re-login. Auth0 updates the idToken with new expiry 

testing - 
- in auth0 set the token expiry to 60 seconds
- trigger fresh login flow
- verify that the exp claim in idToken is in 60 seconds; expiresAt in access_token may be different
- after a few minutes reopen or refresh the page. since it is past the expiry time, you see a login page

### DIFF
--- a/ui/src/auth/auth0OAuth.tsx
+++ b/ui/src/auth/auth0OAuth.tsx
@@ -43,7 +43,8 @@ function Auth0ProviderWithClient() {
     logout,
   } = useAuth0();
 
-  // Needed for user state to get updated: calls auth0 if token is expired
+  // Needed for user state to get updated. getAccessTokenSilently: calls auth0
+  // if there is no valid token in the cache, does not check expiry
   useEffect(() => {
     if (isAuthenticated) {
       getAccessTokenSilently();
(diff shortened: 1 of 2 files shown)


ui/src/auth/auth0OAuth.tsx
```diff
@@ -37,26 +37,28 @@ function Auth0ProviderWithClient() {
...
     loginWithRedirect,
     logout,
   } = useAuth0();
 
-  // This is needed for user state to get updated.
+  // Needed for user state to get updated: calls auth0 if token is expired
   useEffect(() => {
     if (isAuthenticated) {
-      getIdTokenClaims();
+      getAccessTokenSilently();
```
marikomedlock, line 49: Just curious about the function change -- this token is still an ID token right?

marikomedlock (approved):

ui/src/auth/auth0OAuth.tsx
dexamundsen: Yes, we still get the id token when we get the auth token at line 78. 

the reason for this function change - 
from the docs - https://auth0.github.io/auth0-react/interfaces/Auth0ContextInterface.html#getAccessTokenSilently

getAccessTokenSilently will automatically invoke login flow is there is no valid token (does not check expiry). 
Updating the comment

PR #967, opened 2024-08-26
Update doc on files in export GCS bucket, Underlays

dexamundsen:


docs/DEPLOYMENT_OVERVIEW.md
```diff
@@ -83,6 +83,9 @@ A service deployment lives in a GCP project configured with the items below.
     }
    ]
    ```
+    - The files stored in this bucket are either exported to VWB or downloaded to the user's computer, 
+      depending on the export model configured and selected. If desired they can be set to be automatically deleted
+      as part of the bucket's lifecyle configuration.
```
marikomedlock, line 88: There are other options for export also -- I made that comment in the internal PR based on the export options we have configured for the VWB deployment. Maybe no need to list all the export options here, this is just devops info.

nit: Slightly stronger "recommend setting the bucket to automatically delete objects after some expiration time" rather than "if desired". Also could you please include a link to the GCS lifecycle page? https://cloud.google.com/storage/docs/lifecycle

docs/UNDERLAYS.md
```diff
@@ -0,0 +1,15 @@
+# Underlays
```
marikomedlock, line 1: This is good information to document, but most of these underlays are not accessible by any contributor to this repo. Perhaps our [internal wiki](https://github.com/verily-src/verily-tanagra-backend/wiki) is a better place to store this info?

dexamundsen: moved

docs/DEPLOYMENT_OVERVIEW.md
dexamundsen: Done, retained the export models phrase for completeness

docs/DEPLOYMENT_OVERVIEW.md
```diff
@@ -83,6 +83,10 @@ A service deployment lives in a GCP project configured with the items below.
     }
    ]
    ```
+    - The files stored in this bucket are either exported to VWB or downloaded to the user's computer,
```
marikomedlock, line 86: The files being exported is not really part of Tanagra. We just write the files to the bucket, generate a signed URL, and either display that in the UI for the user to click on to download, or redirect to another configurable URL. So I think this statement is misleading. There is no VWB-specific code in Tanagra.

dexamundsen: How about this update

marikomedlock (approved):

PR #970, opened 2024-08-28
Add survey criteria

tjennison-work:
* Loads all survey data for real time filtering.
* Improves question/answer name handling, though indexing changes will
  be needed to handle global search correctly.
* Multi-select plus instance level data still to be dealt with.
* Config changes will come separately since they'll break stored data.
* Copies some filter generation code from entity group criteria but this 
  allows the config and selection data to be distinct from entity group  
  criteria.  There is already some divergence and this allows them to    
  evolve separately. This wasn't originally planned but handling         
  backwards compatability while switching to a new                       
  plugin/config/selection data seemed pretty sketchy.                    
* Fixes a bug where entity group criteria with modifiers but no          
  selection weren't handled correctly.
* Testing is weird because we don't currently have survey data. The
  basics are the same as entity group criteria so using the same data
  for now but TBD how we want to handle this.


underlay/src/main/proto/criteriaselector/configschema/survey.proto
```diff
@@ -0,0 +1,44 @@
...
+    // hierarchy view.
+    SortOrder sort_order = 2;
+  }
+
+  // Entity groups where the related entity is what is selected (e.g. condition
+  // when filtering condition_occurrences).
+  repeated EntityGroupConfig entity_groups = 2;
+
+  // Optional configuration of a categorical or numeric value associated with
+  // the selection (e.g. a measurement value). Applied to the entire selection
```
marikomedlock, line 31: nit: "e.g. a survey numeric answer value"?

underlay/src/main/proto/criteriaselector/configschema/survey.proto
```diff
@@ -0,0 +1,44 @@
...
+    // The id of the entity group.
+    string id = 1;
+
+    // The sort order applied to this entity group when displayed in the
+    // hierarchy view.
+    SortOrder sort_order = 2;
+  }
+
+  // Entity groups where the related entity is what is selected (e.g. condition
+  // when filtering condition_occurrences).
```
marikomedlock, line 27: nit: "e.g. surveyBasics when filtering surveyOccurrence"?

underlay/src/main/proto/criteriaselector/dataschema/survey.proto
```diff
@@ -0,0 +1,39 @@
...
+option java_outer_classname = "DTSurvey";
+
+import "criteriaselector/key.proto";
+import "criteriaselector/value_data.proto";
+
+// Data for an entity group criteria is a list of selected values.
+message Survey {
+  message Selection {
+    // The key of the selected value, which references a related entity (e.g.
+    // condition for a condition_occurrence).
```
marikomedlock, line 16: nit: "e.g. surveyBasics for a surveOccurrence"?

underlay/src/main/proto/criteriaselector/dataschema/survey.proto
```diff
@@ -0,0 +1,39 @@
...
+    Key question_key = 4;
+
+    // If the selected item is an answer, the visible name of the question it
+    // belongs to.
+    string question_name = 5;
+  }
+  repeated Selection selected = 1;
+
+  // Data for an additional categorical or numeric value associated with the
+  // selection (e.g. a measurement value).
```
marikomedlock, line 37: nit: "e.g. a survey numeric answer value"?

underlay/src/main/java/bio/terra/tanagra/filterbuilder/impl/core/SurveyFilterBuilder.java
```diff
@@ -0,0 +1,279 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.apache.commons.lang3.tuple.Pair;
+
+@SuppressFBWarnings(
+    value = "NP_UNWRITTEN_PUBLIC_OR_PROTECTED_FIELD",
+    justification = "The config and data objects are deserialized by Jackson.")
+public class SurveyFilterBuilder extends FilterBuilder {
```
marikomedlock, line 43: Chunks of this class are pretty similar to the existing `EntityGroupFilterBuilder`. I think it would be better to move those chunks into a shared base class or a utility class. Perhaps using a generic type would be the least amount of up-front work (e.g. `<T> = <DTEntityGroupEntityGroup> or <DTSurvey.Survey>`).

underlay/src/test/java/bio/terra/tanagra/filterbuilder/SurveyFilterBuilderTest.java
```diff
@@ -0,0 +1,510 @@
...
+import bio.terra.tanagra.underlay.serialization.SZService;
+import bio.terra.tanagra.underlay.serialization.SZUnderlay;
+import bio.terra.tanagra.underlay.uiplugin.CriteriaSelector;
+import bio.terra.tanagra.underlay.uiplugin.SelectionData;
+import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class SurveyFilterBuilderTest {
```
marikomedlock, line 39: I think these tests should use actual survey entities instead of the same ones the entity group filter builder tests use. The filter builder tests are just testing that the right filter object is created, I don't think they're running any queries (you can confirm by running them in the `noCloudAccessRequiredTests` mode). So you could use the AoU config files here.

ui/src/criteria/survey.tsx
```diff
@@ -0,0 +1,858 @@
...
+type EntityTreeGridData = TreeGridData<EntityNodeItem>;
+
+// Exported for testing purposes.
+export interface Data {
+  selected: Selection[];
+  valueData: ValueData;
+}
+
+// "survey" plugins are designed to handle medium sized amount of survey data in
+// an optimized fashion.
```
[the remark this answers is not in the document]
tjennison-work: Done.

underlay/src/main/proto/criteriaselector/configschema/survey.proto
tjennison-work: Done.

underlay/src/main/proto/criteriaselector/configschema/survey.proto
tjennison-work: Done.

underlay/src/main/proto/criteriaselector/dataschema/survey.proto
tjennison-work: Done.

underlay/src/test/java/bio/terra/tanagra/filterbuilder/SurveyFilterBuilderTest.java
tjennison-work: These are removed now that the code is mostly shared. We can start adding survey specific tests when the functionality starts to diverge.

underlay/src/main/java/bio/terra/tanagra/filterbuilder/impl/core/EntityGroupFilterBuilderBase.java
```diff
@@ -0,0 +1,364 @@
...
+        });
+
+    selectedIdsPerEntityGroup.forEach(
+        (entityGroup, selectedIds) -> {
+          Collections.sort(selectedIds);
+        });
+    return selectedIdsPerEntityGroup;
+  }
+
+  private List<EntityGroup> selectedEntityGroups(
```
marikomedlock, line 243: nit: Could you add a comment here that this returns the list of entity groups that the user selected ids from, or the list of configured entity groups if the user selected no ids? I know that now from our discussions and reviewing this code, but seems like something I'm going to forget.

marikomedlock (approved):

marikomedlock:
RE survey tests: Now that we have the AoU CT test data indexed in `verily-tanagra-dev`, we could run tests against that, as long as the VUMC folks are okay with that. I imagine they will be, since we got the okay to run tests against the 2019 test data -- but that was several years ago now, so probably good to double check.

underlay/src/main/java/bio/terra/tanagra/filterbuilder/impl/core/EntityGroupFilterBuilderBase.java
tjennison-work: Done.

underlay/src/main/java/bio/terra/tanagra/filterbuilder/impl/core/EntityGroupFilterBuilderBase.java
```diff
@@ -0,0 +1,364 @@
...
+              selectedIdsPerEntityGroup.containsKey(entityGroup)
+                  ? selectedIdsPerEntityGroup.get(entityGroup)
+                  : new ArrayList<>();
+          selectedIds.add(key);
+          selectedIdsPerEntityGroup.put(entityGroup, selectedIds);
+        });
+
+    selectedIdsPerEntityGroup.forEach(
+        (entityGroup, selectedIds) -> {
+          Collections.sort(selectedIds);
```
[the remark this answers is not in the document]
tjennison-work: Done.

underlay/src/main/java/bio/terra/tanagra/filterbuilder/impl/core/EntityGroupFilterBuilderBase.java
```diff
@@ -0,0 +1,364 @@
...
+    return selectedIdsPerEntityGroup;
+  }
+
+  private List<EntityGroup> selectedEntityGroups(
+      Underlay underlay, Map<EntityGroup, List<Literal>> selectedIdsPerEntityGroup) {
+    List<EntityGroup> selectedEntityGroups;
+    if (!selectedIdsPerEntityGroup.isEmpty()) {
+      selectedEntityGroups = new ArrayList<>(selectedIdsPerEntityGroup.keySet());
+    } else {
+      selectedEntityGroups =
```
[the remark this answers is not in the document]
tjennison-work: That's actually not the case because `buildForCohort` and `buildForDataFeature` early exit for criteria with no selection or modifiers. In the modifiers only case this is a change as expected.

That being said, I think changing the behavior is actually correct. Adding an empty condition filter should filter on all people who have a condition rather than doing nothing. I'll probably follow up to remove the early exit but I didn't want to put anything else into this PR if I didn't need to.

tjennison-work:
> RE survey tests: Now that we have the AoU CT test data indexed in `verily-tanagra-dev`, we could run tests against that, as long as the VUMC folks are okay with that. I imagine they will be, since we got the okay to run tests against the 2019 test data -- but that was several years ago now, so probably good to double check.

Sounds good. I'll check with them on Wednesday.

PR #972, opened 2024-08-29
[BENCH-3845] Rename ConceptSet to FeatureSet in backend and backend APIs

dexamundsen:
UI already uses `feature set`. This PR renames ConceptSet to FeatureSet in backend and backend APIs
No change to the DB or DB fields. Leaving it as-in for now in DB

----

Plan is to get it reviewed internally before sharing with larger VUMC team


ui/src/data/source.tsx
```diff
@@ -250,12 +250,6 @@ export type Criteria = {
   predefinedDisplayName?: string;
 };
 
-export type ConceptSet = {
```
dexamundsen, line 253: unused

tjennison-work (approved):

marikomedlock (approved):
Thanks for all the updates!

Please give the VUMC teams (SD/eMERGE & AoU) a heads up about the OpenAPI changes because I think they will both need to make sure their other service (VUMC admin & AoU-RW) is running with a compatible version of the Tanagra client library before upgrading to the most recent Tanagra service.

chenchals:
@dexamundsen The backed is still a disconnect if you rename to FeatureSet.  See https://github.com/DataBiosphere/tanagra/blob/d0e0f5fd534a1eb0aa467c4f6527d1c48de3562a/service/src/main/resources/db/changesets/20230922_initial_schema.yaml#L82-L148

chenchals:
@dexamundsen The backed is still a disconnect if you rename to FeatureSet.  See https://github.com/DataBiosphere/tanagra/blob/d0e0f5fd534a1eb0aa467c4f6527d1c48de3562a/service/src/main/resources/db/changesets/20230922_initial_schema.yaml#L82-L148  This change will be a breaking change for tanagra-admin service

dexamundsen:
> @dexamundsen The backed is still a disconnect if you rename to FeatureSet. See
> 
> https://github.com/DataBiosphere/tanagra/blob/d0e0f5fd534a1eb0aa467c4f6527d1c48de3562a/service/src/main/resources/db/changesets/20230922_initial_schema.yaml#L82-L148
> 
> This change will be a breaking change for tanagra-admin service

he DB table and field names were retained as concept_set (at least for now) to avoid more invasive changes.

dexamundsen:
@chenchals please confirm that we can merge this PR

chenchals (commented):
@dexamundsen  Sure please go ahead and merge these changes.

PR #973, opened 2024-08-29
AoU underlay config changes for QC bugs

dolbeew:
- Age: 
  - Set default age range
  - Update SQL to exclude deceased
  - Update attribute UI plugin to set `dataRanges` when `runtimeSqlFunctionWrapper` is set
- Gender Identity/Race/Sex at Birth - Update SQL to match names with current data apps
- Blood pressure - Update SQL to match AoU counts 
- Add Wear Consent criteria selector
- Update columns and widths for cohort review tables

@tjennison-work This is mostly underlay config changes but tagging you to review the change to the attribute plugin


ui/src/criteria/attribute.tsx
```diff
@@ -58,7 +58,18 @@ interface Data {
...
+        attribute?.displayHintRangeMin &&
+        attribute?.displayHintRangeMax
+          ? [
+              {
+                id: generateId(),
+                min: attribute.displayHintRangeMin,
+                max: attribute.displayHintRangeMax,
+              },
+            ]
+          : [],
```
dolbeew, line 72: Forces the `runtimeSqlFunctionWrapper` to run if it's set. Added so the change to exclude deceased from age counts will be applied on the initial selection

freemabd (approved):
Config looks fine as far as I can tell.

ui/src/criteria/attribute.tsx
tjennison-work: This is a good change but I'm not sure I follow how `runtimeSqlFunctionWrapper` factors in. I think it should always apply the range override even for non-calculated fields. I'd also make a couple small adjustments to handle when only one is set or when one is zero: `isValid(attribute?.displayHintRangeMin) || isValid(attribute?.displayHintRangeMax)`, `import {isValid} from "util/valid";`.

tjennison-work (approved):

ui/src/criteria/attribute.tsx
dolbeew: So if nothing is set for `runtimeSqlFunctionWrapper` then there's no need to make a change to `dataRanges` here since using the display hint ranges _should_ return the same count as an empty array (any value).

dolbeew: But it should work for us either way. I used `runtimeSqlFunctionWrapper` to only set `dataRanges` when we need it (for age in this case), but removing `runtimeSqlFunctionWrapper` from the condition should still give us the correct count

tjennison-work: If there's some bad data (e.g. negative) for something not runtime calculated like age at consent, forcing the min to 18 should exclude that data but currently it's actually still (any value) unless you adjust the slider. Kind of a niche case but I think it's more correct this way.

dolbeew: Good call on the non-runtime cases. I tried using `isValid` but the compiler didn't like it, still thought attribute might be undefined. Also not sure what to use for the other value of the range if only one is set, so I left the `&&` instead of `||`

tjennison-work: Oh, probably just needed to change 68/69 to use `?.` as well. Fine to leave one or both of them undefined since that's the default anyways.

dolbeew: I tried that also but min/max wouldn't accept the type `number | undefined`

tjennison-work: Okay, sounds good as is then. I guess I'm a bit off on how I thought it would work.

PR #974, opened 2024-08-30
Fetch studies by created_by email address

dexamundsen:
- Draft change to fetch only studies created by the current logged in user -  meant as draft / proposal option and not intended as final change.
- Current: UI adds the email domain in createdBy filter field of listStudies
image

- Change: send entire email address in the createdBy filter field of listStudies
image

https://verily.atlassian.net/browse/BENCH-3894

### DIFF
--- a/ui/src/sampleApp/studiesList.tsx
+++ b/ui/src/sampleApp/studiesList.tsx
@@ -42,9 +42,8 @@ export function StudiesList() {
   });
 
   const listStudies = useCallback(async () => {
-    const domain = userState.data?.email?.split("@")[1];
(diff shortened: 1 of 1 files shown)


marikomedlock (commented):
What is the motivation for this change?

If this is for private user studies, then making a UI change is not secure. That should be made by adding a new access control model on the backend, which should also not require any UI changes.

dexamundsen:
@marikomedlock clarified in the description. This PR was meant as draft / proposal option and not intended as final change.

dexamundsen:
Changes will be made in backend

PR #983, opened 2024-09-02
Add cohort.clone

dexamundsen:
Tested locally, includes unit test


service/src/main/java/bio/terra/tanagra/app/controller/ReviewsApiController.java
```diff
@@ -109,8 +109,7 @@ public ResponseEntity<ApiReview> createReview(
                 .size(body.getSize()),
             SpringAuthentication.getCurrentUser().getEmail(),
             entityFilter);
-    return ResponseEntity.ok(
-        toApiObject(createdReview, cohortService.getCohort(studyId, cohortId)));
```
marikomedlock, line 113: I don't think this will work, because the cohort's most recent revision will have updated after creating the review, so you have to re-fetch it.

service/src/main/java/bio/terra/tanagra/service/artifact/CohortService.java
```diff
@@ -127,6 +144,87 @@ public Cohort updateCohort(
...
+            MAX_REVIEW_SIZE);
+    List<Review> clonedReviewList = new ArrayList<>();
+    reviewService
+        .listReviews(authorizedReviewIds, 0, MAX_REVIEW_SIZE)
+        .forEach(
+            review ->
+                clonedReviewList.add(
+                    Review.builder()
+                        .displayName("Copy of: " + review.getDisplayName())
+                        .description("Copy of: " + review.getDescription())
```
marikomedlock, line 207: I don't think we need to update the names/descriptions of all child objects also.

service/src/main/java/bio/terra/tanagra/service/artifact/CohortService.java
```diff
@@ -127,6 +144,87 @@ public Cohort updateCohort(
...
+                        .lastModifiedBy(userEmail)
+                        .isDeleted(review.isDeleted())
+                        .build()));
+
+    // Annotations
+
+
+
+    // TODO: pending clone of annotations
+    cohortDao.createCohort(cohortId, cohortBuilder.build());
```
marikomedlock, line 220: It might be easier to do this all inside the Dao classes. e.g. `cohortDao.cloneCohort(originalCohortId, clonedCohortId, clonedCohortDisplayName, clonedCohortDescription)`. Then you can do it all in a single DB transaction, and just copy the table rows instead of worrying about all the objects.

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2234,6 +2264,14 @@ components:
           items:
             $ref: "#/components/schemas/CriteriaGroupSection"
 
+    CohortCloneInfo:
```
marikomedlock, line 2267: It occurred to me now that we need an optional `destinationStudyId` in these clone endpoints also (data feature set & cohort). Cloning within a study is nice to have, but the immediate use case is calling these endpoints as part of cloning an entire study, so we'd need the destination study ids. Sorry I didn't catch that in the data feature set PR!

dexamundsen: Adding it in a separate PR.

service/src/main/java/bio/terra/tanagra/service/artifact/CohortService.java
dexamundsen: Yes, I plan to wrap it in a single function

service/src/main/java/bio/terra/tanagra/service/artifact/CohortService.java
dexamundsen: Only made this for those with Ids (deep copy) / those i thought might be displayed to the user -- in this case the user can differentiate the original from the clone

marikomedlock: At least in the core Tanagra UI, the child objects (reviews, annotations) are only shows in the context of a cohort. So being able to differentiate the original/cloned cohort is important, but I don't think it matters for the child objects.

service/src/main/resources/api/service_openapi.yaml
dexamundsen: https://github.com/DataBiosphere/tanagra/pull/992

service/src/main/java/bio/terra/tanagra/service/artifact/CohortService.java
dexamundsen: Re-read, and i got your comment now. 
I will talk to you tomorrow, here is my thought - copying all the DB rows for all tables is definitely easier, but I am wondering if this could copy data issues (if any). thought of building objects and then writing DB entries in a single transaction
not a priority - if we want to selectively copy either based on revision, or some other constraint this might help.

service/src/main/java/bio/terra/tanagra/service/artifact/CohortService.java
dexamundsen: sg, was not sure if other objects were displayed anywhere.

service/src/main/java/bio/terra/tanagra/app/controller/ReviewsApiController.java
dexamundsen: reverted

service/src/main/java/bio/terra/tanagra/db/AllArtifactsDao.java
```diff
@@ -0,0 +1,190 @@
...
+          Map<Literal, Integer> primaryEntityIdsToStableIndex =
+              reviewDao.getPrimaryEntityIdsToStableIndex(originalId);
+          primaryEntityIdsToStableIndex.forEach(
+              (entityLiteral, index) ->
+                  primaryEntityParamSets.add(
+                      reviewDao.buildPrimaryEntityParam(
+                          clonedId, entityLiteral.getInt64Val(), index)));
+        });
+    reviewDao.insertReviewRows(reviewParamSets); // fk: cohort(id)
+    reviewDao.insertPrimaryEntityRows(primaryEntityParamSets); // fk: review(id)
```
dexamundsen, line 89: check: use a direct sql statement

dexamundsen: Avoiding this for now to use the existing DB functions. helps avoid extra code maintenance if there are changes in the future

service/src/main/java/bio/terra/tanagra/db/AllArtifactsDao.java
```diff
@@ -0,0 +1,195 @@
...
+          reviewParamSets.add(
+              reviewDao.buildReviewParam(
+                  clonedCohortId,
+                  clonedId,
+                  review.getSize(),
+                  review.getDisplayName(),
+                  review.getDescription(),
+                  userEmail));
+
+          Map<Literal, Integer> primaryEntityIdsToStableIndex =
```
marikomedlock, line 82: Here you're reading all the primary entity ids for all reviews for this cohort into memory. Can you copy the rows in SQL instead with an `INSERT ... SELECT` statement to avoid this? You could try joining on the revision version number to get the new review id, or do it one review at a time.

service/src/main/java/bio/terra/tanagra/db/AllArtifactsDao.java
```diff
@@ -0,0 +1,195 @@
...
+                      userEmail,
+                      revision.getRecordsCount(),
+                      reviewIdMap.get(ogRevisionIdReviewIdMap.get(revision.getId()))));
+            });
+    String sql =
+        "INSERT INTO cohort_revision (cohort_id, id, review_id, version, is_most_recent, is_editable, created_by, last_modified_by, records_count) "
+            + "VALUES (:cohort_id, :id, :review_id, :version, :is_most_recent, :is_editable, :created_by, :last_modified_by, :records_count)";
+    cohortDao.insertRevisionRows(sql, revisionParamSets); // fk: cohort(id), review(id)
+
+    // Criteria, sections, groups, tags - fk: cohort_revision(id)
```
marikomedlock, line 119: ditto above: Can you do this copying in SQL instead of reading all the data into memory first? e.g. 
```
INSERT INTO criteria_group_section (cohort_revision_id, id, display_name, operator, is_excluded, ...)
SELECT :new_revision_id, id, display_name, operator, is_excluded, ...
FROM criteria_group_section
WHERE cohort_revision_id = :old_revision_id;
```

service/src/main/java/bio/terra/tanagra/db/AllArtifactsDao.java
dexamundsen: done

service/src/main/java/bio/terra/tanagra/service/artifact/FeatureSetService.java
```diff
@@ -113,14 +113,14 @@ public FeatureSet cloneFeatureSet(
...
-    if (newDisplayName == null) {
+    if (newDisplayName == null && original.getDisplayName() != null) {
       newDisplayName = original.getDisplayName();
       if (studyId.equals(destinationStudyId)) {
         newDisplayName = "Copy of: " + newDisplayName;
       }
     }
     String newDescription = description;
-    if (newDescription == null) {
+    if (newDescription == null && original.getDescription() != null) {
```
dexamundsen, line 123: thia change avoid new text being 'Copy of: null'

service/src/main/java/bio/terra/tanagra/db/AllArtifactsDao.java
dexamundsen: done

service/src/main/java/bio/terra/tanagra/service/artifact/CohortService.java
```diff
@@ -125,6 +129,49 @@ public Cohort updateCohort(
...
+      newDisplayName = original.getDisplayName();
+      if (studyId.equals(destinationStudyId)) {
+        newDisplayName = "Copy of: " + newDisplayName;
+      }
+    }
+    String newDescription = description;
+    if (newDescription == null && original.getDescription() != null) {
+      newDescription = original.getDescription();
+      if (studyId.equals(destinationStudyId)) {
+        newDescription = "Copy of: " + newDescription;
```
tjennison-work, line 153: Don't append this to the description. The name is enough.

service/src/main/java/bio/terra/tanagra/service/artifact/CohortService.java
```diff
@@ -125,6 +129,49 @@ public Cohort updateCohort(
...
+      String destinationStudyId,
+      @Nullable String displayName,
+      @Nullable String description) {
+    Cohort original = getCohort(studyId, cohortId);
+
+    String newDisplayName = displayName;
+    if (newDisplayName == null && original.getDisplayName() != null) {
+      newDisplayName = original.getDisplayName();
+      if (studyId.equals(destinationStudyId)) {
+        newDisplayName = "Copy of: " + newDisplayName;
```
tjennison-work, line 146: Do we need to worry about the maximum length here?

tjennison-work (commented):
LGTM pending tests.

service/src/main/java/bio/terra/tanagra/service/artifact/CohortService.java
dexamundsen: I searched for a max length but could not find anywhere on any artifact and hence did not add it.
Do we want to enforce it?

service/src/main/java/bio/terra/tanagra/service/artifact/CohortService.java
dexamundsen: sgtm, done

service/src/main/java/bio/terra/tanagra/app/controller/FeatureSetsApiController.java
```diff
@@ -58,8 +58,8 @@ public ResponseEntity<ApiFeatureSet> createFeatureSet(
             FeatureSet.builder()
                 .displayName(body.getDisplayName())
                 .description(body.getDescription())
-                .underlay(body.getUnderlayName()),
-            SpringAuthentication.getCurrentUser().getEmail());
+                .underlay(body.getUnderlayName())
+                .createdBy(SpringAuthentication.getCurrentUser().getEmail()));
```
dexamundsen, line 62: copying the change from cohorts

service/src/test/java/bio/terra/tanagra/service/accesscontrol/AccessControlImplTest.java
```diff
@@ -1,17 +1,12 @@
-package bio.terra.tanagra.service;
+package bio.terra.tanagra.service.accesscontrol;
```
dexamundsen, line 1: moved to access control package

service/build.gradle
```diff
@@ -100,7 +100,7 @@ dependencies {
     testFixturesImplementation "org.springframework:spring-webmvc:${gradle.vSpringWeb}"
 
     // [Application DB] MySQL/MariaDB
-    implementation 'org.mariadb.jdbc:mariadb-java-client:3.4.0'
```
dexamundsen, line 103: There is an open bug in java clients 3.x (see https://mariadb.com/kb/en/unsupported-command-in-prepared-statement-using-the-new-jdbc-driver-303/ and https://jira.mariadb.org/browse/MDEV-26673). Batch updates are not working in 3.x versions. 

we have 2 options - 
- Downgrade the client to latest 2.x (verified on separate branch that this error is resolved). Lock file changes do not show major changes - only the client gets downgraded and most other lib changes are for the better - extra deps are removed. 
- Rewrite our code to only have single row updates. This change is not only for this PR (clone cohort) but others which may be using batch update. Going with single row updates is potentially a blocker for performance reasons, and we don't ideally want to that (bad design/implementation). 

hence not choosing option 2. We can upgrade it back again when the client issue is resolved

service/src/main/java/bio/terra/tanagra/service/artifact/CohortService.java
tjennison-work: Looks like it was https://github.com/DataBiosphere/tanagra/pull/945/files. We just don't want random failures because of it so we should truncate here if it's not happening elsewhere, which it doesn't look like.

dexamundsen: Thanks, added a check. However I will create a followup ticket to simplify it in openApi spec - https://verily.atlassian.net/browse/BENCH-4175

service/build.gradle
dexamundsen: Security - accorinding to maven repo, this version has a security vuln due to version of included logback library (see https://mvnrepository.com/artifact/org.mariadb.jdbc/mariadb-java-client/2.7.12). however we have a newer version of logback already, so i would not expect a new security issue.

dexamundsen: Compatibility: would not expect any. We had a 2.7.* version until the recent java upgrade https://github.com/DataBiosphere/tanagra/blob/6f8f86e40e17f796aa3bac200a38d13ea399f029/service/gradle.lockfile#L269

tjennison-work (approved):

PR #986, opened 2024-09-03
UX updates for VWB

dexamundsen:
Changes from past demos, for product demo tomorrow

- sign in page: change text to “Data Explorer”
- Change return app name in VWB export to data explorer
- Other changes in export page based on UX review

image


marikomedlock (approved):
You'll need to update the internal deployment repo(s) with the new env var names with this change.

service/src/main/java/bio/terra/tanagra/service/export/impl/IpynbFileDownload.java
```diff
@@ -32,12 +31,12 @@ public Type getType() {
...
   @Override
   public String getDefaultDisplayName() {
-    return "Download ipynb file with embedded SQL";
+    return "Download Jupyter Notebook";
   }
 
   @Override
   public String getDescription() {
-    return "Signed URL to an ipynb notebook file that includes SQL to get the list of primary entity instances.";
+    return "Get an IPYNB file for download. SQL comes preinstalled.";
```
marikomedlock, line 39: What does it mean to have SQL preinstalled?

tjennison-work (approved):

service/src/main/java/bio/terra/tanagra/service/export/impl/IpynbFileDownload.java
Matthew-Antchak: > What does it mean to have SQL preinstalled?

My assumption is that it includes a SQL kernel and dependencies

marikomedlock: Do you know if this is true of the existing export implementation?

The generated ipynb file contains the SQL query and the idea is to make it easy to run that query against BigQuery. The template file we use is [here](https://github.com/DataBiosphere/tanagra/blob/main/service/src/main/resources/export/notebook_template.ipynb).

dexamundsen: @Matthew-Antchak how about 'Get an IPYNB file for download`. From the template - it uses pandas, not SQL kernel.

Matthew-Antchak: Yes, in that case let's just omit the second sentence about SQL @dexamundsen

PR #987, opened 2024-09-06
Add SLF4J implementation dependency.

marikomedlock:
Adding an explicit dependency on an SLF4J implementation library.

This was causing tests to not log output with the below warning:
```
BQListQueryResultsTest > attributeField() STANDARD_ERROR
    SLF4J(W): No SLF4J providers were found.
    SLF4J(W): Defaulting to no-operation (NOP) logger implementation
    SLF4J(W): See https://www.slf4j.org/codes.html#noProviders for further details.
```


underlay/build.gradle
```diff
@@ -25,6 +25,7 @@ dependencies {
 
     implementation "org.apache.commons:commons-text:${gradle.vApacheCommonsText}"
     testFixturesImplementation "org.slf4j:slf4j-api:${gradle.vSlf4jApi}"
+    implementation 'org.slf4j:slf4j-simple:2.0.16'
```
dexamundsen, line 28: Is this the same version as the one in the variable in previous line

dexamundsen (approved):

underlay/build.gradle
marikomedlock: No, different.

PR #994, opened 2024-09-10
Add hierarchy is leaf filter.

marikomedlock:
This is to support fetching all non-leaf nodes in the AoU survey entities, for global search where we want to exclude answers.


tjennison-work (approved):
This could also be a child count filter, since that's what you're doing under the hood anyways.

marikomedlock:
> This could also be a child count filter, since that's what you're doing under the hood anyways.

Yes, I considered that too and ended up going with the isLeaf for now since that's the only use case. Do you have another use case for filtering on number of children? If so, happy to change this now, otherwise I think I'll hold off until we have a use case.

tjennison-work:
> > This could also be a child count filter, since that's what you're doing under the hood anyways.
> 
> Yes, I considered that too and ended up going with the isLeaf for now since that's the only use case. Do you have another use case for filtering on number of children? If so, happy to change this now, otherwise I think I'll hold off until we have a use case.

No, that would just be the generic version if we did come up with another use case. Fine with me as is too.

PR #995, opened 2024-09-10
Variant search config.

marikomedlock:
First pass at variant search config for AoU 2023 Q3R2 CT underlay.

As part of this PR, I also ran indexing on the `variant` entity and `variantPerson` entity group on AoU CT test data in the Verily dev project.


tjennison-work (commented):
Similar to surveys, we won't want to use the existing entity group config/data protos, so if they're not anxious to check this in we might want to wait until we have new ones.

marikomedlock:
Per offline discussion, moved config files into separate `_testonly` directories, so we can iterate on these config files across multiple PRs without modifying the underlays that AoU is currently using.

tjennison-work (approved):

PR #999, opened 2024-09-12
Util to generateId, remove extra userEmail args

dexamundsen:
- create ServiceUtils.newArtifactId() to wrap id creation
- remove --info option from github workflows. It removed unnecessary console info
- use lastModifiedBy = createdBy during artifact.build. This ensured that the values are same during object creation.
- Add createdBy to the cohortbuilder itself instead of passing it as additional param


.github/workflowsForDownstreamRepo/regression-test-downstream-repo.yaml
```diff
@@ -53,7 +53,7 @@ jobs:
         mkdir -p rendered/
         echo "$GHA_SA_KEY" > rendered/gha_sa_key.json
         export GOOGLE_APPLICATION_CREDENTIALS=$PWD/rendered/gha_sa_key.json
-        ./tanagra/gradlew -p tanagra service:regressionTests -PregressionTestUnderlays="${UNDERLAYS:-$DEFAULT_UNDERLAYS}" --info --scan
+        ./tanagra/gradlew -p tanagra service:regressionTests -PregressionTestUnderlays="${UNDERLAYS:-$DEFAULT_UNDERLAYS}" --scan
```
dexamundsen, line 56: Removes info level console log, makes it easier to load and debug the run logs

indexer/src/main/java/bio/terra/tanagra/indexing/JobSequencer.java
```diff
@@ -497,14 +497,11 @@ public enum JobExecutor {
...
         List<SequencedJobSet> jobSets, boolean isDryRun, IndexingJob.RunType runType) {
-      switch (this) {
-        case SERIAL:
-          return new SerialRunner(jobSets, isDryRun, runType);
-        case PARALLEL:
-          return new ParallelRunner(jobSets, isDryRun, runType);
-        default:
-          throw new IllegalArgumentException("Unknown JobExecution enum type: " + this);
-      }
+      return switch (this) {
```
dexamundsen, line 500: enhanced switch from java 17

service/src/main/java/bio/terra/tanagra/app/controller/CohortsApiController.java
```diff
@@ -67,8 +67,8 @@ public ResponseEntity<ApiCohort> createCohort(String studyId, ApiCohortCreateInf
             Cohort.builder()
                 .displayName(body.getDisplayName())
                 .description(body.getDescription())
-                .underlay(body.getUnderlayName()),
-            SpringAuthentication.getCurrentUser().getEmail());
+                .underlay(body.getUnderlayName())
+                .createdBy(SpringAuthentication.getCurrentUser().getEmail()));
```
dexamundsen, line 71: Useremail is only used for this field in nested functions. Add it to builder and remove the param

service/src/main/java/bio/terra/tanagra/db/CohortDao.java
```diff
@@ -327,10 +327,7 @@ public String createNextRevision(
             .setIsMostRecent(true)
             .version(cohort.getMostRecentRevision().getVersion() + 1)
             .createdBy(userEmail)
-            .lastModifiedBy(userEmail)
```
dexamundsen, line 330: lastmodifiedby is set in build() during creation if its is null

service/src/main/java/bio/terra/tanagra/service/artifact/model/Study.java
```diff
@@ -192,18 +192,21 @@ public Builder isDeleted(boolean isDeleted) {
     }
 
     public Study build() {
+      // true if the id is empty or null
+      if (StringUtils.isEmpty(id)) {
```
dexamundsen, line 196: moved up

indexer/src/main/java/bio/terra/tanagra/indexing/JobSequencer.java
tjennison-work: In general, it's not a good idea to make this sort of sweeping cleanup in a PR that is doing other, unrelated things. It makes this change harder to review because I don't know if any differences I come across are intended or not, and can cause cherry-picks, rollbacks, etc. to be more difficult. Don't bother pulling it out of this PR but something to keep in mind in the future.

service/src/main/java/bio/terra/tanagra/app/controller/CohortsApiController.java
```diff
@@ -195,36 +195,36 @@ public ResponseEntity<ApiInstanceCountList> queryCohortCounts(
...
-        break;
-      case NUM_DAYS_AFTER:
-        operator = BooleanAndOrFilter.LogicalOperator.OR;
-        joinOperator = JoinOperator.NUM_DAYS_AFTER;
-        break;
-      default:
-        throw new IllegalArgumentException(
-            "Unknown criteria group section operator: " + apiObj.getOperator());
-    }
+    JoinOperator joinOperator =
```
tjennison-work, line 198: I don't think this particular conversion is an upgrade. It's not shorter and now there's two different styles instead of one.

tjennison-work (approved):

service/src/main/java/bio/terra/tanagra/service/ServiceUtils.java
```diff
@@ -0,0 +1,11 @@
+package bio.terra.tanagra.service;
+
+import org.apache.commons.lang3.RandomStringUtils;
+
+public class ServiceUtils {
```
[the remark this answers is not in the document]
dexamundsen: reason added - yes, its a wrapper class. but enables us to change it everywhere is needed instead of individual changes, and controls the length etc. 
i am not strongly opinionated. Can remove if insisted.

service/src/main/java/bio/terra/tanagra/app/controller/CohortsApiController.java
dexamundsen: from my understanding it drills down to how it is compiled for execution. was introduced in java 17

service/src/main/java/bio/terra/tanagra/db/CohortDao.java
```diff
@@ -327,10 +327,7 @@ public String createNextRevision(
...
             .id(null) // Builder will generate a new id.
-            .created(null)
-            .lastModified(null)
```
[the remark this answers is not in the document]
dexamundsen: I will add a comment for the time fields - other artifacts have this pattern where time is skipped, hence had removed it here. we dont depend on the default per say

indexer/src/main/java/bio/terra/tanagra/indexing/JobSequencer.java
dexamundsen: Sure -- put these changes together since all of them are cleanup / optimizations. the entire PR is low priority and we could revert the entire thing. hence put them together. dao changes were left separate because that was not a cleanup, but refactored (for later use in next pr)

service/src/main/java/bio/terra/tanagra/app/controller/CohortsApiController.java
dexamundsen: Correct, the activity log used cohort.getCreatedBy(), which might also be more descriptive imo

service/src/main/java/bio/terra/tanagra/service/UnderlayService.java
```diff
@@ -65,23 +64,19 @@ public UnderlayService(
...
       }
       List<String> authorizedNames =
-          authorizedIds.getResources().stream()
-              .map(ResourceId::getUnderlay)
-              .collect(Collectors.toList());
+          authorizedIds.getResources().stream().map(ResourceId::getUnderlay).toList();
       return underlayCache.values().stream()
           .map(CachedUnderlay::getUnderlay)
           .filter(underlay -> authorizedNames.contains(underlay.getName()))
-          .collect(Collectors.toUnmodifiableList());
```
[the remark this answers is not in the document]
dexamundsen: same, reverting -- had used the same pattern (?) as others

service/src/main/java/bio/terra/tanagra/app/controller/CohortsApiController.java
tjennison-work: I'm not sure what you mean. We're not worried about the performance or anything like that here so how it compiles isn't important. In general, the new syntax cleaner and has less boilerplate in a lot of cases. Most of the replacements in the PR are a clear improvement. In this particular case though, because of the two variables are being assigned, changing the way one is assigned just makes it less clear the behavior for both variables is actually the same.

dexamundsen: Not highly opinionated on it, and on second look agree with you - so no concern reverting it. 
Adding this revert to https://github.com/DataBiosphere/tanagra/pull/1001

service/src/main/java/bio/terra/tanagra/service/ServiceUtils.java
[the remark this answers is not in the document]
dexamundsen: reverted in https://github.com/DataBiosphere/tanagra/pull/1001

service/src/main/java/bio/terra/tanagra/db/CohortDao.java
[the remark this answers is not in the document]
dexamundsen: nvm. this is for the most recent revision, so adding it back to ensure documentation that this is overridden

.github/workflows/regression-test.yaml
```diff
@@ -64,7 +64,7 @@ jobs:
       env:
         TEST_PROJECT_SA_KEY: ${{ secrets.TEST_PROJECT_SA_KEY }}
     - name: Gradle Run Regression Tests Only
-      run: ./gradlew service:regressionTests -PregressionTestUnderlays=cmssynpuf,aouSR2019q4r4 --info --scan
+      run: ./gradlew service:regressionTests -PregressionTestUnderlays=cmssynpuf,aouSR2019q4r4 --scan
```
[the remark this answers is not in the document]
dexamundsen: A comparison of without - https://github.com/DataBiosphere/tanagra/actions/runs/10859538266/job/30139008330#step:8:1

and with - 
https://github.com/DataBiosphere/tanagra/actions/runs/10857889854/job/30135081427#step:8:1

For now keeping it out since all of the extra logs are less about the application tests etc and more about the gradle build process, which is not needed regularly. 
keeping logs concide really helps with debugging

PR #1000, opened 2024-09-12
AoU underlay config updates for PFHH survey

dolbeew:
Update SQL for PFHH entity and entity group to use `prep_pfhh_observation` table instead of `observation`


freemabd (approved):
Thanks!

PR #1001, opened 2024-09-12
Refactor dao classes to extract param build, row insert

dexamundsen:
- Will help reuse them in clone cohort
- also contains changes from PR comments : https://github.com/DataBiosphere/tanagra/pull/999


service/src/main/java/bio/terra/tanagra/db/AnnotationDao.java
```diff
@@ -298,4 +277,79 @@ public List<AnnotationValue.Builder> getAllAnnotationValues(String cohortId) {
...
+  }
+
+  MapSqlParameterSource buildEnumValueParam(String cohortId, String keyId, String enumValue) {
+    return new MapSqlParameterSource()
+        .addValue("cohort_id", cohortId)
+        .addValue("annotation_key_id", keyId)
+        .addValue("enum", enumValue);
+  }
+
+  MapSqlParameterSource buildValueParam(
```
dexamundsen, line 336: made them package private to be callable from other dao classes

service/src/main/java/bio/terra/tanagra/db/AnnotationDao.java
```diff
@@ -78,37 +79,21 @@ public AnnotationDao(NamedParameterJdbcTemplate jdbcTemplate) {
...
-    MapSqlParameterSource params =
-        new MapSqlParameterSource()
-            .addValue("cohort_id", cohortId)
-            .addValue("id", annotationKey.getId())
-            .addValue("display_name", annotationKey.getDisplayName())
-            .addValue("description", annotationKey.getDescription())
-            .addValue("data_type", annotationKey.getDataType().name());
-    int rowsAffected = jdbcTemplate.update(sql, params);
-    LOGGER.debug("CREATE annotation key rowsAffected = {}", rowsAffected);
+    MapSqlParameterSource keyParamSets =
```
dexamundsen, line 82: general patter: build param, then insert rows

.github/workflows/regression-test.yaml
```diff
@@ -64,7 +64,7 @@ jobs:
       env:
         TEST_PROJECT_SA_KEY: ${{ secrets.TEST_PROJECT_SA_KEY }}
     - name: Gradle Run Regression Tests Only
-      run: ./gradlew service:regressionTests -PregressionTestUnderlays=cmssynpuf,aouSR2019q4r4 --scan
+      run: ./gradlew service:regressionTests -PregressionTestUnderlays=cmssynpuf,aouSR2019q4r4 --info --scan
```
dexamundsen, line 67: A comparison of without - https://github.com/DataBiosphere/tanagra/actions/runs/10859538266/job/30139008330#step:8:1

and with - 
https://github.com/DataBiosphere/tanagra/actions/runs/10857889854/job/30135081427#step:8:1

For now keeping it out since all of the extra logs are less about the application tests etc and more about the gradle build process, which is not needed regularly. 
keeping logs concide really helps with debugging

tjennison-work (approved):

PR #1010, opened 2024-09-18
Remove duplicate display fields for AoU source queries

dolbeew:
Remove `displayFieldName` attributes where the field is the same as `name`. Was causing an 'Unrecognized name' error in cohort review.


freemabd (approved):
Thanks.

PR #1011, opened 2024-09-18
DT-888 adding config fix for entities that do not have a primary key in the source table.

freemabd:
Adding config fix for entities that do not have a primary key in the source table. This change removes the outer source where clause and replaces it with the index where clause. 

Example:
Original SQL - expects the source table to have a primary key which doesn't exist in this case. This was making it impossible to config source queries in this case.
```
SELECT
  attributes...
FROM
  ${dataset}.activity_summary AS st
WHERE
  st.row_id IN (
  SELECT
    id
  FROM
    ${dataset}.T_ENT_activitySummary
  WHERE
    person_id IN (.... data feature sql....)
)
``` 

Revised SQL - using the where clause from index sql which makes the cleaner and it's not dependent on primary keys
```
SELECT
  attributes...
FROM
  ${dataset}.activity_summary AS st
WHERE
  person_id IN (.... data feature sql....)
```

### DIFF
--- a/underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQQueryRunner.java
+++ b/underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQQueryRunner.java
@@ -616,7 +616,7 @@ public class BQQueryRunner implements QueryRunner {
 
     // SELECT [select fields] FROM [source table]
     // JOIN [display table] ON [display join field]
-    // WHERE [source id field] IN [inner query against index data]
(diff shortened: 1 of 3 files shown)


freemabd:
@dexamundsen @tjennison-work I'm digging a little deeper and I don't think this change will work. When building a more complex cohort that includes attributes that exist in the index but don't exist in source it breaks. At this point I'm not sure how to get source queries configured for tables that don't have primary keys. Any ideas?

freemabd:
Closing PR since it's doesn't solve the problem correctly.

PR #1020, opened 2024-09-30
Add option to get all studies

dolbeew:
Not sure this is the best way to handle this but we need a way to get the list of studies for all users for the SD/eMERGE admin pages. I've added an optional flag to the `listStudies` endpoint here but could also just do a separate endpoint.

### DIFF
--- a/service/src/main/java/bio/terra/tanagra/app/controller/StudiesApiController.java
+++ b/service/src/main/java/bio/terra/tanagra/app/controller/StudiesApiController.java
@@ -88,8 +88,9 @@ public class StudiesApiController implements StudiesApi {
       Integer offset,
       Integer limit) {
     ResourceCollection authorizedStudyIds =
-        allStudies ? ResourceCollection.allResourcesAllPermissions(STUDY, null) :
-            accessControlService.listAuthorizedResources(
+        allStudies
+            ? ResourceCollection.allResourcesAllPermissions(STUDY, null)
+            : accessControlService.listAuthorizedResources(
                 SpringAuthentication.getCurrentUser(),
                 Permissions.forActions(STUDY, READ),
                 offset,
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/tanagra/app/controller/StudiesApiController.java
```diff
@@ -84,14 +84,17 @@ public ResponseEntity<ApiStudyList> listStudies(
...
       Integer offset,
       Integer limit) {
     ResourceCollection authorizedStudyIds =
-        accessControlService.listAuthorizedResources(
-            SpringAuthentication.getCurrentUser(),
-            Permissions.forActions(STUDY, READ),
-            offset,
-            limit);
+        allStudies
+            ? ResourceCollection.allResourcesAllPermissions(STUDY, null)
```
dexamundsen, line 92: A bit concerned bypassing the access control altogether - an unauthorized caller can get studies they should not have access to by passing this flag. This would affect all access control models in all deployments. 

- is there a particular admin user that would call this api to get all studies? 
- how about a new function in accessControlService.listAllResources(adminUser, permissions, offset, limit). This could check that the user requesting the list is an admin user and would return ALL resources or all users (subject to offset / limit)? The default behavior in the interface could be return none, and access control model impl can override it.

dexamundsen: Another very minor issue is that we also ignore the offset / limit. can the caller handler the large response of all studies, assuming that backend service will not have issues with this large data? verily unlikely that we run out of memory for this API, but putting it out there.

Also, could there be a case where the sd/emerge admin caller wants the response paginated? 
instead, is it feasible to send a large limit (int max) in the API request? Another alternate is to set limit=int.Max if it is not set in the request, and its is an admin caller (see comment on access control for admin caller)

This way we retain the support for offset / limit.

dexamundsen (commented):
agree, a separate API is not needed. 
also, any applicable tests could be added

service/src/main/java/bio/terra/tanagra/app/controller/StudiesApiController.java
dolbeew: After discussing with @chenchals, we agree that handling this in the admin api is the better option. Will either modify `listAuthorizedResources` to check if user is admin or add new endpoint for that check. Leaving this pr open for now in case core Tanagra changes will be needed .

PR #1021, opened 2024-09-30
Variant search UI and config

tjennison-work:
First pass at the variant search UI and config. The core functionality is there but not some of the details. Notably missing:
- Backend filter generation
- Exclusions
- Select all stats

### DIFF
--- a/ui/src/data/source.tsx
+++ b/ui/src/data/source.tsx
@@ -26,13 +26,18 @@ export type SearchEntityGroupOptions = {
   query?: string;
   parent?: DataKey;
   limit?: number;
+  pageSize?: number;
   hierarchy?: boolean;
   fetchAll?: boolean;
   isLeaf?: boolean;
+  pageMarker?: string;
+  filters?: tanagra.Filter[];
 };
 
 export type SearchEntityGroupResult = {
   nodes: EntityNode[];
+  pageMarker?: string;
+  total: number;
 };
 
 export type SearchGroupingOptions = {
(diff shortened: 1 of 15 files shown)


underlay/src/main/resources/config/criteria/aouSC2023Q3R2_testonly/criteriaselector/variant/selector.json
```diff
@@ -8,8 +8,8 @@
     "tags": null
   },
   "filterBuilder": "core.EntityGroupFilterBuilder",
```
dexamundsen, line 10: Should this be changes to filterableGroupFilterBuilder ? (i assume we create a new one at the backend with this name)

or should this change after the backend change since the service would look for it during start up

ui/src/data/source.tsx
```diff
@@ -1713,8 +1723,21 @@ function processAttributes(
 ) {
   for (const k in attributes) {
     const v = attributes[k];
-    const value = dataValueFromLiteral(v.value);
-    obj[k] = v.display ?? value;
+    let value: DataValue = null;
+    let display = v.display;
+
+    if (v.isRepeatedValue) {
+      // TODO(tjennison): Consider how best to support repeated values beyond
```
dexamundsen, line 1730: nit: consider jira ticket ids for TODOs

dexamundsen (approved):

underlay/src/main/resources/config/criteria/aouSC2023Q3R2_testonly/criteriaselector/variant/selector.json
tjennison-work: Yes, once it exists in the backend so the UI can be functional in the interim. It's only referenced in the test underlay for now.

ui/src/data/source.tsx
tjennison-work: I feel like that's better for work we actually intend to do. This is more just a comment for anyone who wonders why these values are hardcoded in the future. Maybe a plain comment would have been the right choice..

PR #1024, opened 2024-10-03
Simplify dependency locking and version def

dexamundsen:
- Simplify command for dependency locking by wrapping sub-project tasks within a root task
- Define plugin versions in build.gradle, ensures same version is used across projects
- use Buildscript variables instead of gradle variables. The former is build structure agnostic
- Bump spring boot version from 3.3.3 to 3.3.4; dep management from 1.1.5 to 1.1.6


annotationProcessor/build.gradle
```diff
@@ -3,8 +3,8 @@ plugins {
 
     // These plugins are needed to resolve versions of spring dependencies
     // transitively added by terra-common-lib
-    id 'org.springframework.boot' version '3.3.3'
-    id 'io.spring.dependency-management' version '1.1.5'
+    id 'io.spring.dependency-management' version "${vSpringDepMgt}"
```
dexamundsen, line 6: Versions defined as build script variables in build.gradle.

annotationProcessor/build.gradle
```diff
@@ -39,9 +39,9 @@ dependencies {
         implementation 'com.nimbusds:nimbus-jose-jwt:9.37.3'
     }
 
-    implementation "org.slf4j:slf4j-api:${gradle.vSlf4jApi}"
```
dexamundsen, line 42: Build script variables do not need gradle qualifier, hence removed in all references

build.gradle
```diff
@@ -1,22 +1,35 @@
...
+        vSwaggerCodegen = '3.0.57'
+        vPicocliVersion = '4.7.6'
+        vJunit5 = '5.10.2'
+    }
     repositories {
         mavenCentral()
     }
 }
 
-gradle.ext {
```
dexamundsen, line 8: Moved up.

build.gradle
```diff
@@ -1,22 +1,35 @@
 // Required by spotless. https://github.com/diffplug/spotless/issues/747
 buildscript {
+    ext {
+        // plugin versions
+        vSpringBoot = '3.3.4'
```
dexamundsen, line 5: spring boot bumped from 3.3.3 to 3.3.4
dep management plugin bumped from 1.1.5 to 1.1.6

build.gradle
```diff
@@ -1,22 +1,35 @@
...
-    vApacheCommonsText = '1.12.0'
-    vSlf4jApi = '2.0.16'
-    vSpotBugs = '4.8.5'
-    vJackson = '2.17.2'
-    vJersey = '3.1.8' // Java 17 compatible
-    vSwaggerAnnotations = '2.2.22'
-    vSwaggerCodegen = '3.0.57'
-    vPicocliVersion = '4.7.6'
-    vJunit5 = '5.10.2'
+tasks.named('dependencies') {
```
dexamundsen, line 29: This configures a root task dependencies that depends on below subtasks. hence auto triggered.

tjennison-work (approved):
This is a good improvement. It's surprising to me that Gradle doesn't doesn't have a more automatic way to handle this but I don't have much experience with it.

PR #1035, opened 2024-10-10
bench-4235: Update Hints api to accept filter param

dexamundsen:
local testing payloads for 'variant' entity 
`
{ "filter": { "filterType": "ATTRIBUTE", "filterUnion": { "attributeFilter": { "attribute": "rs_number", "operator": "IN", "values": [ { "dataType": "STRING", "valueUnion": { "stringVal": "rs10062" } } ] } } } } 
`

`
 { "filter": { "filterType": "ATTRIBUTE", "filterUnion": { "attributeFilter": { "attribute": "variant_id", "operator": "EQUALS", "values": [ { "dataType": "STRING", "valueUnion": { "stringVal": "13-10062-G-D" } } ] } } } } \
`

`
 { "filter": { "filterType": "TEXT", "filterUnion": { "textFilter": { "matchType": "EXACT_MATCH", "attribute": "gene", "text": "gene" } } } }
 `

sample generated queries
```
SELECT MIN(year_of_birth) AS minVal, MAX(year_of_birth) AS maxVal FROM (SELECT year_of_birth FROM `verily-tanagra-dev.cmssynpuf_index_060724`.ENT_person WHERE year_of_birth IN (1930, 1932));
SELECT MIN(age) AS minVal, MAX(age) AS maxVal FROM (SELECT CAST(FLOOR(TIMESTAMP_DIFF(CURRENT_TIMESTAMP(), age, DAY) / 365.25) AS INT64) AS age FROM `verily-tanagra-dev.cmssynpuf_index_060724`.ENT_person WHERE year_of_birth IN (1930, 1932));
SELECT race AS enumVal, T_DISP_race AS enumDisp, COUNT(*) AS enumCount FROM `verily-tanagra-dev.cmssynpuf_index_060724`.ENT_person WHERE year_of_birth IN (1930, 1932) GROUP BY enumVal, enumDisp;
SELECT gender AS enumVal, T_DISP_gender AS enumDisp, COUNT(*) AS enumCount FROM `verily-tanagra-dev.cmssynpuf_index_060724`.ENT_person WHERE year_of_birth IN (1930, 1932) GROUP BY enumVal, enumDisp;
SELECT ethnicity AS enumVal, T_DISP_ethnicity AS enumDisp, COUNT(*) AS enumCount FROM `verily-tanagra-dev.cmssynpuf_index_060724`.ENT_person WHERE year_of_birth IN (1930, 1932) GROUP BY enumVal, enumDisp;
```


service/src/main/resources/api/service_openapi.yaml
```diff
@@ -1617,20 +1619,24 @@ components:
...
+            (e.g. condition occurrences where condition id=11). Otherwise, 
+            we return hints computed across all instances (e.g. all persons).
           type: object
           properties:
             name:
               type: string
               description: Related entity name
             id:
               $ref: "#/components/schemas/Literal"
+        entityQuery:
```
tjennison-work, line 1638: You could shorten this to `query`. There isn't anything else for it to refer to.

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -1617,20 +1619,24 @@ components:
 
     HintQuery:
       type: object
-      description: Get display hints
+      description: |
+        Get display hints. Only one of relatedEntity and entityQuery may be provided
```
tjennison-work, line 1623: It would be good to describe what passing a query actual does here since it's quite different from the other behavior. I don't think you can add it to the `entityQuery` parameter because it's a ref but even better there if you actually can.

tjennison-work (commented):
A few small things but otherwise looks good.

service/src/main/resources/api/service_openapi.yaml
dexamundsen: sure.

service/src/main/resources/api/service_openapi.yaml
dexamundsen: It cannot be added to the individual ref property . adding here

annotationProcessor/gradle.lockfile
```diff
@@ -1,6 +1,7 @@
 # This is a Gradle generated file for dependency locking.
 # Manual edits can break the build and are not advised.
 # This file is expected to be part of source control.
+args4j:args4j:2.33=productionRuntimeClasspath,runtimeClasspath,testFixturesRuntimeClasspath,testRuntimeClasspath
```
dexamundsen, line 4: Changes due to adding indexer as a dependency to service. 
(annotationProcessor already depends on service)

underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQQueryRunner.java
```diff
@@ -188,45 +188,8 @@ public CountQueryResult run(CountQueryRequest countQueryRequest) {
 
   @Override
   public HintQueryResult run(HintQueryRequest hintQueryRequest) {
-    // Build the SQL query.
-    StringBuilder sql = new StringBuilder();
```
dexamundsen, line 192: split to separate function below to improve readability

underlay/src/main/java/bio/terra/tanagra/api/query/list/OrderBy.java
```diff
@@ -26,11 +26,11 @@ private OrderBy(
     this.isRandom = isRandom;
   }
 
-  public ValueDisplayField getEntityField() {
+  public @Nullable ValueDisplayField getEntityField() {
```
dexamundsen, line 29: no op for functionality, but improves error handling

underlay/src/main/java/bio/terra/tanagra/api/query/hint/HintQueryRequest.java
```diff
@@ -9,9 +10,10 @@
 public class HintQueryRequest {
   private final Underlay underlay;
   private final Entity hintedEntity;
-  private final @Nullable Entity relatedEntity;
-  private final @Nullable Literal relatedEntityId;
-  private final @Nullable EntityGroup entityGroup;
+  private final Entity relatedEntity;
+  private final Literal relatedEntityId;
+  private final EntityGroup entityGroup;
+  private final SqlQueryRequest sqlQueryRequest;
```
dexamundsen, line 16: SqlQueryRequest is added as a new field

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -1646,20 +1648,25 @@ components:
...
+            (e.g. condition occurrences where condition id=11). Otherwise, 
+            we return hints computed across all instances (e.g. all persons).
           type: object
           properties:
             name:
               type: string
               description: Related entity name
             id:
               $ref: "#/components/schemas/Literal"
+        filter:
```
dexamundsen, line 1668: @tjennison-work query had a filter field as well. Extra fields like return attributes are skipped by using this

service/src/main/java/bio/terra/tanagra/service/UnderlayService.java
```diff
@@ -89,11 +102,33 @@ public Underlay getUnderlay(String underlayName) {
     return underlayCache.get(underlayName).getUnderlay();
   }
 
-  public Optional<HintQueryResult> getEntityLevelHints(String underlayName, String entityName) {
+  private static class CachedUnderlay {
```
dexamundsen, line 105: moved up, no change

service/src/main/java/bio/terra/tanagra/service/UnderlayService.java
```diff
@@ -109,18 +144,7 @@ public CountQueryResult runCountQuery(
       Integer limit,
       PageMarker pageMarker,
       Integer pageSize) {
-    // Get the entity level hints, either via query or from the cache.
```
dexamundsen, line 112: split to separate function

service/src/main/java/bio/terra/tanagra/app/controller/UnderlaysApiController.java
```diff
@@ -237,30 +238,46 @@ public ResponseEntity<ApiDisplayHintList> queryHints(
         ResourceId.forUnderlay(underlayName));
     Underlay underlay = underlayService.getUnderlay(underlayName);
     Entity entity = underlay.getEntity(entityName);
+    HintQueryResult hintQueryResult;
 
-    boolean isEntityLevelHints = body == null || body.getRelatedEntity() == null;
-    HintQueryRequest hintQueryRequest;
-    if (isEntityLevelHints) {
-      hintQueryRequest = new HintQueryRequest(underlay, entity, null, null, null, false);
```
dexamundsen, line 244: fetch from cache if it exists

underlay/src/main/java/bio/terra/tanagra/query/bigquery/BQQueryRunner.java
```diff
@@ -168,13 +173,13 @@ public CountQueryResult run(CountQueryRequest countQueryRequest) {
                         }
                         fieldValues.put(valueDisplayField, valueDisplay);
                       });
-              long count =
```
dexamundsen, line 171: can cause npe during unboxing

indexer/src/main/java/bio/terra/tanagra/indexing/job/bigquery/WriteEntityLevelDisplayHints.java
```diff
@@ -407,33 +418,49 @@ private List<Pair<ValueDisplay, Long>> computeEnumHintForValueDisplay(
...
+    if (attribute.isValueDisplay()) {
+      selectEnumCountSql +=
+          ", " + SqlQueryField.of(attrDispField, ENUM_DISP_ALIAS).renderForGroupBy(null, true);
+    }
+
     LOGGER.info("SQL enum count: {}", selectEnumCountSql);
+    return selectEnumCountSql;
+  }
+
+  private List<Pair<Literal, Long>> computeEnumHintForRepeatedStringValue(
```
tjennison-work, line 460: I maybe don't have a complete enough understanding of the code to understand how everything fits together, but it seems like making `computeEnumHintForRepeatedStringValue` and the others public rather than `buildEnumHintForRepeatedStringValueSql` would be a cleaner interface. It seems like this creates an awkward interface where the caller is responsible for unnecessary details (e.g. reading the results). It looks like the main issue is how the SQL is actually run and perhaps that could be solved by passing in a function that's responsible for the execution but the processing could be done here regardless? Optionally there could public wrappers here (e.g. `computeEnumHintForRepeatedStringValue[Indexing|Runtime]` which supply the execution function so those don't need to be public either?

tjennison-work (commented):
As discussed, please add a test at least to the actual SQL generation.

service/src/main/java/bio/terra/tanagra/app/controller/UnderlaysApiController.java
```diff
@@ -237,30 +238,41 @@ public ResponseEntity<ApiDisplayHintList> queryHints(
...
-    } else { // isInstanceLevelHints
+    if (body.getFilter() != null) { // dynamic entity level hints
```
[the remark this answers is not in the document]
dexamundsen: No, the code generated from open api validates that the body is valid. Caller gets a bad request if body is null

service/src/main/java/bio/terra/tanagra/service/UnderlayService.java
```diff
@@ -1,16 +1,34 @@
 package bio.terra.tanagra.service;
 
+import static bio.terra.tanagra.indexing.job.bigquery.WriteEntityLevelDisplayHints.ENUM_COUNT_ALIAS;
```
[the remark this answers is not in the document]
dexamundsen: As a practice this was recommended against in other VWB services - purely coding convention. 
Only downside of wildcard imports is that it might clutter local namespace. 
No concern changing it

ui/public/manifest.json
```diff
@@ -3,7 +3,7 @@
   "name": "Portable Dataset Builder UI",
   "icons": [
     {
-      "src": "favicon.ico",
+      "src": "favicon.png",
       "sizes": "64x64 32x32 24x24 16x16",
       "type": "image/x-icon"
```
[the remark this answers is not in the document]
dexamundsen: remove the entire icons block?

indexer/src/main/java/bio/terra/tanagra/indexing/job/bigquery/WriteEntityLevelDisplayHints.java
dexamundsen: Probably the function names add the the confusion.
The challenge here is that indexing and the user runtime (current flow) handle outputs differently. They are similar, but optimized for the required end result. I tried to avoid too many changes in the indexing flow earlier. With the filter issue fixed (the where clause added on primary entity is removed), i can add that extra logic in the current changes with minimal perf addition

dexamundsen: looking into it further - changes will require quite a bit of refactor in both indexer and underlay if we want to share the logic to process rows. 

- indexer code is written for big query, hence it does not use any of the Tanagra platform agnostic data models. Instead it uses BQ native data types. example: tableResult is a BQ construct that is converted to SqlQueryResult. 
- underlay only exposes platform independent SqlQueryResult (and other interfaces), though internally it calls  the same BQ functions. Additionally even the executor is platform specific, so it can't be exposed in the underlay interface.
- indexer cannot added as a dependency to underlay (creates a circular dependency) 

so, I would prefer to keep it as it is in the PR - open to renaming the functions if preferred. 
I will however check if i can extract the processing of a single row to a static util

dexamundsen: update - even the individual rows do not align in data types, its best to avoid major refactor now.

dexamundsen:
> As discussed, please add a test at least to the actual SQL generation.

@tjennison-work added a test for entire entity based hints. SQL generation for filter itself is part of indexer / underlay test

ui/public/manifest.json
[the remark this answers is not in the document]
tjennison-work: No, just the type line.

indexer/src/main/java/bio/terra/tanagra/indexing/job/bigquery/WriteEntityLevelDisplayHints.java
tjennison-work: Okay, sounds good.

ui/public/manifest.json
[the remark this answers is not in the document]
tjennison-work: Oh, you changed it to png. That's fine too.

tjennison-work (approved):

PR #1051, opened 2024-10-24
Snyk updates for java projects

dexamundsen:


annotationProcessor/build.gradle
```diff
@@ -34,30 +29,11 @@ repositories {
 }
 
 dependencies {
-    // added to address synk warnings
-    constraints {
-        implementation 'com.nimbusds:nimbus-jose-jwt:9.37.3'
```
dexamundsen, line 39: All of these are transitively managed via service, hence not needed

tjennison-work (approved):

PR #1053, opened 2024-10-25
[VDA-1785,VDA-1847] Added api for all studies for admin user

chenchals:
- updated admin-client for new api call
- updated core code to optionally use `accessGroup` for getting all studies
- ran all local service tests


service/src/test/java/bio/terra/tanagra/service/accesscontrol/AouWorkbenchAccessControlTest.java
```diff
@@ -295,7 +295,11 @@ private void assertStudyPermissions(UserId user, ResourceId resource, Action...
   private void assertStudyNoListPermissions(UserId user, ResourceType resource, Action... actions) {
     ResourceCollection resources =
         accessControlService.listAuthorizedResources(
-            user, Permissions.forActions(resource, actions), 0, Integer.MAX_VALUE);
+            user,
+            Permissions.forActions(resource, actions),
+            (ResourceId) null,
```
dexamundsen, line 300: The effect is the same, but this should be cast to string iiuc. 
for list underlays: resource id is correct since it is for parent resource. this in turn calls the function with string param
for studies, it would be user group. 
---
If the overloaded functions are too confusing we can look at changing it in a followup PR

service/src/test/java/bio/terra/tanagra/service/accesscontrol/impl/MockVumcAdminAccessControl.java
```diff
@@ -24,7 +24,8 @@ protected boolean apiIsAuthorizedUser(String userEmail) {
   }
 
   @Override
-  protected ResourceList apiListAuthorizedResources(String userEmail, ResourceType resourceType) {
+  protected ResourceList apiListAuthorizedResources(
+      String userEmail, ResourceType resourceType, String googleGroup) {
```
dexamundsen, line 28: minor: change param name to userAccessGroup or accessGroup

dexamundsen (commented):
minor comments, not blocker

service/build.gradle
```diff
@@ -146,7 +146,7 @@ dependencies {
     implementation "org.glassfish.jersey.media:jersey-media-multipart:${vJersey}"
     implementation "org.glassfish.jersey.media:jersey-media-json-jackson:${vJersey}"
 
-    implementation 'org.vumc.vda.tanagra:admin-client:0.0.113-SNAPSHOT'
+    implementation 'org.vumc.vda.vumc-cohort-builder:admin-client:1.0.10-SNAPSHOT'
```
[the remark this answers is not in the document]
chenchals: Yes - fixed both the group-name and version: to 1.0.12

service/src/test/java/bio/terra/tanagra/service/accesscontrol/AouWorkbenchAccessControlTest.java
chenchals: The compiler complained, so I did the cast to specifically call what it was previously calling :-) Yes maybe in a follow-up PR we can fix this

dexamundsen: I mean you would need to cast it to string

dexamundsen (approved):

service/src/test/java/bio/terra/tanagra/service/accesscontrol/impl/MockVumcAdminAccessControl.java
chenchals: missed that. Will do the change before merge.  Thanks.

PR #1058, opened 2024-10-30
[bench-4474] Return Cohort SQL in previewExport response

dexamundsen:
change 1: Create a util function to check access to cohorts, feature sets. This will be used in controllers. Also updates access checks in controllers to verify that user has access to cohort and feature sets

change 2: Return Cohort SQL in describeExport response

cohorts
image
image

request 
```
http://localhost:3000/v2/underlays/cmssynpuf/describeExport
{study: "P63tR0rsZJ", cohorts: ["L8v1TKKCfG", "6RyN4mpOd7"], featureSets: ["GmyHqDVAqa", "JmXfpH6BKu"]}
```

response field
```
    SELECT id 
    FROM`verily-tanagra-dev.cmssynpuf_index_060724`.ENT_person 
    WHERE
        (
            (
                CAST(FLOOR(TIMESTAMP_DIFF(TIMESTAMP('2024-11-01 12:26:16.709'), age, DAY) / 365.25) AS INT64) BETWEEN 89.0 AND 95.0
            ) 
            AND (
                gender = 8507
            ) 
            AND (
                ethnicity = 38003564
            ) 
            AND (
                race = 8516
            )
        ) 
        OR (
            (
                CAST(FLOOR(TIMESTAMP_DIFF(TIMESTAMP('2024-11-01 12:26:16.709'), age, DAY) / 365.25) AS INT64) BETWEEN 96.0 AND 98.0
            ) 
            AND (
                gender = 8532
            ) 
            AND (
                race = 0
            ) 
            AND (
                ethnicity = 38003563
            )
        )
```

### DIFF
--- a/service/src/main/java/bio/terra/tanagra/app/controller/ExportApiController.java
+++ b/service/src/main/java/bio/terra/tanagra/app/controller/ExportApiController.java
@@ -19,9 +19,6 @@ import bio.terra.tanagra.generated.model.ApiExportPreviewRequest;
 import bio.terra.tanagra.generated.model.ApiExportRequest;
 import bio.terra.tanagra.generated.model.ApiExportResult;
 import bio.terra.tanagra.generated.model.ApiInstanceListResult;
-import bio.terra.tanagra.query.bigquery.translator.BQApiTranslator;
-import bio.terra.tanagra.query.sql.SqlParams;
-import bio.terra.tanagra.query.sql.SqlQueryRequest;
 import bio.terra.tanagra.service.UnderlayService;
 import bio.terra.tanagra.service.accesscontrol.AccessControlService;
 import bio.terra.tanagra.service.artifact.CohortService;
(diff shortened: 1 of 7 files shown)


service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2087,6 +2087,9 @@ components:
       properties:
         sql:
           type: string
+        entityIdSql:
```
freemabd, line 2090: We would like to add this field to the describeExport API call. We are currently using this endpoint cause the [response](https://github.com/DataBiosphere/tanagra/blob/9f049c438af6002d25636c16503287e45487e636/service/src/main/resources/api/service_openapi.yaml#L2825-L2861) provides : 

- indexSql - if source tables aren't configured for this entity
- sourceSql - if source tables are configured for this entity

We would like to add entityIdSql to this response.

freemabd (commented):
Asking for changes on describeExport API call

service/src/main/resources/api/service_openapi.yaml
dexamundsen: sounds good changed

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2864,10 +2864,14 @@ components:
           description: |
             SQL string against the source tables.
             This will be populated if source queries are configured for this entity.
+        entityIdSql:
```
tjennison-work, line 2867: Should this go in `EntityOutputPreviewList` since it's not unique per entity?

freemabd: Good point. Yes, it isn't unique so makes sense to move it to EntityOutputPreviewList

dexamundsen: moved

service/src/main/java/bio/terra/tanagra/app/controller/ExportApiController.java
```diff
@@ -169,26 +146,18 @@ public ResponseEntity<ApiInstanceListResult> previewExportInstances(
...
-        Permissions.forActions(UNDERLAY, READ),
-        ResourceId.forUnderlay(underlayName));
-    for (String featureSetId : body.getFeatureSets()) {
-      accessControlService.throwIfUnauthorized(
-          SpringAuthentication.getCurrentUser(),
-          Permissions.forActions(FEATURE_SET, READ),
-          ResourceId.forFeatureSet(body.getStudy(), featureSetId));
-    }
+    accessControlService.checkUnderlayAccess(underlayName);
+    accessControlService.checkReadAccess(body.getStudy(), body.getCohorts(), body.getFeatureSets());
```
tjennison-work, line 150: If I'm understanding this correctly, the reason for the refactoring is to check cohorts here, which was previously missing. If that is the case, please also file a task to validate the remaining permissions and add some sort of verification. Either way, please note the change in the PR description.

tjennison-work (approved):

service/src/main/java/bio/terra/tanagra/app/controller/ExportApiController.java
dexamundsen: I had meant to add a comment that I missed. 
created https://verily.atlassian.net/browse/BENCH-4502

PR #1062, opened 2024-11-12
[VDA-1852] Set filter range for BMI, height and weight

chenchals:
- updated display hint range min/max for BMI, Height and Weight
- tested on local deploy of tanagra

### DIFF
--- a/underlay/src/main/resources/config/datamapping/emerge/entity/bmi/entity.json
+++ b/underlay/src/main/resources/config/datamapping/emerge/entity/bmi/entity.json
@@ -4,7 +4,7 @@
   "attributes": [
     { "name": "id", "dataType": "INT64", "valueFieldName": "row_id" },
     { "name": "person_id", "dataType": "INT64" },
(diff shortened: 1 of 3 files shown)


underlay/src/main/resources/config/datamapping/emerge/entity/bmi/entity.json
```diff
@@ -4,7 +4,7 @@
   "attributes": [
     { "name": "id", "dataType": "INT64", "valueFieldName": "row_id" },
     { "name": "person_id", "dataType": "INT64" },
-    { "name": "value_numeric", "dataType": "DOUBLE", "valueFieldName": "value_as_number", "isComputeDisplayHint": true },
+    { "name": "value_numeric", "dataType": "DOUBLE", "valueFieldName": "value_as_number", "isComputeDisplayHint": true, "displayHintRangeMin": 0, "displayHintRangeMax": 500 },
```
dexamundsen, line 7: minor: isn't max 500 too high? may be 50?

dexamundsen (approved):

PR #1063, opened 2024-11-14
Add prepackaged WGS data feature to AoU underlay

dolbeew:
Add a `Short Read WGS` option under the Predefined data feature section. Does not display any columns since this is only for VCF extractions
Screenshot 2024-11-14 at 12 36 43 PM
Screenshot 2024-11-14 at 12 36 55 PM


ui/src/export.tsx
```diff
@@ -386,6 +386,9 @@ function Preview(props: PreviewProps) {
             message: "EXPORT",
             resources: {
               allParticipantsCohort: props.allParticipantsCohort,
+              hasWGSFeatureSet: filteredFeatureSets.some((fs) =>
+                fs.predefinedCriteria.includes("_short_read_wgs")
+              ),
```
dolbeew, line 391: @tjennison-work This is a quick solution for notifying the workbench that an export would need a VCF extraction but may be too AoU-specific. Let me know if this needs to be changed

tjennison-work: How about including the list of predefined criteria instead of searching for a particular one so it could be used for any special exports:
```
predefinedCriteria: [...new Set(filteredFeatureSets.map((fs) => fs.predefinedCriteria).flat())],
```

tjennison-work (approved):
We'll probably want to add some sort of message to the UI for this case eventually but this is fine for now.

ui/src/export.tsx
dolbeew: That makes more sense, good call

PR #1065, opened 2024-11-18
Add UI config flag to disable feature sets

tjennison-work:
Currently only affects the feature set criteria button but can be expanded as needed for cohort-only style underlays/deployments.


chenchals (approved):
Great! How do I specify this in ui.json?

PR #1060, opened 2024-11-08
Update constraints in  java build files / Constrain lib in UI

dexamundsen:
Update constraints in java build files. These are no longer needed since other libs get same or newer transitive libs

Also excludes common-logging jar to address below warning
`
Standard Commons Logging discovery in action with spring-jcl: please remove commons-logging.jar from classpath in order to avoid potential conflicts
`

--

Add constraint to UI lib to address snyk vuln

PR #1068, opened 2024-11-18
Revert "Update constraints in  java build files / Constrain lib in UI (#1060)"

tjennison-work:
The is causing indexing failures due to missing libraries:

```
Exception in thread “main” java.lang.NoClassDefFoundError:
ch/qos/logback/classic/Level
```

Removing the exclusion from indexing/ fails at a different location processing DataFlow results so just rolling back the entire change for now.

This reverts commit fdbc5b3af00551ac9ced03be3ae2d21bce3f5d8d.


freemabd (approved):
Thanks Tim.

PR #1073, opened 2024-11-22
Replace deprecated function RandomStringUtils.randomAlphanumeric

dexamundsen:
Addresses build time warnings


tjennison-work (commented):
Is it possible to run out of secure random data?

dexamundsen:
> Is it possible to run out of secure random data?

1. the current function internally calls the same secure() function. So this change itself does not have any impact
2. Larger question: I had brought this up with Mariko as well. We have not seen conflicts before only because we have not had too much use. My suggestion is to use standard UUID. they are globally unique . 

I can switch over to UUID in this PR itself

tjennison-work:
> > Is it possible to run out of secure random data?
> 
> 1. the current function internally calls the same secure() function. So this change itself does not have any impact
> 2. Larger question: I had brought this up with Mariko as well. We have not seen conflicts before only because we have not had too much use. My suggestion is to use standard UUID. they are globally unique .
> 
> I can switch over to UUID in this PR itself

Ah, I guess don't worry about it then if it hasn't been a problem so far. I wouldn't have expected the function we were calling was a secure version, which is probably why that API is being deprecated.

I'd be disinclined to switch to UUIDs because they'll show up in the URLs in the UI, which makes the URLs very long especially when there are multiple. Maybe that's not a big deal though.

tjennison-work (approved):

PR #1074, opened 2024-11-22
Followup to fix snyk vuln

dexamundsen:
1. remove duplicate repository declaration
2. exclude vuln transitive lib


service/build.gradle
```diff
@@ -106,7 +96,9 @@ dependencies {
 
     // [Application DB] MySQL/MariaDB
     implementation 'org.mariadb.jdbc:mariadb-java-client:2.7.12'
-    implementation 'com.mysql:mysql-connector-j:8.4.0'
+    implementation('com.mysql:mysql-connector-j:8.4.0') {
+        exclude group: 'com.google.protobuf', module: 'protobuf-java'
```
dexamundsen, line 100: This is already included, and current one adds a vuln version

underlay/build.gradle
```diff
@@ -5,19 +5,6 @@ plugins {
 
 sourceCompatibility = JavaVersion.VERSION_17
 
-repositories {
-    mavenCentral()
```
dexamundsen, line 9: these are duplicates. All are declared in build src already

client/build.gradle
```diff
@@ -12,12 +12,6 @@ plugins {
 
 repositories {
     mavenCentral()
-    maven {
```
dexamundsen, line 15: not used

PR #1078, opened 2024-11-25
Use datasets in broad-tanagra-dev for PR checks

dexamundsen:


tjennison-work (approved):
Should there be indexer file changes to go along with these as well?

dexamundsen:
> Should there be indexer file changes to go along with these as well?

There were no corresponding indexer configs, hence skipped them. They were likely indexed in verily-tanagra-dev or sd projects
Also - Not aware if the indexer env infra is set up at all in this project . since it might not be used long term, so setting up the indexing infra might not be needed

PR #1080, opened 2024-11-26
[OOB] Test turning off "Review individuals" button on the overview page

chenchals:
- testing to see if "review individual" can be turned off for eMerge

### DIFF
--- a/ui/src/overview.tsx
+++ b/ui/src/overview.tsx
@@ -137,7 +137,7 @@ export function Overview() {
         rows="minmax(max-content, 100%)"
         cols="minmax(440px, auto) 450px"
         spacing={2}
-        sx={{ display: "flex", px: 5, overflow: "auto" }}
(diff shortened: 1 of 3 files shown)


ui/src/overview.tsx
```diff
@@ -147,18 +148,22 @@ export function Overview() {
...
-              <Button
-                variant="outlined"
-                component={RouterLink}
-                to={absoluteCohortReviewListURL(params, cohort.id)}
-              >
-                Review individuals
-              </Button>
-            }
-          />
+          {underlay.uiConfiguration.featureConfig?.disableCohortReview ? (
```
tjennison-work, line 151: Put this condition inside the extraControls parameter so DemographicCharts isn't duplicated. Also, use undefined rather than null.

tjennison-work (commented):
This worked as expected for me.

PR #1101, opened 2024-12-12
move testonly entities to aouRT folder; use in underlay

dexamundsen:


underlay/src/main/resources/config/underlay/aouSC2023Q3R2_testonly/underlay.json
```diff
@@ -34,11 +34,11 @@
     "aouRT/icd10cmConcept",
     "aouRT/icd10pcsConcept",
 
-    "aouRT/activitySummary",
+    "aouRT_testonly/activitySummary",
     "aouRT/heartRateLevel",
```
tjennison-work, line 38: I'm surprised to see only some of these apply. Do `heartRateLevel` and `stepsIntraday` not need the same setup? Alternatively, should we see about getting the updated tables instead (or also)?

tjennison-work (approved):

underlay/src/main/resources/config/underlay/aouSC2023Q3R2_testonly/underlay.json
dexamundsen: The other entities were already calculating the row_number, so not changing them. as of test data update - lets wait till aou on vwb needs us to update it

PR #1107, opened 2024-12-19
PM, observation changes for AoU underlays

dolbeew:
- Add pregnant at enrollment, wheelchair at enrollment and has any PM data to physical measurements
- Modify observationConcept SQL to more closely match current data apps


freemabd (approved):
LGTM

PR #1121, opened 2025-01-13
DT-343 changing study label and data source to displayName instead

freemabd:
Added featureConfig attribute called renameStudyLabel. This will allow us to override the study label and use workspace instead. 
Screen Shot 2025-01-13 at 12 01 11 PM
Changed underlay name to now be displayName. @tjennison-work wanted to confirm with you that this change is ok. AoU users are used to seeing the display name instead of the actual dataset name. 
Screen Shot 2025-01-13 at 12 01 31 PM

### DIFF
--- a/ui/src/export.tsx
+++ b/ui/src/export.tsx
@@ -136,7 +136,7 @@ export function Export() {
             <StudyName />
             <Typography variant="body1">•</Typography>
             <Typography variant="body1">
-              Data source: {underlay.name}
+              Data source: {underlay.underlayConfig.metadata.displayName}
             </Typography>
           </GridLayout>
         }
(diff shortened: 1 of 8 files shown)


tjennison-work:
The timing on this is a bit unfortunate. I was planning to remove that status line entirely since it's duplicating the information that's already available in the header, but without the ability to click on it for navigation. How would you feel about that approach instead?

freemabd:
> The timing on this is a bit unfortunate. I was planning to remove that status line entirely since it's duplicating the information that's already available in the header, but without the ability to click on it for navigation. How would you feel about that approach instead?
 
I'm fine with the status line being removed. We already show the workspace and dataset display name in the workbench header. Thanks, I will close this PR.

freemabd:
Closing this PR as this status line will be removed in a future PR.

PR #1122, opened 2025-01-13
Adding filterable option to source_value column in labs review tab

C-Burden:
# OOB: Update 
## Intent
Make source_value column in labs review tab searchable


chenchals (approved):
LGTM

PR #1123, opened 2025-01-13
Fix several small issues on the cohort overiew

tjennison-work:
* Swap cohort criteria to left side. Make the visualizations secondary until they're interactive enough to be the primary editing surface.
* Make the cohort criteria wider. Almost all criteria were too long to be effectively displayed.
* Remove interactivity from the background of criteria so only the title selects/unselects. It was causing a number of corner cases where criteria controls behaved unpredictably and some were difficult to solve due to the way JS handles mouse events. This removes the need to squash mouse events throughout the code. Show title hover underlining on selected criteria as well to make it clear it's still clickable.
* Refactor uncontained select style to a shared file.
* Fix rare issue were criteria buttons could spill onto a second row.
* Add a drop down to select which visualizations are displayed. Add optional config to limit the default set displayed.

### DIFF
--- a/ui/src/demographicCharts.tsx
+++ b/ui/src/demographicCharts.tsx
@@ -1,11 +1,19 @@
+import MenuItem from "@mui/material/MenuItem";
+import OutlinedInput from "@mui/material/OutlinedInput";
+import Select, { SelectChangeEvent } from "@mui/material/Select";
+import FormControl from "@mui/material/FormControl";
 import Paper from "@mui/material/Paper";
 import Typography from "@mui/material/Typography";
+import { uncontainedSelectSx } from "components/select";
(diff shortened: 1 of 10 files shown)


ui/src/demographicCharts.tsx
```diff
@@ -38,13 +84,56 @@ export function DemographicCharts(props: DemographicChartsProps) {
...
+              title="No visualizations selected"
+            />
           )}
         </GridBox>
       </GridLayout>
     </Paper>
   );
 }
+
+// TODO(tjennison): Store the selected visualizations in local storage per
```
dexamundsen, line 107: nit: jira tracking id

dexamundsen (approved):
Not too familiar with UI, no issues stand out

PR #1144, opened 2025-01-24
Remove logs meant for auth debugging

dexamundsen:
- These logs are not useful after deployment is tested. removing this to reduce noise
- Update ExportPreviewRequest limit to 50 (was 250)
- rename parameter objects to use prefix Param to avoid name confusion with other schema objects

### DIFF
--- a/service/src/main/resources/api/service_openapi.yaml
+++ b/service/src/main/resources/api/service_openapi.yaml
@@ -1295,7 +1295,7 @@ components:
       description: ID of the activity log entry
       required: true
       schema:
-        type: string
+        $ref: "#/components/schemas/ActivityLogEntryId"
 
   responses:
     BadRequest:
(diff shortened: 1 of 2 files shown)


service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2819,7 +2819,9 @@ components:
           items:
             type: string
         limit:
-          $ref: "#/components/schemas/Limit"
+          type: integer
```
tjennison-work, line 2822: This was probably because done unknowingly because there's a `Limit` parameter which does have a maximum of 50 which is different from the component version. I didn't look enough to see what the best change would be but probably one should get renamed at least.

tjennison-work (approved):

service/src/main/resources/api/service_openapi.yaml
dexamundsen: From our discussion - found the other param_limit. Refactored it to re-use the same schema object with same default of 50

dexamundsen: Update - reusing the schema object does not work (default and required fields are ignored). renamed the param instead

PR #1145, opened 2025-01-24
Add new API to fetch visualizations

dexamundsen:
- in current PR initial cohorts (w/o filters) will be supported. it fetches precomputed values
- For those with filters - looking into adding changes to current PR, otherwise in next PR. does not include optimizing backend queries, just combines it into a single API (moving the parallel api calls from UI to backend)

---

obj struct copied from UI

### DIFF
--- a/service/src/main/resources/api/service_openapi.yaml
+++ b/service/src/main/resources/api/service_openapi.yaml
@@ -556,6 +556,40 @@ paths:
         500:
           $ref: "#/components/responses/ServerError"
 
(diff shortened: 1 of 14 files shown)


service/src/main/resources/api/service_openapi.yaml
```diff
@@ -2364,6 +2399,59 @@ components:
       type: string
       description: Description of the cohort
 
+    VisualizationList:
```
dexamundsen, line 2402: obj struct copied from UI

PR #1147, opened 2025-01-27
Add support for adding criteria from a list of codes

tjennison-work:


underlay/src/main/resources/config/underlay/cmssynpuf/ui.json
```diff
@@ -1,4 +1,7 @@
 {
+  "featureConfig": {
+    "enableAddByCode": true
```
dexamundsen, line 3: should we enable this for aou_test as well?

ui/src/plugins/vumc/biovu.tsx
```diff
@@ -55,10 +55,16 @@ interface Data {
   (
     underlaySource: UnderlaySource,
     c: CommonSelectorConfig,
-    dataEntry?: DataEntry
+    dataEntries?: DataEntry[]
```
dexamundsen, line 58: is array support for future functions? i.e. only 1 entry is supported in this code

dexamundsen (approved):

underlay/src/main/resources/config/underlay/cmssynpuf/ui.json
tjennison-work: I don't think it makes sense to duplicate all of the criteria since then they'll diverge from the underlying ones. Measurement was already duplicated though so I added it there.

PR #1151, opened 2025-01-30
Simplify merge translator + add test

dexamundsen:
Simplify merge translator (will be used in next pr to merge other filters)
add test

### DIFF
--- a/indexer/src/main/java/bio/terra/tanagra/indexing/JobSequencer.java
+++ b/indexer/src/main/java/bio/terra/tanagra/indexing/JobSequencer.java
@@ -74,15 +74,13 @@ public final class JobSequencer {
       jobSet.startNewStage();
 
       entity
-          .getOptimizeSearchByAttributes()
+          .getOptimizeSearchByAttributeNames()
           .forEach(
-              searchTableAttributes -> {
+              searchTableAttributeNames -> {
                 ITEntitySearchByAttributes searchTable =
                     underlay
                         .getIndexSchema()
-                        .getEntitySearchByAttributes(
-                            entity,
-                            searchTableAttributes.stream().map(Attribute::getName).toList());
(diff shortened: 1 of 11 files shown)


indexer/src/main/java/bio/terra/tanagra/indexing/JobSequencer.java
```diff
@@ -74,15 +75,14 @@ public static SequencedJobSet getJobSetForEntity(
       jobSet.startNewStage();
 
       entity
-          .getOptimizeSearchByAttributes()
+          .getOptimizeSearchByAttributeNames()
```
dexamundsen, line 78: rename and change to set since it is better for perf (contains all function)

underlay/src/main/java/bio/terra/tanagra/api/filter/EntityFilter.java
```diff
@@ -47,8 +47,8 @@ public EntityFilter mergeWith(EntityFilter entityFilter) {
   }
 
   public static boolean areSameFilterType(List<EntityFilter> filters) {
-    String filterType = filters.get(0).getClass().getName();
-    return filters.stream().allMatch(filter -> filter.getClass().getName().equals(filterType));
+    Class<?> firstClazz = filters.get(0).getClass();
+    return filters.stream().skip(1).allMatch(filter -> filter.getClass().equals(firstClazz));
```
dexamundsen, line 51: skip 1 to avoid comparing first element to itself

tjennison-work (approved):

indexer/src/main/java/bio/terra/tanagra/indexing/JobSequencer.java
dexamundsen: moved back to is since we need the same order to generate table names

PR #1152, opened 2025-01-30
Adding retry if concurrent update exception during indexing build

freemabd:
Adding retry if concurrent update exception during indexing build

### DIFF
--- a/indexer/src/main/java/bio/terra/tanagra/indexing/job/dataflow/WriteRollupCounts.java
+++ b/indexer/src/main/java/bio/terra/tanagra/indexing/job/dataflow/WriteRollupCounts.java
@@ -178,6 +178,13 @@ public class WriteRollupCounts extends BigQueryJob {
             LOGGER.info("Max retries reached. Giving up.");
             throw e; // Rethrow after max attempts
           }
+
+          try {
+            Thread.sleep(1000); // Sleep for 1 second
+          } catch (InterruptedException ie) {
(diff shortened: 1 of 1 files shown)


indexer/src/main/java/bio/terra/tanagra/indexing/job/dataflow/WriteRollupCounts.java
```diff
@@ -159,9 +160,29 @@ public void run(boolean isDryRun) {
...
+        // count values to the corresponding column in the index entity main table.
+        copyFieldsToEntityTable(isDryRun);
+        return; // Exit if successful
+      } catch (BigQueryException e) {
+        if (e.getMessage().contains("due to concurrent update")) {
+          attempt++;
+          LOGGER.info("Attempt {} failed: {}", attempt, e.getMessage());
+
+          if (attempt == maxRetries) {
+            LOGGER.info("Max retries reached. Giving up.");
```
dexamundsen, line 178: minor: would a sleep help?

dexamundsen (approved):

indexer/src/main/java/bio/terra/tanagra/indexing/job/dataflow/WriteRollupCounts.java
freemabd: It's working in my test runs but won't hurt to add it.

PR #1161, opened 2025-02-05
Change max cohort review count to 9999

dexamundsen:
image

image

### DIFF
--- a/service/src/main/java/bio/terra/tanagra/service/artifact/ReviewService.java
+++ b/service/src/main/java/bio/terra/tanagra/service/artifact/ReviewService.java
@@ -56,7 +56,12 @@ import org.springframework.stereotype.Component;
 @Component
 public class ReviewService {
   private static final Logger LOGGER = LoggerFactory.getLogger(ReviewService.class);
-  public static final int MAX_REVIEW_SIZE = 10_000;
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/tanagra/service/artifact/ReviewService.java
```diff
@@ -56,7 +56,12 @@
 @Component
 public class ReviewService {
   private static final Logger LOGGER = LoggerFactory.getLogger(ReviewService.class);
-  public static final int MAX_REVIEW_SIZE = 10_000;
+  // BigQuery has a parameter limit of 10_000. Cohort reviews generate a query
+  // that lists all the primaryEntityIds. The other cohort filter values are
+  // also counted as parameters. Hence, limit the number of cohorts to
+  // 9_900 (+ 100 for parameters = 10_000)
```
tjennison-work, line 62: Where do the parameters come from in these queries? 9999 (4 digits) would feel like a less arbitrary number than 9900 if we only need space for one.

dexamundsen: Did a little more debugging and found that review person IDs are already stored in a table and used in the final select statement directly. (had assumed it would be a nested query). 
changed it to 9999 to account for currentTimestaml

PR #1164, opened 2025-02-06
Add null values to display hint as 'n/a'

dexamundsen:
issue: The n/a option is not showing up in the filters within the variant search which is excluding some participants.  'n/a not showing up on Filters' -- this is because null (empty) values were excluded from hint tables, and the filters get the list directly from the hint tables.

rootcause: empty values are not added to display hints table

indexer:
- add empty values (repeated data type) to hints
- no change needed for non-repeated types

ui:
- if hint value is null, display as 'n/a'
- if null value is selected, send it as unary.isNull filter

service:
- add support for unary operator support for repeated data type
- update filterbuilder to match UI 
- addressed compiler warnings
- update test

underlay:
- and attribute property emptyValueDisplay
￼
Pasted Graphic 9
Consequence
Filters

----

sample query for a different filter

```
 SELECT COUNT(id) AS T_CTDT FROM T_ENT_person 
    WHERE id IN (SELECT flattened_person_id 
        FROM
            (SELECT DISTINCT vid, flattened_person_id 
            FROM `src.cb_variant_to_person` 
            CROSS JOIN UNNEST(person_ids) AS flattened_person_id  
            WHERE
                vid IN (SELECT
                    id 
                FROM  T_ENT_variant 
                WHERE
                    (id IN (SELECT
                        id 
                    FROM T_ESA_variant_gene 
                    WHERE
                        gene = @val0)) 
                    AND ((NOT EXISTS (SELECT *  FROM UNNEST(consequence))) 
                    OR (EXISTS (SELECT *  FROM UNNEST(consequence) AS flattened WHERE flattened IN (@val1)))) 
                    AND (allele_count BETWEEN @val2 AND @val3)))) 
    ORDER BY T_CTDT DESC 
    LIMIT 1000000
```

merge after https://github.com/DataBiosphere/tanagra/pull/1171 & https://github.com/DataBiosphere/tanagra/pull/1170


ui/src/data/source.tsx
```diff
@@ -2237,7 +2237,7 @@ function fromAPIDisplayHint(hint?: tanagra.DisplayHint): HintData {
         const value = dataValueFromLiteral(enumHintValue.enumVal.value);
         return {
           value,
-          name: enumHintValue.enumVal.display ?? String(value),
+          name: enumHintValue.enumVal.display ?? String(value ?? "n/a"),
```
tjennison-work, line 2240: I think it would be better to index the `n/a` string into the hints. That way it could be configurable per entity eventually and would get sorted properly in the list. It feels better to me to make the hints fully consistent but the UI could potentially do it here too.

underlay/src/test/resources/sql/BQAttributeFilterTest/repeatedAttributeFilterUnaryIsNotNull.sql
```diff
@@ -0,0 +1,10 @@
+
+    SELECT
+        name      
+    FROM
+        ${ENT_condition}      
+    WHERE
+        EXISTS (SELECT
+            *          
+        FROM
+            UNNEST(['foo', 'bar', 'baz', vocabulary]))
```
tjennison-work, line 10: I have no idea what this SQL means. Shouldn't there be and `IS NOT NULL` or something? Does it not work correctly when applied to a repeated field?

ui/src/criteria/filterableGroup.tsx
```diff
@@ -1190,16 +1190,47 @@ function generateFilters(
...
+            },
+          },
+        });
+      }
+
+      operands.push(
+        makeBooleanLogicFilter(
+          tanagra.BooleanLogicFilterOperatorEnum.Or,
+          operandFilters
+        ) as tanagra.Filter
```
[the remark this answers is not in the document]
dexamundsen: done

underlay/src/test/java/bio/terra/tanagra/filterbuilder/FilterableGroupFilterBuilderTest.java
```diff
@@ -374,6 +374,49 @@ void selectAllValueDataFilter() {
     SelectionData selectionData = new SelectionData(null, serializeToJson(data));
     EntityFilter cohortFilter = filterBuilder.buildForCohort(underlay, List.of(selectionData));
     assertEquals(expectedCohortFilter, cohortFilter);
+
+    // Multiple attribute values  including n/a in filter
```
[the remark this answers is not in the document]
dexamundsen: done

underlay/src/test/resources/sql/BQAttributeFilterTest/repeatedAttributeFilterUnaryIsNotNull.sql
dexamundsen: from discussion - this purely a test construct. 
real SQL is of the form

```    
SELECT
        name      
    FROM
        ${ENT_person}      
    WHERE
        EXISTS (SELECT
            *          
        FROM
            UNNEST(consequence))
```    
here consequence is the name of the repeated (not nullable) field. 

repeated fields are not nullable. an empty value is empty array. so the unnest would produce no results (i.e. not exist)

ui/src/data/source.tsx
dexamundsen: done

dexamundsen: update: i reverted this and added the value in backend before returning response to UI.

PR #1168, opened 2025-02-07
Add cohort review plugin with sub tabs for SD Vitals

dolbeew:
First attempt at a new `entityTabs` plugin to display multiple entities within a single cohort review tab. This is for the SD Vitals entities but may be useful as a core plugin. 

Screenshot 2025-02-07 at 3 22 54 PM

### DIFF
--- a/ui/src/cohortReview/plugins/occurrenceTabs.tsx
+++ b/ui/src/cohortReview/plugins/occurrenceTabs.tsx
@@ -27,10 +27,10 @@ class _ implements CohortReviewPlugin {
     config: CohortReviewPageConfig
   ) {
     this.config = config.plugin as Config;
-    this.entities = this.config.tabs.reduce((acc, t) => {
-      acc.push(getCohortReviewPlugin(t).entities[0]);
-      return acc;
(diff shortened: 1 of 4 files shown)


ui/src/cohortReview/plugins/occurrenceTabs.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { useMemo } from "react";
+import { CohortReviewPageConfig } from "underlaysSlice";
+import { Tabs } from "components/tabs";
+
+interface Config {
+  entity: string;
+  tabs: CohortReviewPageConfig[];
+}
+
+@registerCohortReviewPlugin("entityTabs")
```
dolbeew, line 18: @tjennison-work The main reason I have this as a draft is not sure if this should be a core plugin or VUMC-specific.

tjennison-work: Sorry, I was out the past two days. Core sounds good. It's generically useful.

tjennison-work (approved):

PR #1169, opened 2025-02-11
[BENCH-5181] Fix range data conversion in requests

dexamundsen:
allele count filter (and other with numbers there) are broken

rc: the range fields are numbers and not bigInt.

fix: check for both before converting to literal

image

```
    SELECT
        id,
        gene,
        rs_number,
        consequence,
        clinvar_significance,
        protein_change,
        allele_count,
        allele_number,
        allele_frequency,
        T_RCNT_variantPerson_NOHIER 
    FROM
        `vwb-dev-blissful-eggplant-4805.aou_test_data_SC2023Q3R2_011625`.T_ENT_variant 
    WHERE
        (
            id IN (SELECT
                id 
            FROM
                `vwb-dev-blissful-eggplant-4805.aou_test_data_SC2023Q3R2_011625`.T_ESA_variant_gene 
            WHERE
                gene = 'GENE')
        ) 
        AND (
            allele_count BETWEEN 2 AND 490399
        ) 
        AND (
            allele_number BETWEEN 133 AND 490787
        ) 
        AND (
            allele_frequency BETWEEN 3.0E-6 AND 0.9993
        ) 
    ORDER BY
        T_RCNT_variantPerson_NOHIER DESC 
    LIMIT
        10000000
```

### DIFF
--- a/ui/src/data/source.tsx
+++ b/ui/src/data/source.tsx
@@ -14,6 +14,7 @@ import * as tanagra from "tanagra-api";
 import * as tanagraUnderlay from "tanagra-underlay/underlayConfig";
 import { Underlay } from "underlaysSlice";
 import { isValid } from "util/valid";
+import { DataType } from "tanagra-api";
 
 export type EntityNode = {
   data: DataEntry;
@@ -1036,7 +1037,7 @@ export class BackendUnderlaySource implements UnderlaySource {
     }
 
     const limit = options?.fetchAll ? 100000 : options?.limit;
-    const req = {
+    return {
       entityName: entity.name,
       underlayName: this.underlay.name,
       query: {
@@ -1074,7 +1075,6 @@ export class BackendUnderlaySource implements UnderlaySource {
         pageMarker: options?.pageMarker,
       },
     };
-    return req;
   }
 
   private async queryHints(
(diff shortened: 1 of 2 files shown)


ui/src/data/source.tsx
```diff
@@ -1606,19 +1606,35 @@ function isInternalAttribute(attribute: string): boolean {
 }
 
 export function literalFromDataValue(value: DataValue): tanagra.Literal {
-  let dataType = tanagra.DataType.Int64;
-  if (typeof value === "string") {
+  let dataType: DataType;
+  if (typeof value === "bigint") {
+    dataType = tanagra.DataType.Int64;
+  } else if (typeof value == "number") {
+    dataType = Number.isInteger(value)
```
tjennison-work, line 1613: Is this required? The type should reflect the type in the database and numbers should only be used for DOUBLEs now.

dexamundsen: Wasn't sure, so had kept it. removing now

tjennison-work: Looks like it's still here?

dexamundsen: my bad, one change was left out during push from local

tjennison-work (approved):

PR #1170, opened 2025-02-11
Sort HintQueryResult instead of orderBy on ELDH_IndexTable

dexamundsen:
move OrderBy clause from hints index table. this is not used when hints are dynamically calculated. Instead use comparators in controller. create and use comparators in the controller during object creation / conversions.


service/src/main/java/bio/terra/tanagra/app/controller/UnderlaysApiController.java
```diff
@@ -315,33 +324,51 @@ private ApiEntity toApiObject(Entity entity) {
...
+      // order by: attribute ASC
+      return StringUtils.compare(o1.getAttribute().getName(), o2.getAttribute().getName());
+    }
+  }
+
+  private static class ApiDisplayHintEnumValuesComparator
+      implements Comparator<ApiDisplayHintEnumEnumHintValues>, Serializable {
+    @Override
+    public int compare(ApiDisplayHintEnumEnumHintValues o1, ApiDisplayHintEnumEnumHintValues o2) {
+      // order by: value ASC, display ASC, count DESC
```
tjennison-work, line 362: As before, better to sort by display first since that's what the user sees.

tjennison-work (approved):

PR #1178, opened 2025-02-21
"generateUnsignedUrls" for local development

mspector:
adds a shared export configuration, `generateUnsignedUrls`, for use in local development and testing.

### DIFF
--- a/docs/generated/APPLICATION_CONFIG.md
+++ b/docs/generated/APPLICATION_CONFIG.md
@@ -323,14 +323,14 @@ Comma separated list of all GCS bucket names that all export models can use. Onl
 
 *Example value:* `bq-export-uscentral1,bq-export-useast1`
 
-### tanagra.export.shared.generateSignedUrls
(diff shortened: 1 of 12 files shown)


docs/generated/APPLICATION_CONFIG.md
```diff
@@ -323,6 +323,15 @@ Comma separated list of all GCS bucket names that all export models can use. Onl
 
 *Example value:* `bq-export-uscentral1,bq-export-useast1`
 
+### tanagra.export.shared.generateSignedUrls
```
tjennison-work, line 326: It would be better to flip this to `generateUnsignedUrls` so that the default is the expected prod behavior. Less chance of it getting set incorrectly and causing an outage or security issue.

tjennison-work (approved):

mspector:
hold off on merging this for now: it looks like I need to make a slight adjustment to the export URL domains when generateUnsignedUrls is true, maybe depending on the export model

PR #1180, opened 2025-02-25
Fix BP export for SD underlay

dolbeew:
Use indexed `ENT_bloodPressure` table for bp source query instead of `x_blood_pressure`

Screenshot 2025-02-24 at 9 32 54 PM
Screenshot 2025-02-24 at 9 33 54 PM


chenchals (approved):
Great job! Thanks

PR #1181, opened 2025-02-25
Migrate to Vite from CRA and upgrade dependencies

tjennison-work:
This contains a variety of changes because of various dependencies between steps:

* CRA (create-react-app) is deprecated and hasn't had updates for a couple years. Migrate to Vite as a replacement framework. I recreated the project from scratch to avoid any leftovers causing issues.
* Migrate to React 18. Required for Vite and it was fairly out of date along with CRA.
* Upgrade most dependencies. Various requirements coming from the other changes make it much more practical to just blanket update to recent versions rather than update individually.

I attempted to make minimal changes beyond what was required to get everything back into a working state. That includes:

* Removing react-final-form-listeners because it has been abandoned doesn't support React 18.
* Tweaking immerjs `produce` imports.
* Updating the `RouterLink` component to work with the updated  MUI.
* Moving the WIP `newCriteria` to a proper context. It should have been done that way anyways.
* Adding proper typing to useMessageListener.
* Adding a trailing slash to `generatePath` due to react-router changes.
* Renaming the Cypress config.
* Some small test fixes.

Still to be done is fixing a number of lint issues based on the updated lint config, as well as adding additional recommended React linters.


dexamundsen:
Does it make it easier if committed to downstream repo directly? 
Ui integration test pending, but rest should be ready for ui

tjennison-work:
> Does it make it easier if committed to downstream repo directly? Ui integration test pending, but rest should be ready for ui

Not really, since I want to make sure the integration tests work for a change like this anyways. Also, I think it's worthwhile have this change here to lessen the divergence.

dexamundsen:
> > Does it make it easier if committed to downstream repo directly? Ui integration test pending, but rest should be ready for ui
> 
> Not really, since I want to make sure the integration tests work for a change like this anyways. Also, I think it's worthwhile have this change here to lessen the divergence.

Noted. do you plan to create a separate PR for downstream repo?

tjennison-work:
> > > Does it make it easier if committed to downstream repo directly? Ui integration test pending, but rest should be ready for ui
> > 
> > 
> > Not really, since I want to make sure the integration tests work for a change like this anyways. Also, I think it's worthwhile have this change here to lessen the divergence.
> 
> Noted. do you plan to create a separate PR for downstream repo?

Yes, that was my plan.

dexamundsen (approved):

PR #1190, opened 2025-03-12
Fix month for UI date formatter

dolbeew:
Fixes issue where dates are displaying the zero-based index for the month
months


chenchals (approved):
LGTM

PR #1192, opened 2025-03-20
Add endpoint for all cohorts

dolbeew:
Add a new `listAllCohorts` endpoint to the cohorts api for use in SD/eMERGE admin apps. Takes a `userAccessGroup` param and should return all cohorts the current user has access to.


chenchals (approved):
Thank you! LGTM

PR #1203, opened 2025-04-22
Comment out unused export models

dolbeew:
Leave only `IPYNB_FILE_DOWNLOAD` as it's the only model currently in use


dolbeew:
No longer needed

PR #1206, opened 2025-05-13
fix attributes for phewas occurrence modifier

dolbeew:
The phewas occurrence modifier had an entity that didn't exist set in the criteria selector config. This config is working locally for me.
Screenshot 2025-05-12 at 1 24 07 PM


C-Burden (approved):
LGTM

PR #1207, opened 2025-05-19
Fix drug name column for universal search results

dolbeew:
Renamed the column key from `label` to `name` since the `drug` entity has the same value set for both keys. Also added a column for item count as a workaround until we can update the universal search config to work for both counts.
Screenshot 2025-05-19 at 3 26 10 PM


C-Burden (approved):
LGTM

PR #1217, opened 2025-06-12
Cb/vda 2265 doc review bug

C-Burden:
### Overview:
Fix bug in Review where selecting doc type would do nothing. (Caused by trying to use JSON.stringify on big Int)

### DIFF
--- a/ui/src/components/hintDataSelect.tsx
+++ b/ui/src/components/hintDataSelect.tsx
@@ -33,11 +33,9 @@ export function HintDataSelect(props: HintDataSelectProps) {
           if (value === undefined) {
             return undefined;
           }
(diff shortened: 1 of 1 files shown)


ui/src/components/hintDataSelect.tsx
```diff
@@ -11,7 +11,7 @@ import React from "react";
 import { isValid } from "util/valid";
 
 export type Selection = {
-  value: DataValue;
+  value: DataValue | string;
```
dolbeew, line 14: This shouldn't be needed as `DataValue` already includes type `string` ([types.ts](https://github.com/DataBiosphere/tanagra/blob/main/ui/src/data/types.ts#L5C1-L5C74)):
```
export type DataValue = null | string | number | bigint | boolean | Date;
```

PR #1221, opened 2025-07-07
Cb/vda 2282 add review pagination

C-Burden:
# VDA-2282

## Intent
Add Pagination to review UI, limit to a single API call.


ui/src/cohortReview/plugins/occurrenceTable.tsx
```diff
@@ -111,8 +128,14 @@ export function OccurrenceTable({
 
         return 0;
       });
+      data.children = data.children.slice(
+        currentPage * rowsPerPage,
+        (currentPage + 1) * rowsPerPage > data.rows.size
+          ? data.rows.size
+          : (currentPage + 1) * rowsPerPage
```
dolbeew, line 135: AFAICT just using `(currentPage + 1) * rowsPerPage` for the end index should work for all cases but let me know if there's a scenario that I'm missing

dolbeew (approved):

PR #1228, opened 2025-07-22
Fix loading spinner on feature set page

dolbeew:
- Loading spinner was not rendering due to data updating on every React render, fixed by removing `showSnackbar` function from the `useEffect` dependency array
- Prevent duplicate calls of `previewExport` endpoint

https://github.com/user-attachments/assets/1b9aa76b-3f35-45d8-87aa-6f90d58b809e


chenchals (approved):
LGTM

PR #1230, opened 2025-07-29
Cb/vda 1752 add cpt4 codes to indexing

C-Burden:
Updating CPT4 data mapping to show more codes.


underlay/src/main/resources/config/datamapping/sd/entity/cpt4/entity.json
```diff
@@ -26,7 +26,7 @@
       "rootIdFieldName": "id",
       "maxDepth": 10,
       "keepOrphanNodes": false,
-      "cleanHierarchyNodesWithZeroCounts": true
+      "cleanHierarchyNodesWithZeroCounts": false
```
chenchals, line 29: This should be left to `true` else we see all zero counts.  we do not want zero counts concepts

chenchals (changes requested):

PR #1234, opened 2025-08-08
Add review config for label-only boolean attributes

dolbeew:
Add a `showIfTrue` property to the underlay config for cohort review attributes. When enabled, it will show the label of the attribute only when the value is true, currently only enabled for deceased.

https://github.com/user-attachments/assets/83ad22e9-f7b6-47aa-b9f0-800a3ec49614

Also adds a `colorOverride` property to set a custom text color

Screenshot 2025-08-11 at 1 05 42 PM

### DIFF
--- a/ui/src/underlaysSlice.ts
+++ b/ui/src/underlaysSlice.ts
@@ -75,7 +75,8 @@ export type CohortReviewConfig = {
 export type CohortReviewAttribute = {
   title: string;
   key: string;
-  labelOnlyBoolean?: boolean;
+  colorOverride?: string;
+  showIfTrue?: boolean;
 };
 
 export type CohortReviewPageConfig = {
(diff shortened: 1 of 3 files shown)


ui/src/underlaysSlice.ts
```diff
@@ -75,6 +75,7 @@ export type CohortReviewConfig = {
 export type CohortReviewAttribute = {
   title: string;
   key: string;
+  labelOnlyBoolean?: boolean;
```
dolbeew, line 78: Don't really like the name `labelOnlyBoolean` but hard to indicate exactly what the property does wihtout making it super long. Let me know if anyone has a better suggestion.

chenchals: I would suggest: "showLabel"  or "showConditional" or "showIfTrue"

chenchals:
Perfect! We user can easily spot the "deceased" participant!

PR #1239, opened 2025-08-29
VDA-2350-notebook investigate

chenchals:
- updated notebook creation
  - Notebook does NOT use template
  - Notebook cells are created in code
  - SQLs for source tables from output entities are included
  - Exercised the notebook in terra environment successfully


dolbeew (approved):

C-Burden (commented):
LGTM

PR #1242, opened 2025-09-22
Update artifact name length limit

dolbeew:
- Update display name limit from 50 to 128
- Modify UI name validation to display help text when limit is exceeded

Limit of 12 seen in screenshots only for testing
Screenshot 2025-09-22 at 4 01 41 PM
Screenshot 2025-09-22 at 4 02 06 PM
Screenshot 2025-09-22 at 4 07 35 PM
Screenshot 2025-09-22 at 4 23 46 PM

### DIFF
--- a/service/src/main/java/bio/terra/tanagra/service/artifact/model/Study.java
+++ b/service/src/main/java/bio/terra/tanagra/service/artifact/model/Study.java
@@ -19,7 +19,7 @@ import org.apache.commons.lang3.builder.HashCodeBuilder;
  */
 @JsonDeserialize(builder = Study.Builder.class)
 public class Study {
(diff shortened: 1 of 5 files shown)


service/src/main/java/bio/terra/tanagra/service/artifact/model/Study.java
```diff
@@ -19,7 +19,7 @@
  */
 @JsonDeserialize(builder = Study.Builder.class)
 public class Study {
-  public static final int MAX_DISPLAY_NAME_LENGTH = 50;
+  public static final int MAX_DISPLAY_NAME_LENGTH = 128;
```
chenchals, line 22: Please update to 255 based on Hana's input in the ticket

PR #1243, opened 2025-09-24
Remove old annotation char limit

dolbeew:
Remove the `maxLength` prop from the annotation name text field that's no longer needed.


chenchals (approved):
Good catch.

